PR #1, opened 2020-11-09
Setup initial infrasctructure for project and repo

brizental:
Alright, there is a bunch happening in this PR, one could argue that I should have sent this is smaller commits, but it's a bit too late for that now. 

Here we have:

- Build setup;
- Linting setup;
- Testing setup;
- No-op implementation for part of the Glean general API;
- Simple static web site sample;
- Repository configuration: README, CODE_OF_CONDUCT, LICENSE, .gitignore.


.eslintrc
```diff
@@ -0,0 +1,56 @@
...
+    "node": true,
+    "browser": true
+  },
+  "rules": {
+    "indent": ["error", 2],
+    "quotes": ["error", "double"],
+    "no-var": "error",
+    "prefer-const": "error",
+    "strict": ["error", "global"],
+    "notice/notice":[ "error", {
```
brizental, line 13: Copied over from the ion /eslintrc. Very nice addition.

.gitignore
```diff
@@ -0,0 +1,116 @@
+# Logs
```
brizental, line 1: This whole file is copied over from https://github.com/github/gitignore/blob/master/Node.gitignore

I have always used it and it has never failed me.

README.md
```diff
@@ -0,0 +1,25 @@
+# Glean.js
```
brizental, line 1: Kinda copied over from the glean repo README.

package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "glean",
+  "version": "0.1.0",
+  "description": "An implementation of Glean, a modern cross-platform telemetry client, for Javascript environments.",
+  "main": "./dist/index.js",
+  "scripts": {
+    "test": "mocha",
+    "lint": "eslint . --ext .ts,.js,.json",
+    "fix": "eslint . --ext .ts,.js,.json --fix",
+    "build:browser-static": "webpack --config webpack.config.browser-static.js --mode production",
```
brizental, line 10: Not happy about the long command and long name for the config file. The file name issue could be solved by putting all the config files into a folder and nesting website config files (future). For the command I don't really know, but I think it is important to differentiate. We will probably have static and lib variations of builds for websites and webextensions.

.eslintrc
```diff
@@ -0,0 +1,56 @@
+{
+  "root": true,
+  "env": {
+    "node": true,
+    "browser": true
+  },
+  "rules": {
```
brizental, line 7: There are opinions going on in these rules, this would be the best moment to argue about them :)

Dexterp37: No arguing at all. They seem like sane defaults.

copyright.txt
```diff
@@ -0,0 +1,3 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
Dexterp37, line 1: Can we rename this file to something like `copyright.template.txt`? (and change the eslint accordingly)

package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "glean",
```
Dexterp37, line 2: Is this `glean` or `glean.js`?

package.json
```diff
@@ -0,0 +1,43 @@
...
+  "version": "0.1.0",
+  "description": "An implementation of Glean, a modern cross-platform telemetry client, for Javascript environments.",
```
Dexterp37, line 4:
```suggestion
  "description": "An implementation of the Glean SDK, a modern cross-platform telemetry client, for Javascript environments.",
```

package.json
Dexterp37: Do we need to settle it in this PR?

src/index.ts
```diff
@@ -0,0 +1,26 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+export = {
```
Dexterp37, line 5: Mh, why is this `export` and not `exports` or `module.exports`?

webpack.config.browser-static.js
```diff
@@ -0,0 +1,31 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+
+const path = require("path"); // eslint-disable-line @typescript-eslint/no-var-requires
```
Dexterp37, line 6: Why do we need the suppression, here?

Dexterp37 (changes requested):

.eslintrc
brizental: Now, just because I thought we were going into a heated debate of 2 space indents vs. 4 space indents...

copyright.txt
brizental: Sure, reminder that this is copied over from Ion, maybe you'd want to hcange there as well if you think it is a better naming :)

package.json
brizental: I'd prefer only 'glean'. It is implicit from the fact that this is a `package.json` file that this is Javascript we are talking about. Also, I personally don't like when npm packages have the .js suffix, my mind always goes "ISN'T IT OBIVOUS?!?!?!?!?!".

Anyways, I am open for changing, but I'd need to be convinced.

package.json
brizental: No need to settle in this PR, I just wanted to point it out.

.eslintrc
Dexterp37: "Why don't we use tabs instead"?

copyright.txt
Dexterp37: Good point :P

package.json
Dexterp37: I don't have a strong opinion, I just want us to be explicit. Let it be `glean`, then. I don't think the Glean SDK is going to release another package on NPM... right?

brizental: I don't expect it will, no :)

src/index.ts
brizental: Because this is TypeScript and it uses ES6 import / export syntax. See: https://www.typescriptlang.org/docs/handbook/modules.html#export-statements

Dexterp37 (approved):

webpack.config.browser-static.js
brizental: Hm, this is a bug in our eslint configuration. This rule should only be enforced for *.ts files. Nice catch!

PR #3, opened 2020-11-10
Bug 1676143 - Add web extension sample

brizental:
BONUS: housekeeping.


samples/web-extension/README.md
```diff
@@ -0,0 +1,30 @@
+# Web extension sample sample
+
+1. Build Glean.js for the browser. On the root folder of this repository run:
```
Dexterp37, line 3: Can you clarify what "browser" means here?

samples/web-extension/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "web-extension",
+  "version": "1.0.0",
+  "description": "A sample web extension using Glean.js",
+  "main": "index.js",
```
Dexterp37, line 5: Do we need really webpack for this example?

samples/web-extension/README.md
brizental: Browser here means the environment. Regardless if it is a webextension or a a web app static or other.

samples/web-extension/package.json
brizental: It's more straight forward than vendoring the Glean.js script, IMO.

samples/web-extension/README.md
Dexterp37: Web extensions are allowed to use different APIs than webpages. Who's responsible for checking what's used where?

samples/web-extension/package.json
Dexterp37: Got it. This is needed because webextensions can only include files that are stored as children of the directory where the manifest file is. So we either copy the glean distribution here as a pre-build step, or use a bundler.

Any chance you could squeeze this explanation in the README?

samples/web-extension/README.md
```diff
@@ -0,0 +1,30 @@
...
+3. Load the web extension on your browser of choice.
+
+  - **Firefox**
+    1. Go to [about:debugging#/runtime/this-firefox](about:debugging#/runtime/this-firefox);
+    2. Click on `Load Temporary Add-on...`;
+    3. Choose the [`manifest.json`](./manifest.json) file;
+    4. Click on `Inspect`;
+    5. Try clicking on the little Glean logo that should have been added to your browsers toolbar.
+
+  - **Chrome**
```
Dexterp37, line 31:
```suggestion
  - **Chromium-based browsers**
```

samples/web-extension/README.md
brizental: Yes, you are right about that. For now though, it hasn't been necessary to split the build between web extension and web apps, so I propose we cross that bridge when we get there.

Dexterp37: While we could tackle this going forward, since we already know APIs will differ, I believe we're better suited to change and clarify our comments while our codebase is small, setting the right track for others to understand if they want to contribute as well

samples/web-extension/package.json
Dexterp37: I think this is still missing

Dexterp37 (changes requested):

samples/web-extension/package.json
brizental: I added a small note on the README.md https://github.com/brizental/glean.js/pull/3/commits/e43f97282e74546b9b73bb002483e622a4e69119, let me know if you think it is not enough.

Dexterp37: It doesn't explain _why_ we're doing that (the reason that I talked about 2 comments above)

brizental: Ah, I see. Let me add that :)

samples/web-extension/README.md
brizental: My point in thinking it's best that we only bifurcate these build configurations when we need to, is related to the fact that I am not even sure we are going to need to.

On the proof of concept we did not see the need for two different builds for websites and web extensions and in fact the only moment we needed to use specific web extensions APIs was [when getting the application id](https://github.com/brizental/gleanjs/blob/main/src/glean.js#L16-L17), which now we have the user pass explicitly through the `initialize` function, making even that use case unnecessary.

I thought maybe to get platform information we'd need different logic, but web extensions, i just verified, have access to the `navigator.userAgent` property, so they can share code with websites in this regard too.

Dexterp37 (approved):

samples/web-extension/README.md
Dexterp37: All right,let's stick to the current naming and change this if it becomes confusing to others in the future.

PR #6, opened 2020-12-02
Bug 1679375 - Creates a Store interface and implements a WeakStore

brizental:
This creates a generic interface to be implemented by all our stores, which will be:

- The different persistent stores that will need to be implemented per platform
- A weak store that can be used by all platforms

The interface follows the store design described in [Glean.js: Client-side storage for web extensions](https://docs.google.com/document/d/1zHy5NBcH59u_mDEFTToUPBwtV_jOAWGANJtrMOxBFBs/edit?usp=sharing). Applying the suggestion of consolidating `create`/`update` into a single function and renaming `read` to `get` (just sounded nicer, IMO).

Although the functions and their proposed functionalities are basically the same as in the proposal, the function signatures deviate from it. That was necessary, because the signatures proposed did not take into account the fact that our stores may contain multi-level objects.

This PR contains only the WeakStore implementation, the store which will not persist data throughout application runs. This implementation could also be used in the future to serve as a mock implementation on the persistent store for tests as well.

Finally, it also contains a bit of housekeeping work, namely:

- Added generated files to eslintignore;
- Added some more properties to tsconfig;
- Have CircleCI refrain from running the size check job on main (it fails on main and ultimately is unnecessary there).


brizental:
CI failures are unrelated to the bulk of this PR and mean bugs on the webpack configuration for aliases. I'll deal with that, but there is no need to wait for CI to be green to do the review, if you want to :)

brizental:
# Build size report

  Merging https://github.com/brizental/glean.js/pull/6 into [main](https://github.com/brizental/glean.js) will:

  * **Increase** the size of the browser build (`npm run build:browser`) by `0%`.
  * **Increase** the size of the Qt build (`npm run build:qt`) by `0%`.

  ---

  | Build | Current size | New size | Size increase |
  |--:|:---:|:---:|:---:|
  | browser | 4.44K | 4.44K | 📈 0% |
  | qt | 3.39K | 3.39K | 📈 0% |

src/storage/weak.ts
```diff
@@ -0,0 +1,146 @@
...
+  private store: StorageObject;
+
+  constructor() {
+    this.store = {};
+  }
+
+  /**
+   * Gets the value recorded to the given index on the store.
+   *
+   * @param index [optional] The index  of the entry we want to delete.
```
Dexterp37, line 22: I think this comment is wrong :)

src/storage/weak.ts
```diff
@@ -0,0 +1,146 @@
...
+   *
+   * @returns The value found for the given index on the storage.
+   *          If no index is given, the whole store is returned.
+   */
+  get(index?: StorageIndex): Promise<StorageValue> {
+    return new Promise(resolve => {
+      if (!index) return this.store;
```
Dexterp37, line 29: Do you have a solid use-case for this? I'd rather remove this ability unless we know it's really needed.

src/storage/weak.ts
```diff
@@ -0,0 +1,146 @@
...
+
+      let target: StorageValue = this.store;
+      for (const key of index) {
+        if (typeof target !== "undefined" && typeof target !== "string" && key in target) {
```
Dexterp37, line 33:
```suggestion
        if (target !== undefined && typeof target !== "string" && key in target) {
```

I was recently bitten by checking against a string, since I had a typo in that string.

src/storage/weak.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  /**
+   * Gets the value recorded to the given index on the store.
+   *
+   * @param index [optional] The index  of the entry we want to delete.
+   *
+   * @returns The value found for the given index on the storage.
+   *          If no index is given, the whole store is returned.
+   */
+  get(index?: StorageIndex): Promise<StorageValue> {
```
Dexterp37, line 27: Any reason to not make this `async get` since we're always returning a promise?

src/storage/weak.ts
```diff
@@ -0,0 +1,146 @@
...
+    return new Promise(resolve => {
+      if (!index) return this.store;
+
+      let target: StorageValue = this.store;
+      for (const key of index) {
+        if (typeof target !== "undefined" && typeof target !== "string" && key in target) {
+          target = target[key];
+        } else {
+          resolve();
```
Dexterp37, line 36: Can you please add a comment stating that we're bailing out because the "full target path does not exist"?

src/storage/weak.ts
```diff
@@ -0,0 +1,146 @@
...
+  constructor() {
+    this.store = {};
+  }
+
+  /**
+   * Gets the value recorded to the given index on the store.
+   *
+   * @param index [optional] The index  of the entry we want to delete.
+   *
+   * @returns The value found for the given index on the storage.
```
Dexterp37, line 24: It's worth mentioning that this will always resolve and never reject, even for non-existing paths. Are you sure you want this type of behaviour? Wouldn't it be valuable to know if something doesn't exist?

src/storage/weak.ts
```diff
@@ -0,0 +1,146 @@
...
+      resolve(target);
+    });
+  }
+
+  /**
+   * Updates a specific entry from the store.
+   *
+   * # Rejects
+   *
+   * - In case the index in an empty array.
```
Dexterp37, line 49: Shouldn't this be part of a `@returns` docstring?

src/storage/weak.ts
```diff
@@ -0,0 +1,146 @@
...
+   *
+   * ``json
+   * {
+   *  "foo": {
+   *    "bar": "new value!"
+   *  }
+   * }
+   * ```
+   *
+   * @param index The index  of the entry we want to delete.
```
Dexterp37, line 74: Since we're being good citizens, what about signaling the type of the parameter in the `@param` (here and in other places)?

src/storage/weak.ts
```diff
@@ -0,0 +1,146 @@
...
+      if (index.length === 0) {
+        reject(Error("The index must contain at least one property to update."));
+      }
+
+      let target = this.store;
+
+      // Loops through all the keys but the last.
+      // The last is the key we will **set**, not **get**.
+      for (const key of index.slice(0, index.length - 1)) {
+        if (typeof target[key] == "undefined" || typeof target[key] == "string") {
```
Dexterp37, line 91:
```suggestion
        if (target[key] === undefined || typeof target[key] == "string") {
```

Dexterp37: Why do we need to check if `target[key]` is a string and set it to `{}`?

src/storage/index.ts
```diff
@@ -0,0 +1,90 @@
...
+export type StorageValue = undefined | string | StorageObject;
+export interface StorageObject {
+  [key: string]: StorageValue;
+}
+
+export interface Store {
+  /**
+   * Gets the value recorded to the given index on the store.
+   *
+   * @param index [optional] The index  of the entry we want to delete.
```
Dexterp37, line 38: Wrong comment :)

src/storage/index.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  /**
+   * Updates a specific entry from the store.
+   *
+   * # Rejects
+   *
+   * - In case the index in an empty array.
+   *
+   * # Note
+   *
```
Dexterp37, line 53: See my comments below for how to change this.

src/storage/index.ts
```diff
@@ -0,0 +1,90 @@
...
+  update(
+    index: StorageIndex,
+    transformFn: (v: StorageValue) => Exclude<StorageValue, undefined>
+  ): Promise<void>;
+
+  /**
+   * Deletes a specific entry from the store.
+   *
+   * @param index The index of the entry we want to delete.
+   *              If given an empty index, will delete all entries on the store.
```
Dexterp37, line 87: Do we want to indent to the level of `index` here? Or does the linter complain?

Dexterp37 (changes requested):

src/storage/weak.ts
brizental: LOL, yeap it is.

src/storage/weak.ts
brizental: You'll know because StorageValue may be `undefined`. When it is undefined, taht means it doesn't exist. Do you think error is better than that?

brizental: Regarding mentioning it can never reject, yeah, that is good, I'll do that.

src/storage/weak.ts
brizental: No reason to make it async since we are not using the `await` keyword inside it and we **always** return a promise.

src/storage/weak.ts
brizental: No solid use case, but I do use it wuite a lot for tests.

src/storage/weak.ts
brizental: Heh, but we are still checking agains a string right next to this first condition!

I check for `undefined` enough times that it may be good to acutally have a helper function such as `isUndefined` to check it instead of re-writing this everytime. Would you mind if I did that as a follow up though?

src/storage/weak.ts
brizental: You are right. I see also this docstring doesn't even have a @returns part, which it definitely should. I should audit all these comments... Its also a bit annoying to copy paste them for every new implementation of Store. Would you be against just having these comments on the interface declaration?

src/storage/weak.ts
brizental: I was following jan-eriks advice of not adding types in the docstring when the types are already in the signature of the function. Do you think that is not enough?

(geez, this param comment also says `delete`...)

src/storage/weak.ts
brizental: All of the indexes (but the last one) are expected to contain an object. If it doesn't (the other possible values are `undefined` and `string` as defined in the `StorageValue` type) then we overwrite that with an object. This behaviour is also described in the docstring.

src/storage/index.ts
brizental: Wrong comments everywhere!!!!!!

src/storage/index.ts
brizental: > Do we want to indent to the level of index here?

What do you mean?

Linter is very forgiving at this point, doesn't really complain about much, heh.

src/storage/weak.ts
brizental: I just verified and hover over still works when we have only the comment on the interface, FWIW.

Screenshot 2020-12-02 at 17 20 40

src/storage/weak.ts
Dexterp37: Cool, works for me.

src/storage/weak.ts
Dexterp37: Would a helper function be more concise than `obj === undefined`? I'm not sure we need a helper, but if you really want to add one, please add it in a follow-up.

src/storage/weak.ts
Dexterp37: I disagree. I think we should at the very least use `async` because it makes it clearer how the function behaves just looking at the signature:, especially after building and when we'll have to deal with sad JS alone.

Moreover, `async` gives the option at the callsite to decide whether to block or not, easily, while using promises makes it a bit harder.

src/storage/weak.ts
Dexterp37: That's perfect! Let's just have them on the interface!

src/storage/weak.ts
Dexterp37: Can't we check for `StorageValue` type directly?

src/storage/index.ts
Dexterp37:
```js
   * @param index The index of the entry we want to delete.
   *        If given an empty index, will delete all entries on the store.
```

I mean aligning with `index` instead of to `The`

src/storage/weak.ts
Dexterp37: > 
> 
> You'll know because StorageValue may be `undefined`. When it is undefined, taht means it doesn't exist. Do you think error is better than that?

If we can store `undefined` (e.g. `asd.lol.rotfl` is a proper key with `undefined` as a value) then returning `undefined` can mean two things: (1) the path doesn't exist or (2) the path exists but stores `undefined`.

If we cannot store `undefined`, then it's ok to not throw an error.

Regardless of what you choose, we should improve the documentation.

src/storage/weak.ts
Dexterp37: Ok, I fear this could potentially be error-prone in the future. If you're using it just for test, would you consider one of the following:

- add a test only function?
- change the test to always use a key?

src/storage/weak.ts
brizental: > Moreover, async gives the option at the callsite to decide whether to block or not, easily, while using promises makes it a bit harder.

In pratice, adding the async keyword wraps the whole return value in a Promise, which is what I am doing manually here. I can totally change the signature to have the async keyword which would allow me to remove all the `return new Promise(...` shenanigans, but the way this function is right now it can already be called which an `await` keyword. That is exactly what I do in tests. You can `await` **any** promise, not only on the return values of `async` functions.

src/storage/weak.ts
brizental: > If we can store undefined (e.g. asd.lol.rotfl is a proper key with undefined as a value) then returning undefined can mean two things: (1) the path doesn't exist or (2) the path exists but stores undefined.

Actually we can't. At least not if we stick to using the provided API for the Store. The `update` function doesn't allow an update to an `undefined` value, see: 

```ts
update(
    index: StorageIndex,
    transformFn: (v: StorageValue) => Exclude<StorageValue, undefined>
  ): Promise<void>;
```

`Exclude<StorageValue, undefined>` means `transformFn` may return any valid `StorageValue`, excluding `undefined`.

> Regardless of what you choose, we should improve the documentation.

I agree. Even though I re-read this docstrings multiple times I see in the PR that they have many errors and missing infos 👎

src/storage/weak.ts
Dexterp37: I believe that `async` gives more clarity about the semantics of the function and should be used instead. That's especially true when getting outside of typescript land back into the JS land (for example when debugging in the browser). Unless there's a **strong reason** not to do that, I'd advocate changing to `async`.

src/storage/weak.ts
Dexterp37: > > Regardless of what you choose, we should improve the documentation.
> 
> I agree. Even though I re-read this docstrings multiple times I see in the PR that they have many errors and missing infos 👎

:-) Let's retain the current behaviour then and just fix the docs :-)

src/storage/weak.ts
brizental: > Can't we check for StorageValue type directly?

Sadly, we can't _just_ do that. `StorageValue` only exists in Typescript and will go away once this is compiled do JS, so we can't just check against it. What we can do is write a helper function that checks if something is a valid `StorageValue`, I actually did that for the web-ext `Store` implementation, maybe I should just go ahead and make that a utility for the `storage` module...

It looks something like this:

```ts
function isStorageValue(v: browser.storage.StorageValue): v is StorageValue;
```

Specifically this is a [User defined type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards). In pratice its a function that returns a boolean, but because the signature has this special return value Typescript will trust that `v` is indeed that type if it return `true`.

src/storage/index.ts
brizental: Ah! Yeah I can do that, I have no preference. I'll see if I can also add a eslint rule to reviry against this so we are sure all docstrings are styled the same :)

src/storage/weak.ts
brizental: Ok, I'll change this to error in case index is empty or undefined. And go ahead with 

> change the test to always use a key?

Which is simpler I think.

src/storage/weak.ts
Dexterp37: :-(

src/storage/weak.ts
brizental: Yeap, I agree. I'll change it :)

src/storage/weak.ts
brizental: Just for the record, if we allow consumers to use the storage API directly somehow, they will be able to update to `undefined` if they are coming from JS. See: [TS playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAaghgGwK7QLxQM7AE4EsB2A5lAD5RL4AmEAZgRJaVAEoQDGA9tpQDxZ5EANOXwBrfBwDu+AHwBuAFAKaFNsFwd8UbBQAUwDswoAuKLoBup+MggBKKKhlQAogA82yaj2sphFanT4DDL2AN4KUNoQwEjYWgZG+LoARJIAFnAxYMm2igC+QA).

All the type guarantees go away once in JSland. Now, I this is meant as an internal module and interally we only use Typescript. Still, we should probably keep this is mind for other stuff we implement, specially consumer facing...

src/storage/index.ts
brizental: Did not find a way to add a specific rule o check this, but I did add jsdoc rules to our eslint config, commit coming shortly.

brizental:
# Build size report

  Merging https://github.com/brizental/glean.js/pull/6 into [main](https://github.com/brizental/glean.js) will:

  * **Increase** the size of the browser build (`npm run build:browser`) by `0%`.
  * **Increase** the size of the Qt build (`npm run build:qt`) by `0%`.

  ---

  | Build | Current size | New size | Size increase |
  |--:|:---:|:---:|:---:|
  | browser | 4.75K | 4.75K | 📈 0% |
  | qt | 3.69K | 3.69K | 📈 0% |

src/storage/weak.ts
brizental: Wasn't simpler, I added a test only function.

src/index.ts
```diff
@@ -2,6 +2,11 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+// Importing this here just so the size increase will show on the PR comments,
+// once everything is implemented we remove it.
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import * as storage from "storage";
```
brizental, line 8: Comment turned out to be wrong, it is saying increase is zero. But also this import is wrong, I should add another one right below it `import WeakStorage from storage/weak`... I'll file a bug about the comment and remove this changes from this Pr since its does nothing ATM.

package.json
```diff
@@ -4,7 +4,8 @@
   "description": "An implementation of the Glean SDK, a modern cross-platform telemetry client, for Javascript environments.",
   "main": "./dist/glean.js",
   "scripts": {
-    "test": "mocha",
+    "test": "ts-mocha --paths -p ./tsconfig.json",
+    "test:debug": "ts-mocha --paths -p ./tsconfig.json --inspect-brk",
```
Dexterp37, line 8: Out of curiosity: why `test:debug` and not, for example, `test-debug`?

src/storage/index.ts
```diff
@@ -0,0 +1,99 @@
...
+
+export interface Store {
+  /**
+   * **Test-only API**
+   *
+   * Gets the whole store.
+   *
+   * @returns An object representing all the data recorded in the store.
+   */
+  _testOnly_getWholeStore(): Promise<StorageObject>;
```
Dexterp37, line 42: Do you need this on the interface, or do you really just need this on the object implementing the interface?

Dexterp37 (approved):
r+wc!

package.json
brizental: No reason at all, preference and mostly that is what I am used to from other projects. Would you rather I change? No big deal for me at all, but I'd rather do this sort of changes now, if we are going to :)

Dexterp37: No no, that's fine! I was asking because we used `test-something` in Rally and I was curious about whether or not using `:` was recommended and I was doing it wrong :) Nothing to do here!

src/storage/index.ts
brizental: I'd say on the interface. We want the interface to behave the same for all Stores. Making even the test interface standardized makes it easy to run the same tests in all stores and consequently write the same test suites for every implementation.

PR #7, opened 2020-12-04
Bug 1679375  - Add tests for web ext storage impl

brizental:
...and have the unit tests run directly on Firefox (or try)

[This is draft, full description will come later]

[UPDATE]

From the commit message:

> In order to access the web extensions APIs directly on
> the browser we need to play a bit of a telephone game.
>
> We use selenium to instantiate and run code as if on the
> browser console, but on the console we still don't have access
> to the webextensions APIs neither can we send messages directly
> the a web extensions background script.

It occured to me that I may be wrong about not being able to directly message the background script, but such a change, if confirmed, could be done in a follow up. It would remove a bit of the complexity, by making the content script unnecessary.

> With this in mind, we send a custom event ("test") from the console,
> which will contain information about the exact web extention API we
> want to access and with what arguments we want to execute it.
>
> Packed with this commit is a webextension that I created to help with
> these tests. The content script of this web extension listens for the
> "test" custom events. Whenever a new one is received, the content
> script, which still doesn't have access to most web ext APIs, will pass
> the message from the website to the background script.
>
> The background script of this webextension, is listening for such
> messages, and will process the message and execute the requested API.
> When that is done, the background script will pass the response back to
> the content script. The content script will send the response back in
> another custom event ("testResponse").
>
> The tests will listen for "testResponse" events, when this event is
> received we step out of the executeAsyncScript block and return the
> response received.
>
> All of this will happen inside a proxy around the desired web ext APIs,
> so that calling code doesn't even need to know all that is happening and
> can hapilly call web ext APIs without worrying.

PR #8, opened 2020-12-07
Bug 1679375 - Implement a Storage that uses the web ext storage API

brizental:
This is a split off of #7, containing only the storage implementation for web extensions. This doesnt contain tests as I am still figuring out last issue with testing direcly on browsers. This might mean that there are bugs on this implementation, but since I have done manual testing, no big changes should come on the tests PR and the overall approach will still be the same.

### DIFF
--- a/bin/build-size-check.sh
+++ b/bin/build-size-check.sh
@@ -19,6 +19,7 @@ webext_size_pretty=$(wc -c ./dist/glean.js | awk '{printf "%0.2f\n",$1/1024"."su
 git branch -f original-main origin/main
 # TODO: This is done in case there were changes to the package.json,
 # we should find a better way to deal with that though.
+# See: [Bug 1681484](https://bugzilla.mozilla.org/show_bug.cgi?id=1681484)
 git reset --hard HEAD
 git checkout original-main
 
--- a/bin/build-size-check.sh
+++ b/bin/build-size-check.sh
@@ -16,7 +16,11 @@ npm run build:webext
 webext_size=$(wc -c ./dist/glean.js | awk '{print $1}')
 webext_size_pretty=$(wc -c ./dist/glean.js | awk '{printf "%0.2f\n",$1/1024"."substr($2,1,2)}')
 
-git checkout -b original-main origin/main
+git branch -f original-main origin/main
+# TODO: This is done in case there were changes to the package.json,
+# we should find a better way to deal with that though.
+git reset --hard HEAD
+git checkout original-main
 
 npm install
 
--- a/bin/build-size-check.sh
+++ b/bin/build-size-check.sh
@@ -12,11 +12,11 @@ npm run build:qt
 qt_size=$(wc -c ./dist/glean.js | awk '{print $1}')
 qt_size_pretty=$(wc -c ./dist/glean.js | awk '{printf "%0.2f\n",$1/1024"."substr($2,1,2)}')
 
-npm run build:browser
-browser_size=$(wc -c ./dist/glean.js | awk '{print $1}')
-browser_size_pretty=$(wc -c ./dist/glean.js | awk '{printf "%0.2f\n",$1/1024"."substr($2,1,2)}')
+npm run build:webext
+webext_size=$(wc -c ./dist/glean.js | awk '{print $1}')
+webext_size_pretty=$(wc -c ./dist/glean.js | awk '{printf "%0.2f\n",$1/1024"."substr($2,1,2)}')
 
-git checkout -t origin/main
+git checkout -b original-main origin/main
 
 npm install
 
@@ -24,32 +24,32 @@ npm run build:qt
 qt_size_main=$(wc -c ./dist/glean.js | awk '{print $1}')
 qt_size_main_pretty=$(wc -c ./dist/glean.js | awk '{printf "%0.2f\n",$1/1024"."substr($2,1,2)}')
 
-npm run build:browser
-browser_size_main=$(wc -c ./dist/glean.js | awk '{print $1}')
-browser_size_main_pretty=$(wc -c ./dist/glean.js | awk '{printf "%0.2f\n",$1/1024"."substr($2,1,2)}')
+npm run build:webext
+webext_size_main=$(wc -c ./dist/glean.js | awk '{print $1}')
+webext_size_main_pretty=$(wc -c ./dist/glean.js | awk '{printf "%0.2f\n",$1/1024"."substr($2,1,2)}')
 
 qt_diff=$(((($qt_size-$qt_size_main)%$qt_size_main)*100))
-browser_diff=$(((($browser_size-$browser_size_main)%$browser_size_main)*100))
+webext_diff=$(((($webext_size-$webext_size_main)%$webext_size_main)*100))
 
 [[ $qt_diff -ge 0 ]] && qt_emoji="📈" || qt_emoji="📉"
 [[ $qt_diff -ge 0 ]] && qt_result="Increase" || qt_result="Decrease"
 
-[[ $browser_diff -ge 0 ]] && browser_emoji="📈" || browser_emoji="📉"
-[[ $browser_diff -ge 0 ]] && browser_result="Increase" || browser_result="Decrease"
+[[ $webext_diff -ge 0 ]] && webext_emoji="📈" || webext_emoji="📉"
+[[ $webext_diff -ge 0 ]] && webext_result="Increase" || webext_result="Decrease"
 
 content="
   # Build size report
 
   Merging $CIRCLE_PULL_REQUEST into [main](https://github.com/brizental/glean.js) will:
 
-  * **$browser_result** the size of the browser build (\`npm run build:browser\`) by \`${browser_diff}%\`.
+  * **$webext_result** the size of the webext build (\`npm run build:webext\`) by \`${webext_diff}%\`.
   * **$qt_result** the size of the Qt build (\`npm run build:qt\`) by \`${qt_diff}%\`.
 
   ---
 
   | Build | Current size | New size | Size increase |
   |--:|:---:|:---:|:---:|
-  | browser | ${browser_size_main_pretty}K | ${browser_size_pretty}K | $browser_emoji ${browser_diff}% |
+  | webext | ${webext_size_main_pretty}K | ${webext_size_pretty}K | $webext_emoji ${webext_diff}% |
   | qt | ${qt_size_main_pretty}K | ${qt_size_pretty}K | $qt_emoji ${qt_diff}% |
 "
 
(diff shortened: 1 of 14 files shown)


src/storage/index.ts
```diff
@@ -31,6 +33,19 @@ export interface StorageObject {
   [key: string]: StorageValue;
 }
 
+export function isStorageValue(v: unknown): v is StorageValue {
```
Dexterp37, line 36: Can you document what's the purpose of this?

src/storage/index.ts
```diff
@@ -31,6 +33,19 @@ export interface StorageObject {
...
+  if (isUndefined(v) || isString(v)) {
+    return true;
+  } else if (isObject(v)) {
+    for (const key in v) {
+      return isStorageValue(v[key]);
+    }
+  } else {
+    return false;
+  }
+  return true;
```
Dexterp37, line 46: Do we ever get here?

src/storage/index.ts
```diff
@@ -31,6 +33,19 @@ export interface StorageObject {
...
+export function isStorageValue(v: unknown): v is StorageValue {
+  if (isUndefined(v) || isString(v)) {
+    return true;
+  } else if (isObject(v)) {
+    for (const key in v) {
+      return isStorageValue(v[key]);
+    }
+  } else {
+    return false;
+  }
```
Dexterp37, line 45:
```suggestion
  if (isUndefined(v) || isString(v)) {
    return true;
  } 
  
  if (isObject(v)) {
    for (const key in v) {
      return isStorageValue(v[key]);
    }
  }

  return false;
```

This seems slightly more readable. What do you think?

src/storage/persistent/webext.ts
```diff
@@ -0,0 +1,121 @@
...
+import { Store, StorageIndex, StorageValue, StorageObject, isStorageValue } from "storage";
+import { isObject, isString, isUndefined } from "utils";
+
+/**
+ * Persistent storage implementation based on the Promise-based
+ * [storage API](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage)
+ * for web extensions.
+ *
+ * To make sure this implementation works on Chrome, the user must install the peer dependency
+ * [`mozilla/webextension-polyfill`](https://github.com/mozilla/webextension-polyfill).
```
Dexterp37, line 14: Do we need this here? Should we move this at README level? It will apply to everything that uses `browser`. Moreover, this has less to do with Chrome and more to do with "Chromium-based browsers"

src/storage/persistent/webext.ts
```diff
@@ -0,0 +1,121 @@
...
+    } else {
+      throw new Error(`Unexpected value found in storage for index ${index}. Ignoring.\n${target}`);
+    }
+  }
+
+  async update(
+    index: StorageIndex,
+    transformFn: (v: StorageValue) => Exclude<StorageValue, undefined>
+  ): Promise<void> {
+    debugger;
```
Dexterp37, line 82: stray ?

Dexterp37 (changes requested):

src/storage/index.ts
brizental: Not that I know of, but Typescript complains about it. It gives an error that goes more or less "can't return undefined if return value is bool" /o\

src/storage/persistent/webext.ts
brizental: Yeap good point. I'll move it.

src/storage/index.ts
Dexterp37: Does reshaping it as recommended in the other comment help?

brizental: Yes, it does :)

src/storage/persistent/webext.ts
brizental: So, I was going to make this change, but as is, our README.md is not really meant for this type of information. 

On the proposal we talked about having Glean.js information on the Glean Book, in practice I am not really sure how that will work yet. For brevity's sake, how about we leave this here for now (since this is the only code that it will appy to ATM, anyways) and deal with the whole documentation issue later? 

I'll make sure to file a follow up to define how we will merge Glean.js with Glean SDK documentation, if you agree.

brizental: I'll still leave the change about "Chromium based browsers"

brizental:
The check-size job will fail because it will try to run the new build:webext command on main which doesn't have it yet. I propose we ignore that failure for now.

bin/build-size-check.sh
```diff
@@ -16,7 +16,11 @@ npm run build:webext
 webext_size=$(wc -c ./dist/glean.js | awk '{print $1}')
 webext_size_pretty=$(wc -c ./dist/glean.js | awk '{printf "%0.2f\n",$1/1024"."substr($2,1,2)}')
 
-git checkout -b original-main origin/main
+git branch -f original-main origin/main
+# TODO: This is done in case there were changes to the package.json,
+# we should find a better way to deal with that though.
```
brizental, line 21: Self-reminder to file a follow-up for this.

Dexterp37: Let's file a bug and add the bug number here, please :)

src/index.ts
```diff
@@ -6,6 +6,13 @@
 // once everything is implemented we remove it.
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 import StorageWeak from "storage/weak";
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import StoragePersistent from "storage/persistent";
+
+console.log(
+  StorageWeak,
+  StoragePersistent
+);
```
Dexterp37, line 15: Stray log?

src/storage/index.ts
```diff
@@ -26,11 +28,32 @@ export type StorageIndex = string[];
 /**
  * The possible values to be retrievd from storage.
  */
-export type StorageValue = undefined | string | StorageObject;
+export type StorageValue = null | undefined | string | StorageObject;
```
Dexterp37, line 31: Why do we need `null` here, too?

webpack.config.qt.js
```diff
@@ -4,12 +4,21 @@
...
 module.exports = {
   ...baseConfig,
   output: {
     ...baseConfig.output,
     libraryTarget: "var",
+  },
+  resolve: {
+    ...baseConfig.resolve,
+    alias: {
+      // TODO: This is temporary until we actually have a persistent storage impl for Qt.
```
Dexterp37, line 20: Let's file a bug about this and add the bug number here please.

Dexterp37 (changes requested):

src/storage/index.ts
brizental: This was a workaround for me to be able to create a query that has `null` as default for the web ext storage. When running the tests this morning I actually realized this causes a bug, I'll send a commit removing this, shortly.

src/index.ts
brizental: Oops.

brizental: Hm, sorry actually there are not stray. If I leave the imports unused I realized webpacks doesn't include them in the final bundle. I will leave a comment about that though, to be clear.

src/storage/persistent/webext.ts
```diff
@@ -54,38 +56,36 @@ class WebExtStore implements Store {
...
     const query = this._buildQuery(index);
-    const response = await this.store.get(query);
+    const response: browser.storage.StorageObject = await this.store.get(query);
     if (!response) {
       return;
     }
 
     if (isStorageValue(response)) {
-      if (isObject(response)) {
+      if (!isUndefined(response) && !isString(response)) {
```
brizental, line 78: Typescript complains about not being sure that `response` is not null with only the `isObject` check... `isUndefined` + `isString` do the trick though... go figure.

src/storage/persistent/webext.ts
```diff
@@ -4,7 +4,9 @@
 
 import { Store, StorageIndex, StorageValue, StorageObject, isStorageValue } from "storage";
 import { updateNestedObject, getValueFromNestedObject, deleteKeyFromNestedObject } from "storage/utils";
-import { isObject } from "utils";
+import { isString, isUndefined } from "utils";
+
+type WebExtStoreQuery = { [x: string]: { [x: string]: null; } | null; };
```
brizental, line 9: This is what I was trying to avoid with adding `null` to StorageValue. No surprise that introduces bugs since it then allows users to save `null` values to the Storage.

src/storage/persistent/webext.ts
```diff
@@ -54,38 +56,36 @@ class WebExtStore implements Store {
...
    *
-   * @param index The index to the given entry on the storage.
    * @param transformFn The transformation function to apply to the store.
    *
    * @returns The query object with the modified store.
    */
-  private async _buildQueryFromStore(index: StorageIndex, transformFn: (s: StorageObject) => StorageObject) {
+  private async _buildQueryFromStore(transformFn: (s: StorageObject) => StorageObject): Promise<StorageObject> {
     const store = await this._testOnly_getWholeStore();
-    if (isStorageValue(store) && isObject(store)) {
```
brizental: No need to check here since _testOnly_getWholeStore already returns a StorageObject.

src/storage/persistent/webext.ts
```diff
@@ -54,38 +56,36 @@ class WebExtStore implements Store {
...
+   * Retrieves the full store and builds a query object on top of it.
    *
-   * @param index The index to the given entry on the storage.
    * @param transformFn The transformation function to apply to the store.
    *
    * @returns The query object with the modified store.
    */
-  private async _buildQueryFromStore(index: StorageIndex, transformFn: (s: StorageObject) => StorageObject) {
+  private async _buildQueryFromStore(transformFn: (s: StorageObject) => StorageObject): Promise<StorageObject> {
     const store = await this._testOnly_getWholeStore();
```
brizental, line 66: Question, is it to terrible to use this test function inside the actual code. I wondered about changing its name to _getWholeStore, but then it wouldn't be clear it is supposed to be used for tests?

src/index.ts
Dexterp37: oh, that's interesting. Can you file a follow-up bug to investigate this? I'm fairly sure there's some option to make Webpack not do that.

Dexterp37: Let's add the bug number to this commet please.

src/storage/persistent/webext.ts
Dexterp37: It is terrible :) Make it `_getWholeStore` if you're relying on it internally.

Dexterp37 (approved):
r+wc

src/index.ts
brizental: I am not sure it would be worth the effort looking into this. I am only doing it now so that these modules are added to the final bundle and the size increase is added to the size comment, so that we keep track. All of this goes away when we are done with milestone 1 and overall not including unused imports to the bundle is IMO the preferred behaviour in the long run.

PR #9, opened 2020-12-10
Bug 1679375 - Implement the base of the metrics database module

brizental:
As was described in the proposal, the database module will differ from the glean-core database module.

While it will serve as an abstraction layer over the underlying storage, it will also be the interface to retrive the ping / metric payload data from the storage. This is a bit easier for Glean.js, because the data is saved already in the format that is expected by the ping schema.

I will leave comments on the code about some doubts that lingered after I was finished with this implementation.

### DIFF
--- a/src/database.ts
+++ b/src/database.ts
@@ -99,7 +99,6 @@ class Database {
    * @param value The value we want to record to the given metric.
    */
   async record(metric: Metric, value: string): Promise<void> {
-    debugger;
     if (metric.disabled) {
       return;
     }
--- a/src/database.ts
+++ b/src/database.ts
@@ -17,6 +17,8 @@ export interface PingPayload {
  * Verifies if a given value is a valid PingPayload.
  *
  * @param v The value to verify
+ *
+ * @returns Whether or not `v` is a valid PingPayload.
  */
 export function isValidPingPayload(v: StorageValue): v is PingPayload {
   if (isObject(v)) {
@@ -62,22 +64,9 @@ class Database {
   private appStore: Store;
 
   constructor() {
-    this.userStore = new PersistentStore("user");
-    this.pingStore = new PersistentStore("ping");
-    this.appStore = new PersistentStore("app");
-  }
-
-  /**
-   * The key under which to save a metrics value.
-   *
-   * @param metric The metric for which to build the storage key
-   */
-  private _buildStorageKey(metric: Metric): string {
-    if (metric.category) {
-      return `${metric.category}.${metric.name}`;
-    } else {
-      return metric.name;
-    }
+    this.userStore = new PersistentStore("userLifetimeMetrics");
+    this.pingStore = new PersistentStore("pingLifetimeMetrics");
+    this.appStore = new PersistentStore("appLifetimeMetrics");
   }
 
   /**
@@ -111,11 +100,13 @@ class Database {
    */
   async record(metric: Metric, value: string): Promise<void> {
     if (!metric.disabled) {
-      const store = this._chooseStore(metric.lifetime);
-      const storageKey = this._buildStorageKey(metric);
-      for (const ping of metric.sendInPings) {
-        await store.update([ping, metric.type, storageKey], () => value);
-      }
+      return;
+    }
+
+    const store = this._chooseStore(metric.lifetime);
+    const storageKey = metric.identifier;
+    for (const ping of metric.sendInPings) {
+      await store.update([ping, metric.type, storageKey], () => value);
     }
   }
 
@@ -128,17 +119,19 @@ class Database {
    */
   async transform(metric: Metric, transformFn: (v?: string) => string): Promise<void> {
     if (!metric.disabled) {
-      const store = this._chooseStore(metric.lifetime);
-      const storageKey = this._buildStorageKey(metric);
-      for (const ping of metric.sendInPings) {
-        const finalTransformFn = (v: StorageValue): string => {
-          if (isObject(v)) {
-            throw new Error(`Unexpected value found for metric ${metric}: ${JSON.stringify(v)}.`);
-          }
-          return transformFn(v);
-        };
-        await store.update([ping, metric.type, storageKey], finalTransformFn);
-      }
+      return;
+    }
+
+    const store = this._chooseStore(metric.lifetime);
+    const storageKey = metric.identifier;
+    for (const ping of metric.sendInPings) {
+      const finalTransformFn = (v: StorageValue): string => {
+        if (isObject(v)) {
+          throw new Error(`Unexpected value found for metric ${metric}: ${JSON.stringify(v)}.`);
+        }
+        return transformFn(v);
+      };
+      await store.update([ping, metric.type, storageKey], finalTransformFn);
     }
   }
 
@@ -153,7 +146,7 @@ class Database {
    */
   async getMetric(ping: string, metric: Metric): Promise<string | undefined> {
     const store = this._chooseStore(metric.lifetime);
-    const storageKey = this._buildStorageKey(metric);
+    const storageKey = metric.identifier;
     const value = await store.get([ping, metric.type, storageKey]);
     if (isObject(value)) {
       console.error(`Unexpected value found for metric ${metric}: ${JSON.stringify(value)}. Clearing.`);
@@ -165,12 +158,38 @@ class Database {
   }
 
   /**
-   * Gets all of the persisted metrics related to a given ping.
+   * Helper function to validate and get a specific lifetime data
+   * related to a ping from the underlying storage.
    *
    * # Note
    *
-   * If the value in storage for any of the metrics in a ping is of an unexpected type,
+   * If the value in storage for any of the metrics in the ping is of an unexpected type,
    * the whole ping payload for that lifetime will be thrown away.
+   *
+   * @param ping The ping we want to get the data from
+   * @param lifetime The lifetime of the data we want to retrieve
+   *
+   * @returns The ping payload found for the given parameters or an empty object
+   *          in case no data was found or the data that was found, was invalid.
+   */
+  private async getAndValidatePingData(ping: string, lifetime: Lifetime): Promise<PingPayload> {
+    const store = this._chooseStore(lifetime);
+    const data = await store.get([ping]);
+    if (isUndefined(data)) {
+      return {};
+    }
+
+    if (!isValidPingPayload(data)) {
+      console.error(`Unexpected value found for ping ${ping} in ${lifetime} store: ${JSON.stringify(data)}. Clearing.`);
+      await store.delete([ping]);
+      return {};
+    }
+
+    return data;
+  }
+
+  /**
+   * Gets all of the persisted metrics related to a given ping.
    *
    * @param ping The name of the ping to retrieve.
    * @param clearPingLifetimeData Whether or not to clear the ping lifetime metrics retrieved.
@@ -179,33 +198,15 @@ class Database {
    *          `undefined` in case the ping doesn't contain any recorded metrics.
    */
   async getPing(ping: string, clearPingLifetimeData: boolean): Promise<PingPayload | undefined> {
-    const getAndValidatePingData = async (storeName: string, store: Store): Promise<PingPayload> => {
-      const data = await store.get([ping]);
-      if (isUndefined(data)) {
-        return {};
-      }
-
-      if (!isValidPingPayload(data)) {
-        console.error(`Unexpected value found for ping ${ping} in ${storeName} store: ${JSON.stringify(data)}. Clearing.`);
-        await store.delete([ping]);
-        return {};
-      }
-
-      return data;
-    };
-
-    const userData = await getAndValidatePingData("user", this.userStore);
-    const pingData = await getAndValidatePingData("ping", this.pingStore);
-    const appData = await getAndValidatePingData("app", this.appStore);
+    const userData = await this.getAndValidatePingData(ping, Lifetime.User);
+    const pingData = await this.getAndValidatePingData(ping, Lifetime.Ping);
+    const appData = await this.getAndValidatePingData(ping, Lifetime.Application);
 
     if (clearPingLifetimeData) {
       await this.clear(Lifetime.Ping);
     }
 
-    // Start out with the ping lifetime data,
-    // because it is the one that will probably have more data since it is the default.
     const response: PingPayload = { ...pingData };
-    // Merge the other two lifetimes data with it.
     for (const data of [userData, appData]) {
       for (const metricType in data) {
         response[metricType] = {
@@ -226,17 +227,19 @@ class Database {
    * Clears currently persisted data for a given lifetime.
    *
    * @param lifetime The lifetime to clear.
-   *        If not provided, data for all lifetimes will be cleared.
    */
-  async clear(lifetime?: Lifetime): Promise<void> {
-    if (lifetime) {
-      const store = this._chooseStore(lifetime);
-      await store.delete([]);
-    } else {
-      await this.userStore.delete([]);
-      await this.pingStore.delete([]);
-      await this.appStore.delete([]);
-    }
+  async clear(lifetime: Lifetime): Promise<void> {
+    const store = this._chooseStore(lifetime);
+    await store.delete([]);
+  }
+
+  /**
+   * Clears all persisted metrics data.
+   */
+  async clearAll(): Promise<void> {
+    await this.userStore.delete([]);
+    await this.pingStore.delete([]);
+    await this.appStore.delete([]);
   }
 }
 
--- a/src/database.ts
+++ b/src/database.ts
@@ -50,7 +50,7 @@ export function isValidPingPayload(v: StorageValue): v is PingPayload {
  * {
  *  "pingName": {
  *    "metricType (i.e. boolean)": {
- *      "metricCategory?.metricName": metricPayload
+ *      "metricIdentifier": metricPayload
  *    }
  *  }
  * }
--- a/src/database.ts
+++ b/src/database.ts
@@ -1,0 +1,243 @@
+// /* This Source Code Form is subject to the terms of the Mozilla Public
+//  * License, v. 2.0. If a copy of the MPL was not distributed with this
+//  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { StorageValue, Store } from "storage";
+import PersistentStore from "storage/persistent";
+import Metric, { Lifetime } from "metrics";
+import { isObject, isString, isUndefined } from "utils";
+
+export interface PingPayload {
+  [aMetricType: string]: {
+    [aMetricIdentifier: string]: string
+  }
+}
+
+/**
+ * Verifies if a given value is a valid PingPayload.
+ *
+ * @param v The value to verify
+ */
+export function isValidPingPayload(v: StorageValue): v is PingPayload {
+  if (isObject(v)) {
+    // The root keys should all be metric types.
+    for (const metricType in v) {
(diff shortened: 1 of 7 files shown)


src/database.ts
```diff
@@ -0,0 +1,243 @@
+// /* This Source Code Form is subject to the terms of the Mozilla Public
+//  * License, v. 2.0. If a copy of the MPL was not distributed with this
+//  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { StorageValue, Store } from "storage";
+import PersistentStore from "storage/persistent";
+import Metric, { Lifetime } from "metrics";
+import { isObject, isString, isUndefined } from "utils";
+
+export interface PingPayload {
```
brizental, line 10: The PingPayload will evolve as we add more metrics. I hope that we can use it to validate that it only contain valid metrics and that the metrics payloads are correct. I did not add that to this PR because we have no metrics yet.

src/database.ts
```diff
@@ -0,0 +1,243 @@
...
+   *
+   * @returns The string encoded payload persisted for the given metric,
+   *          `undefined` in case the metric has not been recorded yet.
+   */
+  async getMetric(ping: string, metric: Metric): Promise<string | undefined> {
+    const store = this._chooseStore(metric.lifetime);
+    const storageKey = this._buildStorageKey(metric);
+    const value = await store.get([ping, metric.type, storageKey]);
+    if (isObject(value)) {
+      console.error(`Unexpected value found for metric ${metric}: ${JSON.stringify(value)}. Clearing.`);
```
brizental, line 159: Please let me know what you think about this. This is JSland and anyone can do anything to our data, I decided to simply delete if we find something that is not of the expected format.

src/database.ts
```diff
@@ -0,0 +1,243 @@
...
+    }
+  }
+
+  /**
+   * Gets all of the persisted metrics related to a given ping.
+   *
+   * # Note
+   *
+   * If the value in storage for any of the metrics in a ping is of an unexpected type,
+   * the whole ping payload for that lifetime will be thrown away.
```
brizental, line 173: Now, this I am not so sure. We probably should only delete the metrics that are incorrect, right? Also, what if we end up deleting stuff like `first_run_date` or `client_id`? This is possible although unlikely, again: JSland.

I purposefully did not rabbit hole into this for this PR, but do let me know your thoughts I think this will result in new bugs to figure out what  to do with unexpected data in the storage.

brizental:
FYI: CI didn;t run for this Pr because it was not set up to run for PRs from forks. I changed that and it should run for next commits.

src/database.ts
```diff
@@ -0,0 +1,243 @@
...
+  }
+}
+
+/**
+ * Verifies if a given value is a valid PingPayload.
+ *
+ * @param v The value to verify
+ */
+export function isValidPingPayload(v: StorageValue): v is PingPayload {
+  if (isObject(v)) {
```
Dexterp37, line 22: To reduce indentation, privilege early returns. For example, you could change this by doing:

```js
if (!isObject(v)) {
  return false;
}

for (...)
```

src/database.ts
```diff
@@ -0,0 +1,243 @@
...
+export interface PingPayload {
+  [aMetricType: string]: {
+    [aMetricIdentifier: string]: string
+  }
+}
+
+/**
+ * Verifies if a given value is a valid PingPayload.
+ *
+ * @param v The value to verify
```
Dexterp37, line 19: Please document the return value (even though it's trivial).

src/database.ts
```diff
@@ -0,0 +1,243 @@
...
+ *
+ * {
+ *  "pingName": {
+ *    "metricType (i.e. boolean)": {
+ *      "metricCategory?.metricName": metricPayload
+ *    }
+ *  }
+ * }
+ *
+ * We have one store such in format for each lifetime: user, ping and application.
```
brizental, line 56:
```suggestion
 * We have one store in this format for each lifetime: user, ping and application.
```

src/storage/utils.ts
```diff
@@ -75,8 +75,15 @@ export function updateNestedObject(
...
   const current = target[finalKey];
-  target[finalKey] = transformFn(current);
-  return returnObject;
+  try {
+    const value = transformFn(current);
+    target[finalKey] = value;
+    return returnObject;
+  } catch(e) {
+    console.error("Error while transforming stored value. Ignoring old value.", e.message);
+    target[finalKey] = transformFn(undefined);
```
brizental, line 84:
```suggestion
    target[finalKey] = transformFn();
```

src/database.ts
```diff
@@ -0,0 +1,243 @@
...
+
+/**
+ * The metrics database is an abstraction layer on top of the underlying storage.
+ *
+ * Metric data is saved to the database in the following format:
+ *
+ * {
+ *  "pingName": {
+ *    "metricType (i.e. boolean)": {
+ *      "metricCategory?.metricName": metricPayload
```
Dexterp37, line 51: Why the question mark between the category and the name?

src/database.ts
```diff
@@ -0,0 +1,243 @@
...
+  }
+
+  /**
+   * The key under which to save a metrics value.
+   *
+   * @param metric The metric for which to build the storage key
+   */
+  private _buildStorageKey(metric: Metric): string {
+    if (metric.category) {
+      return `${metric.category}.${metric.name}`;
```
Dexterp37, line 77: You would probably want a method on the `Metric` called `id`, because `category.name` is the id. Each metric will know how to build the id, and the database knows how to map ids to keys.

src/database.ts
```diff
@@ -0,0 +1,243 @@
...
+
+  /**
+   * Records a given value to a given metric.
+   * Will overwrite in case there is already a value in there.
+   *
+   * @param metric The metric to record to.
+   * @param value The value we want to record to the given metric.
+   */
+  async record(metric: Metric, value: string): Promise<void> {
+    if (!metric.disabled) {
```
Dexterp37, line 113: To reduce indentation, privilege early returns:

```js
if (metric.disabled) {
  return;
}
```

src/database.ts
```diff
@@ -0,0 +1,243 @@
...
+
+  /**
+   * Records a given value to a given metric,
+   * by applying a transformation function on the value currently persisted.
+   *
+   * @param metric The metric to record to.
+   * @param transformFn The transformation function to apply to the currently persisted value.
+   */
+  async transform(metric: Metric, transformFn: (v?: string) => string): Promise<void> {
+    if (!metric.disabled) {
```
Dexterp37, line 130: I'd privilege the early return here too, to improve the readability of the code below.

src/database.ts
```diff
@@ -0,0 +1,243 @@
...
+      const store = this._chooseStore(metric.lifetime);
+      const storageKey = this._buildStorageKey(metric);
+      for (const ping of metric.sendInPings) {
+        const finalTransformFn = (v: StorageValue): string => {
+          if (isObject(v)) {
+            throw new Error(`Unexpected value found for metric ${metric}: ${JSON.stringify(v)}.`);
```
Dexterp37, line 136: Does this mean our specific metrics APIs will potentially throw? If so, this could mean breaking the user's functionalities. The Glean SDK intentionally avoids this, I believe we should do the same.

src/database.ts
Dexterp37: I think this makes sense. But why is an object an invalid value here?

src/database.ts
Dexterp37: Yeah, I believe we should only delete the specific value that's not expected and leave the rest alone.

src/database.ts
```diff
@@ -0,0 +1,243 @@
...
+ *
+ */
+class Database {
+  private userStore: Store;
+  private pingStore: Store;
+  private appStore: Store;
+
+  constructor() {
+    this.userStore = new PersistentStore("user");
+    this.pingStore = new PersistentStore("ping");
```
Dexterp37, line 66: Since we'll likely be storing pings in a  `PersistentStorage` as well, does it make sense to have more explicit names for keys, e.g. "pingLifetimeMetrics"?

src/database.ts
```diff
@@ -0,0 +1,243 @@
...
+    const userData = await getAndValidatePingData("user", this.userStore);
+    const pingData = await getAndValidatePingData("ping", this.pingStore);
+    const appData = await getAndValidatePingData("app", this.appStore);
+
+    if (clearPingLifetimeData) {
+      await this.clear(Lifetime.Ping);
+    }
+
+    // Start out with the ping lifetime data,
+    // because it is the one that will probably have more data since it is the default.
```
Dexterp37, line 206: And what difference does this make? :)

src/database.ts
```diff
@@ -0,0 +1,243 @@
...
+   * # Note
+   *
+   * If the value in storage for any of the metrics in a ping is of an unexpected type,
+   * the whole ping payload for that lifetime will be thrown away.
+   *
+   * @param ping The name of the ping to retrieve.
+   * @param clearPingLifetimeData Whether or not to clear the ping lifetime metrics retrieved.
+   *
+   * @returns An object containing all the metrics recorded to the given ping,
+   *          `undefined` in case the ping doesn't contain any recorded metrics.
```
Dexterp37, line 179: The function returns either `{}` or `undefined`. I think we should settle on one :)

src/database.ts
```diff
@@ -0,0 +1,243 @@
...
+      for (const metricType in data) {
+        response[metricType] = {
+          ...response[metricType],
+          ...data[metricType]
+        };
+      }
+    }
+
+    if (Object.keys(response).length === 0) {
+      return;
```
Dexterp37, line 219: This will return `undefined`, but we're also returning `{}` above. How do you want this API to behave?

src/database.ts
```diff
@@ -0,0 +1,243 @@
...
+    } else {
+      return response;
+    }
+  }
+
+  /**
+   * Clears currently persisted data for a given lifetime.
+   *
+   * @param lifetime The lifetime to clear.
+   *        If not provided, data for all lifetimes will be cleared.
```
Dexterp37, line 229: When do we delete data for all the stores?

I'd rather have a separate function just to do that, to make sure we don't mistakenly clear the all data when we don't want to.

src/metrics/index.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  constructor(type: string, meta: CommonMetricData) {
+    this.type = type;
+
+    this.name = meta.name;
+    this.sendInPings = meta.sendInPings;
+    this.lifetime = meta.lifetime;
+    this.disabled = meta.disabled;
+
+    if (meta.category) this.category = meta.category;
```
Dexterp37, line 51: Can we retain the same style that we used in the rest of the code?

```suggestion
    if (meta.category) {
      this.category = meta.category;
    }
```

src/metrics/index.ts
```diff
@@ -0,0 +1,55 @@
...
+
+class Metric implements CommonMetricData {
+  readonly type: string;
+  readonly name: string;
+  readonly category?: string;
+  readonly sendInPings: string[];
+  readonly lifetime: Lifetime;
+  readonly disabled: boolean;
+
+  constructor(type: string, meta: CommonMetricData) {
```
Dexterp37, line 43: Can we add a computed property called `id`?

tests/database.spec.ts
```diff
@@ -0,0 +1,474 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import assert from "assert";
+
+import Database, { isValidPingPayload } from "database";
+import Metric, { Lifetime } from "metrics";
+
+describe("Database", function() {
```
Dexterp37, line 10: Did you make sure to port back all the applicable tests from the Glean SDK related module?

Dexterp37 (changes requested):

src/database.ts
brizental: The question mark means that we may not have a category. This will go away when I move the identifier building logic to the metric code anyways.

src/database.ts
brizental: Nope, I changed the update function to expect errors from this transformation function. It will catch this error and ignore the incorrect value if it happens.

src/database.ts
brizental: I am assuming for now that metric payloads are be strings (or undefined if it hasn't been recorded yet). When we start having actual metrics we will probably have a union type `MetricPayload` with the possible metric payloads and here we will change this conditional to `isMetricPayload`.

src/database.ts
brizental: I agree, would mind if I made this into a folow up though? I think there is a rabbit hole potential here, I am not sure what we should do for example if the problematic value is a user lifetime metric, or a key metric such as client_id. This is not as simple as just delete whatever is wrong.

src/database.ts
brizental: Argh, this was a mistake, it this it is best to return `undefined` when there are no metrics... Then callers can just do `if (ping) { ... }`. `{}` is a truthy value.

src/database.ts
brizental: Less iterations, because we don't iterate on the store that probably has the most entries. It's not a big deal, I can remvoe the comment so as not to make it a big deal :D

src/database.ts
brizental: Yes, that was mistake as I said above... Nice catch though.

src/database.ts
brizental: > When do we delete data for all the stores?

When upload is disabled.

I agree about creating adifferent function for that though.

tests/database.spec.ts
brizental: Good point, I didn't check.

src/database.ts
Dexterp37: Cool, ok then!

src/database.ts
Dexterp37: Ok, that's fine: let's file a follow-up .

src/database.ts
brizental: Hm, actually taking a closer look I realized there is no mistake, the function doesn't ever return an empty object.

The inner function I defined `getAndValidatePingData` does, but the actual `getPing` function only checks and if the ping object is empty it returns undefined.

Dexterp37: The two returns [here](https://github.com/mozilla/glean.js/pull/9/files/7b428a6d8d680042f9ef8ea78e2e24d7ea80301d#diff-724f334a2b47a5e7cbcab252ce04fc232071d0bc058c96ed6fd4ac96e5dff308R185-R191) explicitly return `{}`?

brizental: Yup, but they are inside the inner function `getAndValidatePingData`. `getPing` returns [here](https://github.com/mozilla/glean.js/pull/9/files/7b428a6d8d680042f9ef8ea78e2e24d7ea80301d#diff-724f334a2b47a5e7cbcab252ce04fc232071d0bc058c96ed6fd4ac96e5dff308R218) and it checks if the object is empty before doing so.

Dexterp37: SHISH. Can we please take that helper function out ? :)

brizental: Heh, I already did. I was just waiting for that request :D

src/database.ts
brizental: If I do this for this specific case, Typescript complains that `v` may not be an object...

tests/database.spec.ts
brizental: I checked and this suite cover all the cases covered by the suite in glean-core. We of course don't have coverage for the `remove_metric` function because I didn't add it to the Database for now (it is only necessary for experiments).

tests/database.spec.ts
```diff
@@ -0,0 +1,474 @@
...
+      assert.strictEqual(false, isValidPingPayload({
+        "aPing": {
+          "string": {
+            "too.nested": "not quite"
+          }
+        }
+      }));
+      assert.strictEqual(false, isValidPingPayload({ "string": "almost!" }));
+      // Valids
+      assert.strictEqual(true, isValidPingPayload({ "string": {} }));
```
brizental, line 256: Hey Alessio, would you mind confirming that this should actually be an invalid payload? I will file a follow up to fix the `delete` function in case that is the case.

src/database.ts
Dexterp37: Ok leave it be

tests/database.spec.ts
Dexterp37: What's valid or not is dictated by the Glean schema. You can check how validation is performed in the SDK and apply schema validation here, too: we don't need to reinvent the wheel :)

brizental: I have looked at the schema, but it is still unclear to me wether a metric type containing an empty object is considered valid. I _think_ so, but I just wanted to confirm :)

src/database.ts
brizental: https://bugzilla.mozilla.org/show_bug.cgi?id=1682273

tests/database.spec.ts
Dexterp37: I think so too, but my point is: we shouldn't "know" :-) Let's have a test running the schema in a validator :) Can you file a follow-up?

src/database.ts
```diff
@@ -0,0 +1,247 @@
...
+
+  /**
+   * Records a given value to a given metric.
+   * Will overwrite in case there is already a value in there.
+   *
+   * @param metric The metric to record to.
+   * @param value The value we want to record to the given metric.
+   */
+  async record(metric: Metric, value: string): Promise<void> {
+    debugger;
```
Dexterp37, line 102: Stray `debugger` statement :)

src/database.ts
```diff
@@ -0,0 +1,247 @@
...
+   *          `undefined` in case the metric has not been recorded yet.
+   */
+  async getMetric(ping: string, metric: Metric): Promise<string | undefined> {
+    const store = this._chooseStore(metric.lifetime);
+    const storageKey = metric.identifier;
+    const value = await store.get([ping, metric.type, storageKey]);
+    if (isObject(value)) {
+      console.error(`Unexpected value found for metric ${metric}: ${JSON.stringify(value)}. Clearing.`);
+      await store.delete([ping, metric.type, storageKey]);
+      return;
```
Dexterp37, line 155: I'm still a bit puzzled by this, but we can revisit later: promises can be resolved or rejected, so maybe we can just assume that if it is resolved it has a value, otherwise rejected? Nothing to do now.

tests/database.spec.ts
brizental: Ah! I see! That is a task for after we implement the ping maker, but I'll still file that bug :)

Dexterp37 (approved):
r+wc

src/database.ts
brizental: AAAAAAAAAAAAA 🤦‍♀️

src/database.ts
brizental: Yeah, I considered just leaving the data there. But that would just mean this would fail until we record/transform again, because we leave the corrupted data there...

src/database.ts
brizental: Added a lint rule to check for these...

tests/database.spec.ts
brizental: https://bugzilla.mozilla.org/show_bug.cgi?id=1682282

src/database.ts
brizental: We can keep up this discussion on the bug I opened to investigate the best behaviour for this :)

PR #10, opened 2020-12-11
Bug 1681516 - Fix percentage calc for size check CI

brizental:
I also try to make the percentage prettier by limiting it to two decimal cases.


Dexterp37 (approved):
Thanks for disabling it.

PR #11, opened 2020-12-14
Bug 1679379 - Implement the boolean metric type 

brizental:


src/glean.ts
```diff
@@ -0,0 +1,20 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Database from "database";
+
+class Glean {
```
Dexterp37, line 7: Do you have a bug open for implementing the general API?

src/database.ts
```diff
@@ -139,17 +132,22 @@ class Database {
...
+    validateFn: (v: unknown) => v is T,
+    metric: Metric
+  ): Promise<T | undefined> {
     const store = this._chooseStore(metric.lifetime);
     const storageKey = metric.identifier;
     const value = await store.get([ping, metric.type, storageKey]);
-    if (isObject(value)) {
-      console.error(`Unexpected value found for metric ${metric}: ${JSON.stringify(value)}. Clearing.`);
+    if (!isUndefined(value) && !validateFn(value)) {
+      console.error(`Unexpected value found for metric ${metric.identifier}: ${JSON.stringify(value)}. Clearing.`);
```
Dexterp37, line 150: Thinking again about the deletion, I'm not sure we should do that. We should rather investigate what we do with the SDK, because we have a weird edge case here: what if user adds a new metric with the same id but a different type? Leaving the deletion in place, trying to record the new type will make the old data go away, even though it's not "wrong", is the new type that's wrong.

Can you file a bug for investigating this aspect (if you can't do it right now) and add  a comment here?

src/metrics/payload.ts
```diff
@@ -0,0 +1,19 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { BooleanMetricPayload, isBooleanMetricPayload } from "metrics/boolean";
+import { isString } from "utils";
+
+export function isMetricPayload(type: string, v: unknown): v is MetricPayload {
```
Dexterp37, line 8: Nit: what about `isValidDataForMetricType`?

Dexterp37, line 8: Please add a comment explaining what's the purpose of this function and why it's used.

src/metrics/payload.ts
```diff
@@ -0,0 +1,19 @@
...
+  switch (type) {
+  case "boolean":
+    return isBooleanMetricPayload(v);
+  default:
+    return isString(v);
+  }
+}
+
+// Leaving the `string` as a valid metric payload here so that tests keep working for now.
```
Dexterp37, line 17: Mh, would it be hard to fix tests right now in this PR? If so, can you please file a bug and add a reference to it here?

src/storage/index.ts
```diff
@@ -42,7 +42,7 @@ export interface StorageObject {
  *          stating wether `v` is a valid StorageValue.
  */
 export function isStorageValue(v: unknown): v is StorageValue {
-  if (isUndefined(v) || isString(v)) {
+  if (isUndefined(v) || isString(v) || isBoolean(v)) {
```
Dexterp37, line 45: Mh, why can't this use `isMetricPayload`?

Right now, adding a new metric type is a bit convoluted. We need to add checks in the storage, which is weird, specifically for the metric type.

src/utils.ts
```diff
@@ -3,13 +3,17 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 export function isObject(v: unknown): v is Record<string | number | symbol, unknown> {
-  return (v && typeof v === "object" && v.constructor === Object);
+  return !!(v && typeof v === "object" && v.constructor === Object);
```
Dexterp37, line 6:
```suggestion
  return (!!v && typeof v === "object" && v.constructor === Object);
```

Is this what you're really looking for? Forcing truthiness outside makes it a bit harder to understand it on the first look otherwise.

Dexterp37 (changes requested):
I took a first high level pass :) Will take a closer look after this first round.

src/glean.ts
brizental: Yes:
- https://bugzilla.mozilla.org/show_bug.cgi?id=1677444
- https://bugzilla.mozilla.org/show_bug.cgi?id=1677443

src/database.ts
brizental: If the user has a metric with the same id but with a different type it will be recorded in a different index on the database, which means they won't interfere with each other. The index for a metric is `[ "thePing", "theType", "theIdentifier" ]`.

src/metrics/payload.ts
brizental: Uhum. Unrelated but worth mentioning, I realized I forgot to add the jsdoc rules to our linter config, this is why this bugs keep happening. I created a bug to fix this: https://bugzilla.mozilla.org/show_bug.cgi?id=1682482

src/metrics/payload.ts
brizental: Tests will be fixed as soon as I implement a String metric type which I was planning to be the one I take after Boolean. If I want to keep tests working without this workaround, I'd have to change all example metrics in the Database tests to be Boolean metrics, which I don't think is worth the effort. I'd rather change all metrics to Strings on those tests once we implement that metric type.

src/storage/index.ts
brizental: > Mh, why can't this use isMetricPayload?

The storage shouldn't know about the metrics. It will not only store metrics, but also ping payloads and maybe other stuff.

> Right now, adding a new metric type is a bit convoluted. We need to add checks in the storage, which is weird, specifically for the metric type.

The storage has the capability to store any valid JSON type, so string, arrays, booleans, numbers and other JSON objects. One should not have to add anything to the storage module when they add a new metric type, I only had to do it now because it didn't include boolean yet. I will also have to do it when I add the Counter metric type, which will present the need to store numbers. After that we will have covered all the JSON possible types and shouldn't need to change this module when adding metrics in the future.

src/utils.ts
brizental: No, so I had to add this `!!` because typescript was giving me an error `Type 'unknown' is not assignable to type 'boolean'.`. I searched this, but this seems to be the accepted workaround to coerce to a boolean, TS does not do that on its own. Another option would be to do this `<boolean>(v && typeof v === "object" && v.constructor === Object);` I have no strong opinions, if you think the second option is clearer I can change it.

brizental: What I needed to do with this change, was coerce the whole expression to a boolean, not just `v`.

src/metrics/payload.ts
brizental: Hm, all the other type validation function are named like this though. I'd rather leave it as is for standardization sake.

src/utils.ts
brizental: Okay, I was attending to your review comments and I added an extra test to the storage suite that I realized was missing. Which was for the best because I found a bug in this utils functions (I'll file a bug to add tests specifically to this module, I should have done that from the start). 

Anyways, this works: `(typeof v === "boolean" || (typeof v === "object" && v !== null && v.constructor === Boolean))`. I need to check if `v` is `null` **after** I check that it is of type object. This happens because `null` is of type object and TS apparently only takes into account the last condition that you set when asserting the possible type of something.

src/database.ts
Dexterp37: Is this the same behaviour we have in the SDK?

src/metrics/boolean.ts
```diff
@@ -0,0 +1,58 @@
...
+    super("boolean", meta);
+  }
+
+  /**
+   * Sets to the specified boolean value.
+   *
+   * @param glean the Glean instance this metric belongs to.
+   * @param value the value to set.
+   */
+  async set(glean: Glean, value: BooleanMetricPayload): Promise<void> {
```
Dexterp37, line 33: Is there any way for us to not have to pass in the Glean instance?

src/metrics/boolean.ts
```diff
@@ -0,0 +1,58 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Metric, { CommonMetricData } from "metrics";
+import Glean from "glean";
+import { isBoolean } from "utils";
+
+export type BooleanMetricPayload = boolean;
```
Dexterp37, line 9: Why do we need this alias?

src/metrics/boolean.ts
```diff
@@ -0,0 +1,58 @@
...
+   *
+   * This doesn't clear the stored value.
+   *
+   * @param glean the Glean instance this metric belongs to.
+   * @param ping the ping from which we want to retrieve this metrics value from.
+   *
+   * @returns The value found in storage or `undefined` if nothing was found.
+   */
+  async testGetValue(glean: Glean, ping: string): Promise<BooleanMetricPayload | undefined> {
+    return glean.db.getMetric(ping, isBooleanMetricPayload, this);
```
Dexterp37, line 54: We need some way to make this this is only used in test mode. Can you add a flag or something to the Glean object, check it here and throw if we're not in test mode?

src/storage/index.ts
Dexterp37: Got it, thanks for explaining.

tests/metrics/boolean.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import assert from "assert";
+
+import Glean from "glean";
+import BooleanMetric from "metrics/boolean";
+import { Lifetime } from "metrics";
+
+describe("BooleanMetric", function() {
```
Dexterp37, line 11: Did you check that we have all the tests for the boolean type that we have in the SDK? Bonus points for having the same tests with the same names..

Dexterp37 (changes requested):

src/database.ts
brizental: No, but that is on purpose. On the SDK we actually panic in this situation: https://github.com/mozilla/glean/blob/main/glean-core/src/database/mod.rs#L501. But the error message for that panic is the core difference between both situations, on glean-core it is deemed impossible for this error to happen, while on Glean.js it is completely plausible that it will happen given that anyone can go to the debug console of a web extension and mess with the data.

src/metrics/boolean.ts
brizental: I am thinking that each metric should define their payload type and the payload validation function the same way. For boolean it seems superfluous, but it is just for future standardization sake.

src/database.ts
Dexterp37: Ok please add this as a comment in the code.

tests/metrics/boolean.spec.ts
brizental: I did check this time :) There are more tests here actually than on the SDK, but all the tests that we have there we have here. The names are not tha same though.
- `boolean_serializer_should_correctly_serialize_boolean` is the name on the SDK for the test I called `ping payload is correct`. I changed because I thought the name didn't really make sense given Glean.js structure. 
- `set_properly_sets_the_value_in_all_stores` I changed to `set properly sets the value in all ping` because we are not using the "store" terminology in Glean.js for the same purposes we are on the SDK.

src/metrics/boolean.ts
Dexterp37: There's pros and cons about this architectural choice: this is leaking the payload representation and responsibilities to the type implementation, while this should probably be something the storage has authority on. We're unlikely to move from JSON to some other format in the immediate future, but this choice will definitely make it a bit harder to move somewhere else (e.g. binary?) if needed. So, nothing to do here, other than thinking of future implications.

src/metrics/boolean.ts
brizental: I'll add a flag.

src/metrics/boolean.ts
brizental: Hm, we'd have to investigate that further. What would be the advantage of that?

src/metrics/boolean.ts
brizental: I am not sure I fully understand the problem, would you mind expanding?

Also, how is it so different from what we do in glean-core https://github.com/mozilla/glean/blob/main/glean-core/src/metrics/mod.rs#L78-L120?

src/metrics/boolean.ts
brizental: The Glean constructor cannot have any arguments, because we will export an instance of Glean, the user won't instantiate it. This means that this flag would need to be on the `initialize` function, which I have made a point of not implementing in this PR since it is sorta unrelated and also we already have [a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1677443) for that.

My point being, would you mind if I file a follow up for this that depends on having implemented the initialize function?

Dexterp37: Sure, just add a todo in here and reference that bug number :)

src/metrics/boolean.ts
Dexterp37: Good point, i was wrong.

src/metrics/boolean.ts
Dexterp37: This is an API ergonomics problem: it sucks to provide the Glean object everywhere, especially knowing that's a "singleton". This is the whole point why, in the SDK, the Glean object is a singleton: to have a nice ergonomic API :)

I'd take the opportunity to investigate that, before we scale up the number of metrics.

Dexterp37 (changes requested):
See my other comments. My major concern is about the requirement to pass in the Glean object to the specific metrics API.

src/metrics/boolean.ts
brizental: I see. My mind complicated this quite a bit first time I read this comment. Now that I think about it, it is 100% doable. I'll send a new commit shortly. Thanks for pressing this :)

src/glean.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    return Glean._instance;
+  }
+
+  static get db(): Database {
+    return Glean.instance._db;
+  }
+
+  static get uploadEnabled(): boolean {
```
Dexterp37, line 41: Is there any way to mark this "internal" and not part of the "public API"? (I know this is JS at the end of the day, but I'd love to make this explicit).

Dexterp37 (approved):
r+wc

Dexterp37:
@brizental can you also file a bug for adding documentation? Did we settle on adding docs to the Glean Book?

src/glean.ts
brizental: Not right now, otherwise we won't be able to access it from the tests. I'll add a TODO comment to make it `private` once we have the `resetGlean` function (which we do already have a bug for https://bugzilla.mozilla.org/show_bug.cgi?id=1682769).

Note that I can make this methods `private` to the Glean class, but once a user is consuming it from Javascript all bets are off.

brizental:
> @brizental can you also file a bug for adding documentation? Did we settle on adding docs to the Glean Book?

Done https://bugzilla.mozilla.org/show_bug.cgi?id=1682785.

Yes, on the MVP proposal we settled on the Glean book and I still stand by that decision, even though I am not clear on _how_ we add Glean.js docs there.

PR #13, opened 2020-12-16
Bug 1679378 - Implement the string metric type

brizental:


src/metrics/string.ts
```diff
@@ -0,0 +1,81 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Metric, { CommonMetricData } from "metrics";
+import Glean from "glean";
+import { isString } from "utils";
+
+export const MAX_LENGTH_VALUE = 100;
+
+export type StringMetricPayload = string;
+/**
```
Dexterp37, line 12:
```suggestion

/**
```

Uber-nit.

src/metrics/string.ts
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * @returns A special Typescript value (which compiles down to a boolean)
+ *          stating wether `v` is a valid string metric payload.
+ */
+export function isStringMetricPayload(v: unknown): v is StringMetricPayload {
+  if (!isString(v)) {
+    return false;
+  }
+
+  if (v.length > MAX_LENGTH_VALUE) {
```
Dexterp37, line 30: Mh, i'm not sure we should do the check here. The fact that the string is "long" doesn't make it not a string :)

This is mostly for preventing people to hack the storage and put long strings in there? If so, I think we should not do that. There's many other ways to put big stuff in the ping, though, and we have no way to really prevent that from happening on the client. We should assume "good faith" I think. @brizental WDYT?

src/metrics/string.ts
```diff
@@ -0,0 +1,81 @@
...
+   * Gets the currently stored value as a string.
+   *
+   * This doesn't clear the stored value.
+   *
+   * @param ping the ping from which we want to retrieve this metrics value from.
+   *
+   * @returns The value found in storage or `undefined` if nothing was found.
+   */
+  async testGetValue(ping: string): Promise<StringMetricPayload | undefined> {
+    return Glean.db.getMetric(ping, isStringMetricPayload, this);
```
Dexterp37, line 77: Please add a todo for the test mode here, too, with the bug number

tests/metrics/boolean.spec.ts
```diff
@@ -9,6 +9,15 @@ import BooleanMetric from "metrics/boolean";
...
 
 describe("BooleanMetric", function() {
+  beforeEach(async function() {
+    // TODO: Replace this code with a resetGlean function, after Bug 1682769 is resolved.
+
+    // Reset upload enabled state, not to inerfere with other tests.
+    Glean.uploadEnabled = true;
+    // Clear the database.
+    await Glean.db.clearAll();
+  });
```
Dexterp37, line 19: You're now copying over the code in two places already: what about making a temporary testResetGlean function in this PR?

tests/metrics/string.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+
+import assert from "assert";
+
+import Glean from "glean";
+import StringMetric, { MAX_LENGTH_VALUE } from "metrics/string";
+import { Lifetime } from "metrics";
+ 
+describe("StringMetric", function() {
+  beforeEach(async function() {
+    // TODO: Replace this code with a resetGlean function, after Bug 1682769 is resolved.
```
Dexterp37, line 13: See my comment about the temporary test reset function.

Dexterp37 (changes requested):
This looks pretty good already. there's just a few minor changes requested.

src/metrics/string.ts
brizental: > This is mostly for preventing people to hack the storage and put long strings in there?

This is a side effect, FWIW this check is not very effective against people hacking the storage anyways. Mostly it has to do with the fact that a valid string payload is not longer than 100 chars, and this function is checking if `v` is a valid string payload. We could go further and define a special Typescript type that is a "string not longer than 100 characters", I decided against that because I  thought it introduced unnecessary complexity.

> There's many other ways to put big stuff in the ping, though, and we have no way to really prevent that from happening on the client.

This check does prevents us from retrieving a string that is definitely wrong from the storage. It does not prevent us from getting values that were messed with, but are still valid strings payloads. Preventing that reliably would require more investigation and I am not even sure it's possible, honestly.

> @brizental WDYT?

Assuming that this function should not only check if something is a string but also if it is a valid string payload, I'd argue for leaving the check in. Long strings are not valid string payloads.

Dexterp37: >
> This check does prevents us from retrieving a string that is definitely wrong from the storage. It does not prevent us from getting values that were messed with, but are still valid strings payloads. Preventing that reliably would require more investigation and I am not even sure it's possible, honestly.

Please this as a comment above the size check.

brizental: I will add this as a comment note on the `getMetric` function, otherwise I'd have to copy paste that everywhere and also I think it makes more sense there. Commit incoming with that change :)

Dexterp37 (approved):

PR #14, opened 2020-12-16
Bug 1679380 - Implement the counter metric type

brizental:
The second commit (008bce4) is a bonus. I am not sure that is a good pattern, but it does centralize payload validation instead of doing it in a bunch of different places and in different ways, and I think that at least is good. Please let me know what you think :)

### DIFF
--- a/src/database.ts
+++ b/src/database.ts
@@ -110,7 +110,7 @@ class Database {
    * @param metric The metric to record to.
    * @param transformFn The transformation function to apply to the currently persisted value.
    */
-  async transform(metric: Metric, transformFn: (v?: MetricPayload) => MetricPayload): Promise<void> {
+  async transform<T extends MetricPayload>(metric: Metric, transformFn: (v?: T) => T): Promise<void> {
     if (metric.disabled) {
       return;
     }
@@ -119,7 +119,7 @@ class Database {
     const storageKey = metric.identifier;
     for (const ping of metric.sendInPings) {
       const finalTransformFn = (v: StorageValue): Exclude<StorageValue, undefined> => {
-        if (!isUndefined(v) && !isMetricPayload(metric.type, v)) {
+        if (!isUndefined(v) && !isMetricPayload<T>(metric.type, v)) {
           throw new Error(`Unexpected value found for metric ${metric.identifier}: ${JSON.stringify(v)}.`);
         }
         return transformFn(v);
@@ -146,7 +146,6 @@ class Database {
    * Currently Glean.js does not include mechanisms to reliably prevent that.
    *
    * @param ping The ping from which we want to retrieve the given metric.
-   * @param validateFn A validation function to verify if persisted payload is in the correct format.
    * @param metric An object containing the information about the metric to retrieve.
    *
    * @returns The payload persisted for the given metric,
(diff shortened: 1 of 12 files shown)


src/database.ts
```diff
@@ -110,7 +110,7 @@ class Database {
    * @param metric The metric to record to.
    * @param transformFn The transformation function to apply to the currently persisted value.
    */
-  async transform(metric: Metric, transformFn: (v?: MetricPayload) => MetricPayload): Promise<void> {
+  async transform<T extends MetricPayload>(metric: Metric, transformFn: (v?: T) => T): Promise<void> {
```
Dexterp37, line 113: Good call.

src/metrics/counter.ts
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  if (v <= 0) {
+    return false;
+  }
+
+  return true;
+}
+
+class CounterMetric extends Metric {
```
Dexterp37, line 39: I now noticed that we're missing the class documentation. Can you kindly either file a bug or add it for this and the other metrics as well? Just copy over and adapt the one from the SDK

src/metrics/counter.ts
```diff
@@ -0,0 +1,102 @@
...
+  constructor(meta: CommonMetricData) {
+    super("counter", meta);
```
Dexterp37, line 41: I suspect we can do better than storing a string for each metric. Can you please file a bug about this?

src/metrics/counter.ts
```diff
@@ -0,0 +1,102 @@
...
+    if (amount <= 0) {
+      // TODO: record error once Bug 1682574 is resolved.
+      console.warn(`Attempted to add an invalid amount ${amount}. `);
+      return;
+    }
+
+    const transformFn = ((amount) => {
+      return (v?: CounterMetricPayload): CounterMetricPayload =>  {
+        let result = v ? v + amount : amount;
+        if (result > Number.MAX_SAFE_INTEGER) {
```
Dexterp37, line 74: Does this change the maximum representable number compared to the SDK?

Dexterp37 (changes requested):

src/metrics/counter.ts
brizental: SHISH, you are right. Nah, I can do it in this PR, no need for yet another bug.

src/metrics/counter.ts
brizental: We can have an enum (which will just compile down to a list of strings), like for Lifetime (see: https://github.com/mozilla/glean.js/blob/main/src/metrics/index.ts#L10-L17). Do you think that is better?

Dexterp37: Let's discuss about this in a separate bug. Storing one string per metric instance could be problematic, when it comes to memory consumption of the library. We should measure this along with the overall performances of Glean.js.

src/metrics/counter.ts
brizental: Good question, let me check... Actually is much larger then the maximum representable number on the SDK:

- Javascript: 9007199254740991
- Rust: 4294967295

src/metrics/counter.ts
brizental: FWIW now that I think about we can also just remove the type have `instaceof` checks for checking the metric type. I'll file an investigation bug.

brizental: https://bugzilla.mozilla.org/show_bug.cgi?id=1682864

Dexterp37 (approved):

PR #16, opened 2020-12-17
Bug 1679381 - Implement the datetime metric type

brizental:
Get ready for some fun ~~string~~ date and timezone manipulation!


src/metrics/time_unit.ts
```diff
@@ -0,0 +1,281 @@
...
+ *
+ * @param unit The time unit for which to get the pattern.
+ *
+ * @returns The pattern as a string.
+ *
+ * @throws In case an unknown time unit is passed.
+ */
+function getTimeUnitDateStringPattern(unit: TimeUnit): string {
+  switch (unit) {
+  case TimeUnit.Nanosecond:
```
brizental, line 39: Javascript Date object only has milliseconds resolution. Which means nano and microseconds datetime metrics don't do anything different than record a payload with some extra zeroes at the end.

We could use the browser specific `window.performance` API to get up to microsecond precision, but I personally I am not sure if it is worth the effort and if it is I'd argue for making it a follow up.

My recommendation is that we leave it like this and have a glinter warning for people adding nanoseconds or microseconds datetimes and building for JS. If that is not possible a documentation + docstring note could suffice.

src/metrics/time_unit.ts
```diff
@@ -0,0 +1,281 @@
...
+ * It is expected that the string has already been validated by `validateDateString`
+ * before it is given to this function.
+ *
+ * # Note
+ *
+ * The Date object is timezone unaware,
+ * thus the object created here will be relative to local time.
+ *
+ * If the string is in a different timezone,
+ * the timezone offset will be applied before transforming to an object.
```
brizental, line 225: This is not out of my head. It is the same behaviour as when you build a Date object from an ISO string with a different timezone than your local one.

tests/metrics/time_unit.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  afterEach(function () {
+    // Undo the monkeypatch.
+    //
+    // Need to @ts-ignore here because the `getTimezoneOffset` type
+    // doesn't mention anything about the `restore` method,
+    //
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    Date.prototype.getTimezoneOffset.restore();
```
brizental, line 25: Hey, Alessio, you are more familiar with sinon than I am at this point. Should I be using sinon's sandbox here? Why is that better, if so?

brizental: Nevermind, answered my own question and moved to user a sandbox.

src/metrics/time_unit.ts
Dexterp37: > 
> 
> Javascript Date object only has milliseconds resolution. Which means nano and microseconds datetime metrics don't do anything different than record a payload with some extra zeroes at the end.

Yeah, I feared as much :)

> We could use the browser specific `window.performance` API to get up to microsecond precision, but I personally I am not sure if it is worth the effort and if it is I'd argue for making it a follow up.
 
 I agree with you we probably don't need it for `DateTime`, but that's presumably what we'll want for `Timespan` and `TimingDistribution` or `Event`.
 
> My recommendation is that we leave it like this and have a glinter warning for people adding nanoseconds or microseconds datetimes and building for JS. If that is not possible a documentation + docstring note could suffice.

yes, I agree. Let's do this for the `DateTime` metric.

src/metrics/time_unit.ts
```diff
@@ -0,0 +1,281 @@
...
+ *
+ * @returns The pattern as a string.
+ *
+ * @throws In case an unknown time unit is passed.
+ */
+function getTimeUnitDateStringPattern(unit: TimeUnit): string {
+  switch (unit) {
+  case TimeUnit.Nanosecond:
+    // Reference string: 2020-12-17T10:56:43.528000000
+    return "(\\d{4}-([0]\\d|1[0-2])-([0-2]\\d|3[0-1])(T(([0-1]\\d)|([0-2][0-3])))(:[0-5]\\d)(:[0-5]\\d)(\\.\\d{9}))";
```
Dexterp37, line 41: Instead of the hard to debug string formatting madness here, what about using `theDate.setXX` and then `theDate.toISOString` or `toUTC` as needed?

src/metrics/time_unit.ts
```diff
@@ -0,0 +1,281 @@
...
+ *
+ * This function does not validate the date string.
+ *
+ * @param value The string for which to guess the time unit.
+ *
+ * @returns The tie unit variant for the precision of the given date.
+ *
+ * @throws In case the time unit cannot be guesses because this string has an unespected length.
+ */
+function guessTimeUnitPrecision(value: string): TimeUnit {
```
Dexterp37, line 78: Wow. Why do we need this? Don't we always have the time unit?

src/metrics/time_unit.ts
```diff
@@ -0,0 +1,281 @@
...
+ * for the `time_unit` property on Glean's metrics.yaml file:
+ * https://mozilla.github.io/glean/book/user/metrics/datetime.html#configuration
+ *
+ * @param unit A string representing a specific time unit.
+ *
+ * @returns The TimeUnit variant related to the given string.
+ *
+ * @throws In case the given string is not in the list of accepted strings.
+ */
+export function timeUnitFromString(unit: string): TimeUnit {
```
Dexterp37, line 128: Instead of rolling out our own, can't we use [this](https://www.typescriptlang.org/docs/handbook/enums.html#reverse-mappings)? Do we **need** the string value to start by lowercase letters (since this makes it not possible to use the reverse-mappings above)

src/metrics/datetime.ts
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * Checks whether or not `v` is a valid datetime metric payload.
+ *
+ * @param v The value to verify.
+ *
+ * @returns A special Typescript value (which compiles down to a boolean)
+ *          stating whether `v` is a valid boolean metric payload.
+ */
+export function isDatetimeMetricPayload(v: unknown): v is DatetimeMetricPayload {
```
Dexterp37, line 20: I think we should store the time unit (and any metadata we need for interpreting the data) in the database, as we do in the [SDK](https://searchfox.org/glean/rev/c0f26c283b1df6f354662ca8388a80e2075e7fbd/glean-core/src/metrics/mod.rs#101,173) and then have one specific function (as done in the SDK) to only get the data as Glean-payload JSON.

src/metrics/datetime.ts
```diff
@@ -0,0 +1,106 @@
...
+  constructor(meta: CommonMetricData, timeUnit: string) {
+    super("datetime", meta);
+
+    this.timeUnit = timeUnitFromString(timeUnit);
+  }
+
+  /**
+   * Sets the metric to a date/time which including the timezone offset.
+   *
+   * @param value Some Data value, with offset, to set the metric to.
```
Dexterp37:
```suggestion
   * @param value Some `Date` value, with offset, to set the metric to.
```

Dexterp37 (changes requested):
My main concern with this PR is the "guessing" part, which uncovered a problem that we'll definitely have with other metric types, too. We need to store extra metadata along with the data we store, and we don't necessarily want to send that. See my comments.

src/metrics/time_unit.ts
brizental: > I agree with you we probably don't need it for DateTime, but that's presumably what we'll want for Timespan and TimingDistribution or Event.

Note that this will give us microsecond precision, nanosecond precision, as far as I know, is out of reach in the browser.

src/metrics/time_unit.ts
brizental: I am not sure I get what you mean here. If you mean that we can transform the date string into a Date object by calling `new Date(dateString)`, sadly no. Some of the strings formats Glean supports work, but micro and nanoseconds don't work, minutes and hours also don't. 

`theDate.toISOString` is not in the format Glean expects either.

src/metrics/time_unit.ts
brizental: I thought about this, but wouldn't it be possible for a person to change the time unit of a given metric? That would render previous entries invalid if we always assume the entries are in the time unit of the Metric instance and we probably wouldn;t want to throw away previous entries because of a time unit change? Or is this not possible / we don't care to delte previous entries if the time unit changes?

src/metrics/time_unit.ts
brizental: We can do [what I did for the Lifetime enum](https://github.com/mozilla/glean.js/blob/main/src/metrics/index.ts#L10-L17) and throw away this function. I had your comment about maybe not using strings to store informatin because that might be a problem performance issue when I made the decision to map the enum to numbers instead of strings. I can change it back to strings though, one less function wouldn't be a bad thing.

src/metrics/time_unit.ts
brizental: Yeah disregard I see the problem with my comment already. If the person changes the time unit, the recorded time unit is still the same... Let me think on this a bit and send a new commit, or come back here with more thoughts.

Please do still let me know about this
> Or is this not possible / we don't care to delete previous entries if the time unit changes?

src/metrics/time_unit.ts
Dexterp37: Yes, that's correct. We get microsecond resolution [according to our docs](https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp).

src/metrics/time_unit.ts
Dexterp37: I was thinking of something like this (pseudocode):

```js
case TimeUnit.Hour: {
  let truncated = new Date(original.getTime());
  truncated.setMinute(0);
  truncated.setSecond(0);
  ...
}
```

As for the `theDate.toISOString`, what do you mean it's not in a format Glean expects? The function seems to return [ISO8601](https://mozilla.github.io/glean/book/dev/core/internal/payload.html#datetime) which is the same thing we [claim to send](https://mozilla.github.io/glean/book/dev/core/internal/payload.html#datetime) :)

src/metrics/time_unit.ts
Dexterp37: I don't have a good answer to this off the top of my head. Can you check what the SDK is doing in this case? I guess the old data gets discarded.

src/metrics/time_unit.ts
brizental: > I was thinking of something like this (pseudocode):

Doing this wouldn't prevent us from doing string manipulation though. The resulting ISO string Javascript generates is always in the milliseconds format even if some of the information is `0`. For reference, this is what `toISOString` output looks like: `"2020-12-18T13:39:34.567Z"`.

> As for the theDate.toISOString, what do you mean it's not in a format Glean expects?

It doesn't include de timezone info for one, it ends with a `Z`. Also it is aways in the `milliseconds` format. As I mention above, we would need to manipulate the resulting string either way.

Dexterp37: Coo, let's ignore this, then! Thanks for looking into it.

src/metrics/time_unit.ts
brizental: On the SDK we store the actual Datetime object. Then we serialize the stored object based on the stored time unit, we don't take into account the time unit of the current instance. We also don't discard anything.

Now that I think about this further, I guess the whole point of recording the datetime **and** the time unit is to be able to serialize it to the correct time unit at the time of recording, thus allowing people to change the time unit without breaking serialization if they want to. Otherwise we could just use the current time unit of the instance to serialize and not persist the time unit at all. Does this make sense?

What I mean is, if we don't want to have to guess the time unit and we also want to allow people to change the time unit if they want to, we must persist the time unit.

brizental: > thus allowing people to change the time unit without breaking serialization if they want to

Not that it would break it on the SDKs case... I think it wouldn't break anything, since it stores the whole Datetime object. But the SDK does make a point about serializing to the recorded and not necessarily current time unit. I will file a documentation bug, the expected behaviour for when people change the time unit is not obvious and it is unclear if it is okay or not.

brizental:
Converting this to a draft because it turns out I have been wrongly assuming that `getISOString` returns the local time, when it actually returns UTC.

brizental:
Closing this in favour of the much simple (and correct) approach in #20.

PR #17, opened 2020-12-18
Bug 1683057 - Implement the UUID metric type

brizental:


src/metrics/uuid.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  /**
+   * Sets to the specified value.
+   *
+   * @param value the value to set.
+   *
+   * @throws In case `value` is not a valid UUID.
+   */
+  async set(value?: string): Promise<void> {
```
brizental, line 46: Yes, this API is different from the SDK API. But all the other metrics work like this, if they have defaults the argument is optional and if it is absent the default is included. Why did we decide on the SDK that specifically for UUID we should have two different APIs `set` and `generate_and_set`?

Is it related to the fact that for `generate_and_Set` we actually return the generated UUID? If so, we probably could unify `set` and `generate_and_set` after https://bugzilla.mozilla.org/show_bug.cgi?id=1673017?

tests/metrics/uuid.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+import assert from "assert";
+import { v4 as UUIDv4 } from "uuid";
+
+import Glean from "glean";
+import UUIDMetric, { isUUIDMetricPayload } from "metrics/uuid";
+import { Lifetime } from "metrics";
+ 
+describe("UUIDMetric", function() {
+  beforeEach(async function() {
+    await Glean.resetGlean();
```
Dexterp37, line 14: Mh, should we rename this to `testResetGlean` so that the intent is clearer.

Dexterp37: (Can be done as a follow-up PR).

src/metrics/uuid.ts
Dexterp37: Please let's stick to the documented API for now: let's have `set` and `generateAndSet` and then change accordingly to the result of [the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1673017) when needed.

Dexterp37 (changes requested):

tests/metrics/uuid.spec.ts
brizental: Eh, no need for a follow up this was just search and replace took 1s.

src/metrics/uuid.ts
brizental: Done................

Dexterp37: > 
> 
> Why did we decide on the SDK that specifically for UUID we should have two different APIs `set` and `generate_and_set`?

I noticed I forgot to answer this: this API was decided on mainly to power the client id case (and similar ones).

Dexterp37 (approved):

PR #19, opened 2021-01-04
Bug 1683846 - Decouple metrics recorded format from expected ping payload format

brizental:


src/metrics/validation.ts
```diff
@@ -0,0 +1,48 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
brizental: Cannot put this code in `metrics/index.ts` because it will cause circular import errors.

src/database.ts
```diff
@@ -40,7 +40,30 @@ export function isValidPingPayload(v: unknown): v is PingPayload {
...
+ *
+ * Metrics are stored in an internal representation format,
+ * but pings must be sent with metrics in their external format
+ * (i.e. the format described in the Glean schema).
+ *
+ * @param v The ping payload to transform.
+ *
+ * @returns The ping payload in its external format.
+ */
+function createExternalMetricsPayload(v: MetricsPayload): MetricsPayload {
```
Dexterp37, line 56: I don't think I like the idea of having "two payloads" (internal vs external). There's no such thing as "internal payload", it's just metrics data in the end. Payload as a concept only makes sense when it comes to pings.

The internal data should probably just be "storedMetrics" and the payload... well, is the payload .

brizental: Yeap I agree, names chosen here are terrible. I'll try to think of something better and get back here. Suggestions 100% welcome.

brizental: I tried something (https://github.com/mozilla/glean.js/pull/19/commits/d2e5ab36cdce913aca29c644c68137949aee9cc4), let me know what you think.

src/database.ts
```diff
@@ -40,7 +40,30 @@ export function isValidPingPayload(v: unknown): v is PingPayload {
   } else {
     return false;
   }
+}
 
+/**
+ * Creates an external metrics payload from an internal metrics payload.
```
Dexterp37, line 46: This comment needs an update :=)

src/metrics/index.ts
```diff
@@ -2,8 +2,76 @@
...
 
+/**
+ * The Metric class describes the shared behaviour amongst concrete metrics.
+ *
+ * A concrete metric will always have two possible representations:
+ *
+ * - `InternalRepresentation`: is the format in which this metric will be stored in memory.
+ * This format may contain extra metadata, in order to allow deserializing of this data for testing purposes.
+ * - `PayloadRepresentation`: is the format in which this metric will be represented in the ping payload.
+ * This format must be the exact same as described in [the Glean schema](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/master/schemas/glean/glean/glean.1.schema.json).
```
Dexterp37, line 16:
```suggestion
 * - `InternalRepresentation`: is the format in which this metric will be stored in memory.
 *    This format may contain extra metadata, in order to allow deserializing of this data for testing purposes.
 * - `PayloadRepresentation`: is the format in which this metric will be represented in the ping payload.
 *    This format must be the exact same as described in [the Glean schema](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/master/schemas/glean/glean/glean.1.schema.json).
```

Dexterp37 (changes requested):

Dexterp37 (approved):

PR #20, opened 2021-01-04
Bug 1679381 - Implement the datetime metric type

brizental:
This is built on top of https://github.com/mozilla/glean.js/pull/19 thus it is blocked by it.

I am opening anyways to illustrate how the changes added by https://github.com/mozilla/glean.js/pull/19 can be used in a more complex way than the cases in https://github.com/mozilla/glean.js/pull/19 itself.

The commit that matters here is fb5c0e8. All the other are part of #19.


src/metrics/datetime.ts
```diff
@@ -0,0 +1,226 @@
...
+  get date(): Date {
+    return new Date(this._inner.date);
+  }
+
+  private get timezone(): number {
+    return this._inner.timezone;
+  }
+
+  private get timeUnit(): TimeUnit {
+    return this._inner.timeUnit as TimeUnit;
```
brizental, line 63: No need for this casting here... This is leftover from development.

src/metrics/datetime.ts
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  validate(v: unknown): v is DatetimeInternalRepresentation {
+    if (!isObject(v) || Object.keys(v).length !== 3) {
+      return false;
+    }
+
+    const timeUnitVerification = "timeUnit" in v && isString(v.timeUnit) && Object.values(TimeUnit).includes(v.timeUnit as TimeUnit);
+    const timezoneVerification = "timezone" in v && isNumber(v.timezone);
```
brizental, line 72: Unsure if I should go further and check that timezone is a possible timezone value e.g. between -12 * 60 and 14 * 60.

src/metrics/datetime.ts
```diff
@@ -0,0 +1,226 @@
...
+
+  /**
+   * Gets this metrics value in its payload representation.
+   *
+   * For this metric, the payload is the timezone aware ISO date string truncated to the time unit
+   * given at the time of recording.
+   *
+   * @returns The metric value.
+   */
+  payload(): string {
```
brizental, line 89: This function turned out being the uglyest... I am sorry and open for suggestions on how to make it nicer :)

src/metrics/datetime.ts
```diff
@@ -0,0 +1,226 @@
...
+    if (this.timeUnit === TimeUnit.Microsecond) {
+      return `${year}-${month}-${day}T${hours}:${minutes}:${seconds}.${milliseconds}000${timezone}`;
+    }
+
+    return `${year}-${month}-${day}T${hours}:${minutes}:${seconds}.${milliseconds}000000${timezone}`;
+  }
+}
+
+/**
+ *  A boolean metric.
```
brizental, line 128: Ops, forgot to update comment.

src/metrics/datetime.ts
```diff
@@ -0,0 +1,226 @@
...
+   * If the currently stored datetime is in a different timezone than local,
+   * the resulting Date object will contain the relative local date to the recorded value.
+   *
+   * TODO: Only allow this function to be called on test mode (depends on Bug 1682771).
+   *
+   * @param ping the ping from which we want to retrieve this metrics value from.
+   *
+   * @returns The value found in storage or `undefined` if nothing was found.
+   */
+  async testGetValue(ping: string): Promise<Date | undefined> {
```
brizental, line 220: On the Glean SDK we zero out information that is too precise for the current time unit (see: https://github.com/mozilla/glean/blob/main/glean-core/src/metrics/datetime.rs#L161). I am not doing it here, seemed like superfluous work for testing APIs. Do you have issues with this choice?

src/metrics/datetime.ts
Dexterp37: Mh, no, let's not for now. In the future, it might be worth reporting an error.

src/metrics/datetime.ts
```diff
@@ -0,0 +1,227 @@
...
+   *
+   * For this metric, the payload is the timezone aware ISO date string truncated to the time unit
+   * given at the time of recording.
+   *
+   * @returns The metric value.
+   */
+  payload(): string {
+    const timezone = formatTimezoneOffset(this.timezone);
+
+    const year = this.date.getFullYear().toString().padStart(2, "0");
```
Dexterp37, line 92: Another option for all these date/time manipulation insanities would be to use [`toLocaleDateString`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString) and provide `en-US` as the locale (since that's the one we stick to). Did you try that out already?

src/metrics/datetime.ts
```diff
@@ -0,0 +1,227 @@
...
+  constructor(meta: CommonMetricData, timeUnit: string) {
+    super("datetime", meta);
+    this.timeUnit = timeUnit as TimeUnit;
+  }
+
+  /**
+   * Sets to the specified datetime value.
+   *
+   * @param value the value to set.
+   */
```
Dexterp37, line 145: Can you please mention the same things we mention in the [other bindings](https://searchfox.org/glean/rev/58fa076222be375a4e5f46baa6275043721603c0/glean-core/android/src/main/java/mozilla/telemetry/glean/private/DatetimeMetricType.kt#54-58) for all the docstrings ?

src/metrics/datetime.ts
```diff
@@ -0,0 +1,227 @@
...
+  async set(value?: Date): Promise<void> {
+    if (!this.shouldRecord()) {
+      return;
+    }
+
+    if (!value) {
+      value = new Date();
+    }
+
+    const metric =  new DatetimeMetric({
```
Dexterp37, line 155:
```suggestion
    const metric = new DatetimeMetric({
```

Does eslint support a rule to detect "double spacing"?

src/metrics/datetime.ts
Dexterp37: This makes writing tests using this impossible/harder :-) We can't check if a date we set is then returned properly, after truncation.

Don't we have a test specifically for testing this, in the SDK?

src/metrics/time_unit.ts
```diff
@@ -0,0 +1,26 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * Different resolutions supported by the time related
+ * metric types (e.g. DatetimeMetric).
+ */
+enum TimeUnit {
+  // Truncate to nanosecond precision.
```
Dexterp37, line 10: This does not represent a truncation, necessarily! See https://searchfox.org/glean/source/glean-core/android/src/main/java/mozilla/telemetry/glean/private/TimeUnit.kt :-)

src/metrics/datetime.ts
Dexterp37: Oh yes, we do: https://searchfox.org/glean/source/glean-core/tests/datetime.rs

tests/metrics/datetime.spec.ts
```diff
@@ -0,0 +1,240 @@
...
+import assert from "assert";
+import sinon from "sinon";
+
+import Glean from "glean";
+import DatetimeMetricType, { DatetimeMetric } from "metrics/datetime";
+import { Lifetime } from "metrics";
+
+const sandbox = sinon.createSandbox();
+
+describe("DatetimeMetric", function() {
```
Dexterp37, line 14: Can you kindly port over the tests from glean-core, too? https://searchfox.org/glean/source/glean-core/tests/datetime.rs

Don't forget that in the SDK they are split between the bindings and the core, but here everything is together so tests can be together too.

Dexterp37 (changes requested):

src/metrics/datetime.ts
brizental: Truncating in Glean.js happens when we call the `payload` function on a concrete Datetime metric. I added tests for each one of the time units truncations. This specific function returns a Date object, no truncation happens here.

tests/metrics/datetime.spec.ts
brizental: I based myself off of these tests. Which one do you see missing? I did not check the bindings tests though, will do that too.

src/metrics/datetime.ts
brizental: toLocalteDateString does not return an ISO string.

tests/metrics/datetime.spec.ts
brizental: I checked bidnings tests and added some extra date cases (prior to UNIX epoch  and date after 2038), other than that I did not see any other missing test cases.

src/metrics/datetime.ts
Dexterp37: Isn't it easier to convert it to that format starting from what toLocaleDateString gives? Alternatively, we could also try to just use `Date().toISOString()` and split/crop. Otherwise, we can leave it as it is.

tests/metrics/datetime.spec.ts
```diff
@@ -0,0 +1,252 @@
...
+        sendInPings: ["aPing"],
+        lifetime: Lifetime.Ping,
+        disabled: false
+      }, "second");
+
+      await metric.set(new Date(1995, 5, 25, 8, 15, 45, 385));
+      assert.strictEqual(await metric.testGetValueAsString("aPing"), "1995-05-25T08:15:45+05:00");
+
+      const snapshot = await Glean.db.getPing("aPing", true);
+      assert.deepStrictEqual(snapshot, {
```
Dexterp37, line 158: This tests something a bit different than [this test](https://searchfox.org/glean/source/glean-core/tests/datetime.rs#110). You're testing the snapshot, which ideally users won't have access to. Users will have to use the testing APIs, which currently do not truncate.

brizental: `testGetValueAsString` does truncate because it returns `DatetimeMetric.payload`, only when we get it as a Date object it doesn't truncate. This means that, if the metric is `TimeUnit.Hours` the `Date` object returned by `testGetValue` may come with minute, seconds and etc not zeroed out. Do you think that is a problem?

About this specific test, I can change to do exactly what the SDK does.

Dexterp37: Yes, I think this is a problem (sorry, that's what I was trying to say). Using the testing API returns a different result than what user would expect.

As for the tests, in general I'd recommend sticking to the same tests we have for the SDK, to make sure we're not introducing problems in the tests.

src/metrics/datetime.ts
brizental: So, the issue with toISOString is that it always returns a string in UTC, so we lose timezone information. It also wouldn't be easier to split by toLocaleDateString, IMO, but even if it were, it doesn't contain all the information that we can get directly from the date object, such as timezone information.

I did spot a bug here, if the user changes timezones from the moment they recorded the metric to the moment they access the metrics payload, we create an incorrect string... I will fix this in my next commit.

brizental: Fixed it https://github.com/mozilla/glean.js/pull/20/commits/065f63f01a05faeacbfbb05c6a262f66a25aaddf.

src/metrics/datetime.ts
brizental: I now zero out everything that is too precise at the time of recording: https://github.com/mozilla/glean.js/pull/20/commits/11d9387b95a7356a55f06b0283e83743b2e2d2af

Dexterp37 (approved):

PR #23, opened 2021-01-12
Bug 1677443 - Implement initialize and core metrics

brizental:
This is mostly a continuation of my previous PR (#22), becuase I started implementing it accidentally on that work and then realized it was not part of that PR.

Anyways, `initialize` is much leaner right now than it is in the Android implementation (I've based myself off of that one). It will simply sanitize and set the app id and initialize the core metrics. That, I think, is enough for us right now, concepts such as the dirty flag I am not even sure apply to Glean.js and other stuff related to flushing events or sending deletion request pings need to be implemented yet.

This branch is on top of the branch from #22, so this PR is currently blocked until the other is merged.


src/platform_info/webext.ts
```diff
@@ -0,0 +1,47 @@
...
+  },
+
+  async arch(): Promise<string> {
+    const platformInfo = await browser.runtime.getPlatformInfo();
+    // Possible values are: "arm", "x86-32", "x86-64"
+    return platformInfo.arch;
+  },
+
+  async locale(): Promise<string> {
+    return Promise.resolve(navigator.language || "Unknown");
```
Dexterp37, line 44: Instead of `Unknown` this should return `und` in case of problems, like the other bindings. See [our docs](https://mozilla.github.io/glean/book/user/pings/index.html#the-client_info-section).

tests/pings/index.spec.ts
```diff
@@ -12,6 +12,7 @@ import Glean from "glean";
 describe("PingType", function() {
   beforeEach(async function() {
     await Glean.testRestGlean();
```
Dexterp37, line 14: GAH! Darn, typo? `Rest` -> `Reset`? Let's change this in a follow-up PR (not here please)

tests/pings/index.spec.ts
```diff
@@ -12,6 +12,7 @@ import Glean from "glean";
...
+    await Glean.initialize("something something");
```
Dexterp37, line 15: The initialization should be part of `testResetGlean`, as for the other bindings :)

Dexterp37 (changes requested):

src/glean.ts
```diff
@@ -99,16 +122,36 @@ class Glean {
...
+   * **Test-only API**
+   *
+   * Resets the Glean singleton to its initial state and re-initializes it.
+   *
+   * TODO: Only allow this function to be called on test mode (depends on Bug 1682771).
+   *
+   * @param applicationId The application ID (will be sanitized during initialization).
+   * @param optionalInitializeArgs Optional arguments to be passed to `initialize`.
+   */
+  static async testRestGlean(applicationId: string, ...optionalInitializeArgs: never[]): Promise<void> {
```
Dexterp37, line 144: Can you file a bug for renaming this so that it doesn't get lost?

Dexterp37 (approved):

src/glean.ts
brizental: https://bugzilla.mozilla.org/show_bug.cgi?id=1686452

PR #24, opened 2021-01-15
Bug 1677440 - Implement a ping uploader

brizental:
Note that there are no _actual_ uploading tests in this PR. I thought it was big enough as it is. 

Will write the remaning tests + implement a fetch API adapter while this review is going on.


src/config.ts
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * The Configuration class describes how to configure Glean.
+ */
+interface Configuration {
+  // The build identifier generated by the CI system (e.g. "1234/A").
+  readonly appBuild?: string,
+  // The user visible version string fro the application running Glean.js.
+  readonly appDisplayVersion?: string,
+  // The server pings are sent to.
+  readonly serverEndpoint: URL
```
brizental, line 14: Hm, this should be optional.

src/glean.ts
```diff
@@ -158,6 +170,10 @@ class Glean {
     Glean.instance._initialized = false;
     // Reset upload enabled state, not to inerfere with other tests.
     Glean.uploadEnabled = true;
+
+    // Stop ongoing upload jobs.
```
brizental, line 174: This comment is wrong. Should be "Stop ongoing jobs and clear pending pings queue".

tests/upload.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+  }
+}
+
+/**
+ * Disables the uploader on the Glean singleton,
+ * so that it doesn't interefe with tests.
+ */
+function disableGleanUploader(): void {
+  // Stub the trigger upload on the Glean singleton's ping uploader,
+  // so that it doesn't interfere with this test.
```
brizental, line 59: This comment is unnecessary.

src/upload/adapter/index.ts
```diff
@@ -0,0 +1,30 @@
...
+export interface UploadAdapter {
+  /**
+   * Makes a POST request to a given url, with the given headers and body.
+   *
+   * @param url The URL to make the POST request
+   * @param options The optional headers and body to include in the request.
+   *
+   * @returns The status code of the response.
+   */
+  post(url: URL, options: {
```
Dexterp37, line 17: Do you expect this to have more than headers and body? Does an `options` object really buy us much?

src/upload/adapter/index.ts
```diff
@@ -0,0 +1,30 @@
...
+    body: string
+  }): Promise<number>;
+}
+
+// Default export for tests sake.
+const MockUploadAdapter: UploadAdapter = {
+  post(): Promise<number> {
+    return Promise.resolve(200);
+  }
+};
```
Dexterp37, line 28: Can we do this in test files?

src/upload/index.ts
```diff
@@ -0,0 +1,284 @@
...
+  } {
+    const stringifiedBody = JSON.stringify(ping.payload);
+
+    let headers = ping.headers || {};
+    headers = {
+      ...ping.headers,
+      "Content-Type": "application/json; charset=utf-8",
+      "Content-Length": stringifiedBody.length.toString(),
+      "Date": (new Date()).toISOString(),
+      "X-Client-Type": "Glean.js",
```
Dexterp37, line 103: You should check with @fbertsch if we're doing any special magic on this header when ingesting.

src/upload/index.ts
```diff
@@ -0,0 +1,284 @@
...
+      await Glean.pingsDatabase.deletePing(identifier);
+      return false;
+    case (status && status >= 400 && status < 500):
+      console.warn(`Unrecoverable upload failure while attempting to send ping ${identifier}. Error was ${status}.`);
+      await Glean.pingsDatabase.deletePing(identifier);
+      return false;
+    default:
+      console.warn(`Recoverable upload failure while attempting to send ping ${identifier}, will retry. Error was ${status}.`);
+      return true;
+    }
```
Dexterp37, line 185: That's an interesting pattern :-) Do we really need a "fake switch" here? It's really 2-3 ifs. We can probably get away by checking `if (!status)` at the beginning and returning accordingly. Then checking the values.

Why wouldn't we have a `status` anyway?

src/upload/index.ts
```diff
@@ -0,0 +1,284 @@
...
+   *    Otherwise it will reach maximum retriable failures and bail out.
+   */
+  async triggerUpload(): Promise<void> {
+    if (this.status !== PingUploaderStatus.Idle) {
+      return;
+    }
+
+    this.status = PingUploaderStatus.Uploading;
+    this.currentJob = this.triggerUploadInternal();
+    await this.currentJob;
```
Dexterp37, line 226: If this throws then the status remains stuck to `PingUploaderStatus.Uploading`, right? We should probably catch any error here, if we want to safely transition the states.

src/upload/index.ts
```diff
@@ -0,0 +1,284 @@
...
+ * The possible status of the ping uploader.
+ */
+const enum PingUploaderStatus {
+  // Not currently uploading pings.
+  Idle,
+  // Currently uploading pings.
+  Uploading,
+  // Currently processing a signal to stop uploading pings.
+  Cancelling,
+}
```
Dexterp37, line 24: Any specific reason why we're moving away from the status names (DONE, WAIT, UPLOAD) from glean-core?

src/config.ts
```diff
@@ -0,0 +1,17 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * The Configuration class describes how to configure Glean.
+ */
+interface Configuration {
```
Dexterp37, line 8: Will you file a bug for allowing custom uploaders?

Dexterp37 (changes requested):

src/upload/adapter/index.ts
brizental: Not really. We need a default export from this index file for tests to work. I've done this for all modules that have platform specific implementation. Webpack is the one that knows about the platform specific implementation, mocha just expects that the index file will have a default export.

src/upload/adapter/index.ts
brizental: Nope, this signature is compatible with the fetch API for no strong reason. I can change it.

src/upload/index.ts
brizental: Good point, I actually copied these from the POC. Asked on Riot.

src/upload/index.ts
brizental: > Why wouldn't we have a status anyway?

This is a question I plan to answer on my next PR, when I implement a real upload adapter for the browser. For now not sure, but I am thinking if no internet connection or timeout will result in `undefined` status. (will we even have timeout? fetch API doesnj't have IIRC).

> That's an interesting pattern :-) Do we really need a "fake switch" here?

Yeah this the trick to have conditios on each switch clause for Javascript. Honestly, really only used a switch here because that is what we do in glean-core. I can change, the main argument for not changing would be the fact that the function would look much different than the glean-core function. But who really cares? The module is already pretty different anyways.

src/upload/index.ts
brizental: Good point. I can't think of a situation where `triggerUploadInternal` could throw, but I am sure it can, heh.

src/upload/index.ts
brizental: These enums are not the same. This enum is about the status of the Glean.js PingUploader, whether or not an uploading job is happening right now. The enum in glean-core is an uploading task for a language binding, whether it should come back later, upload a new request or be done.

I would say it is even better that the names are not same, so as not to cause confusion.

Honestly, I should probably file a follow up to document how this uploader works, like we do for glean-core (https://mozilla.github.io/glean/book/dev/core/internal/upload.html). The uploaders are similar but their inner workings are different. There are enough similarities and differences to be super confusing.

src/config.ts
brizental: Yes, this configuration is not in any way complete.

src/upload/index.ts
Dexterp37: Yes, it seems like a good idea.

Dexterp37 (approved):

src/upload/index.ts
brizental: https://bugzilla.mozilla.org/show_bug.cgi?id=1687288

src/config.ts
brizental: https://bugzilla.mozilla.org/show_bug.cgi?id=1687287

PR #25, opened 2021-01-18
Bug 1677440 - Implement a browser upload adapter

brizental:
Adding tests for the timeout mechanism and cookie removal is blocked because we need to figure out the mest way _bot_ to mock fetch on these tests.

Filed [a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1687322) to do that later, for now I manually tested and it "works on my machine"™.

### DIFF
--- a/src/upload/adapter/index.ts
+++ b/src/upload/adapter/index.ts
@@ -2,10 +2,15 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+import { UploadResult, UploadResultStatus } from "upload";
+
 /**
  * Uploader interface, actualy uploading logic varies per platform.
  */
-export interface UploadAdapter {
+export abstract class UploadAdapter {
+  // The timeout, in seconds, to use for all operations with the server.
+  protected defaultTimeout = 10_000;
+
   /**
    * Makes a POST request to a given url, with the given headers and body.
    *
@@ -15,14 +20,18 @@ export interface UploadAdapter {
    *
    * @returns The status code of the response.
    */
-  post(url: URL, body: string, headers?: Record<string, string>): Promise<number>;
+  abstract post(url: URL, body: string, headers?: Record<string, string>): Promise<UploadResult>;
 }
 
 // Default export for tests sake.
-const MockUploadAdapter: UploadAdapter = {
-  post(): Promise<number> {
-    return Promise.resolve(200);
+class MockUploadAdapter extends UploadAdapter {
+  post(): Promise<UploadResult> {
+    const result: UploadResult = {
+      result: UploadResultStatus.Success,
+      status: 200
+    };
+    return Promise.resolve(result);
   }
-};
+}
 
-export default MockUploadAdapter;
+export default new MockUploadAdapter();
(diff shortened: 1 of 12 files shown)


src/upload/adapter/index.ts
```diff
@@ -2,10 +2,15 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+import { UploadResult, UploadResultStatus } from "upload";
+
 /**
  * Uploader interface, actualy uploading logic varies per platform.
  */
-export interface UploadAdapter {
+export abstract class UploadAdapter {
```
Dexterp37, line 10: Looking at this again.. why `Adapter`? In pattern-land, it means a very specific thing

brizental: When I was doing the POC I read through axios code to see how they dealt with node vs. browser differences and stumbled upon the adapter terminology (https://github.com/axios/axios/tree/master/lib/adapters). 

No strong opinions about changing it, though. How about `UploadMechanism`? Or maybe `Upload`/`Uploader`? Open for suggestions.

Dexterp37: I'd rather stick to `Uploader` if you don't mind :-) Sorry!

tests/upload/adapter.spec.ts
```diff
@@ -2,14 +2,49 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+import "jsdom-global/register";
```
brizental, line 5: I literally only have this because it includes `DOMException`. One less thing for me to mock.

tests/upload/adapter.spec.ts
```diff
@@ -2,14 +2,49 @@
...
 import assert from "assert";
 import sinon from "sinon";
 
 import { UploadResultStatus } from "upload";
 import BrowserUploadAdapter from "upload/adapter/browser";
+import { JSONObject } from "utils";
 
 const sandbox = sinon.createSandbox();
 
+global.fetch = () => {
```
brizental, line 15: I know this code is hacky AF. But the bug I opened to look into mocking fetch is what should fix it.

src/upload/uploader/index.ts
```diff
@@ -2,10 +2,15 @@
...
 
+import { UploadResult, UploadResultStatus } from "upload";
+
 /**
  * Uploader interface, actualy uploading logic varies per platform.
  */
-export interface UploadAdapter {
+export abstract class Uploader {
+  // The timeout, in seconds, to use for all operations with the server.
+  protected defaultTimeout = 10_000;
```
brizental, line 12: I copied this from the python bindings, on the android bindings we have two different timeouts: one for connection timeout and another for request timeout. I don't know that there is a way to make this differentiation here.

Dexterp37: It's fine to not have the differentiation if the uploader doesn't allow it :)

src/upload/uploader/index.ts
```diff
@@ -15,14 +20,18 @@ export interface UploadAdapter {
    *
    * @returns The status code of the response.
    */
-  post(url: URL, body: string, headers?: Record<string, string>): Promise<number>;
+  abstract post(url: URL, body: string, headers?: Record<string, string>): Promise<UploadResult>;
 }
 
 // Default export for tests sake.
```
Dexterp37, line 26: Can you add a quick note (you mentioned that in a comment) about why we need a default export and why that has to be a test mock?

Dexterp37 (approved):
r+wc

PR #26, opened 2021-01-19
Bug 1677444 - Implement upload status toggling logic

brizental:


src/glean.ts
```diff
@@ -59,30 +69,124 @@ class Glean {
...
+    await Glean.metricsDatabase.clearAll();
+    await Glean.pingsDatabase.clearAll();
+
+    // We need to briefly set upload_enabled to true here so that `set`
+    // is not a no-op.
+    //
+    // Note that we can't provide the same guarantees as glean-core here.
+    // If by any change another actor attempts to record a metric while
+    // we are setting the known client id and first run date, they will be allowed to.
+    Glean.uploadEnabled = true;
```
Dexterp37, line 137: This sounds really bad. This is probably something the dispatcher would guard us against. Since we're not using this library yet, I'm fine with living with this, as long as we take the dispatcher bug right after this one.

Dexterp37: Let's add a TODO string here referencing the dispatcher's bug number.

src/glean.ts
```diff
@@ -118,31 +222,66 @@ class Glean {
...
+   * When enabling, the core Glean metrics are recreated.
+   *
+   * If the value of this flag is not actually changed, this is a no-op.
+   *
+   * @param flag When true, enable metric collection.
+   *
+   * @returns Whether the flag was different from the current value,
+   *          and actual work was done to clear or reinstate metrics.
+   */
+  static async setUploadEnabled(flag: boolean): Promise<boolean> {
```
Dexterp37, line 273: Why do we need to return a boolean here? We're increasing the API surface and introducing a different semantics wrt the other bindings.

src/glean.ts
```diff
@@ -118,31 +222,66 @@ class Glean {
...
   }
 
+  /**
+   * Determines whether upload is enabled.
+   *
+   * When upload is disabled, no data will be recorded.
+   *
+   * @returns Whether upload is enabled.
+   */
+  static isUploadEnabled(): boolean {
```
Dexterp37, line 248: This should only be available for internal used, should not be exposed to the outside.

src/glean.ts
```diff
@@ -59,30 +69,124 @@ class Glean {
     return Glean.instance._coreMetrics;
   }
 
-  private static get instance(): Glean {
-    if (!Glean._instance) {
-      Glean._instance = new Glean();
-    }
+  private static get uploadEnabled(): boolean {
+    return Glean.instance._uploadEnabled || false;
+  }
```
Dexterp37, line 75: This should only be available for internal used, should not be exposed to the outside.

src/glean.ts
```diff
@@ -59,30 +69,124 @@ class Glean {
...
 
-    return Glean._instance;
+  private static set uploadEnabled(value: boolean) {
```
Dexterp37, line 76: Any specific reason why we're adding accessors here? We're basically adding another API.

src/glean.ts
```diff
@@ -59,30 +69,124 @@ class Glean {
...
+   * Should only be called when the state actually changes.
+   *
+   * A deletion_request ping is sent, all pending metrics, events and queued
+   * pings are cleared, and the client_id is set to KNOWN_CLIENT_ID.
+   * Afterward, the upload_enabled flag is set to false.
+   */
+  private static async onUploadDisabled(): Promise<void> {
+    Glean.uploadEnabled = false;
+    await Glean.clearMetrics();
+  }
```
Dexterp37, line 104: I think it would make more sense, for the sake of comparing the logic with the existing implementation, to follow the SDK tracks and have this as a single function (see https://searchfox.org/glean/source/glean-core/rlb/src/lib.rs#378 )

src/glean.ts
```diff
@@ -118,31 +222,66 @@ class Glean {
...
-  // TODO: Make the following functions `private` once Bug 1682769 is resolved.
-  static get uploadEnabled(): boolean {
-    return Glean.instance._uploadEnabled;
-  }
-
-  static set uploadEnabled(value: boolean) {
-    Glean.instance._uploadEnabled = value;
-  }
-
   static get applicationId(): string | undefined {
```
Dexterp37, line 225: Should this be private?

Dexterp37 (changes requested):

src/glean.ts
brizental: Yes, indeed it is terrible and I agree the dispatcher bug should fix it. I am not totally sure it will though.

My plan was to take it right after this one, yes :) I'll leave a comment.

src/glean.ts
brizental: Honestly no idea, just copying glean-core on this one.

src/glean.ts
brizental: That is fine, this one is private :)

src/glean.ts
brizental: It is nice to have this public here, so that other parts of the code don't need to access this information in a kinda hacky way. Note that the Glean singleton is not exposed to clients anyways, we only expose initialize and setUploadEnabled methodd, but hide the singleton.

src/glean.ts
brizental: Just so that I don't need do `Glean.instance._setUploadEnabled` all the time. I did this for all the other properties as well. It is not a new API because it is not public.

src/glean.ts
Dexterp37: Yup, but for the public API you should really look at the RLB :-)

src/glean.ts
brizental: I implemented a `setUploadEnabled` function as well. This Glean singleton is more akin to the Glean object on glean's rust core than the Glean object from the bindings. The implementation here follows _exactly_ what we are doing on the glean-core. We will expose a different `setUploadEnabled` function to clients, that hides the singelton logic, see: https://github.com/mozilla/glean.js/blob/main/src/index.ts#L23

src/glean.ts
brizental: Ideally not, we are using it on the ping maker.

src/glean.ts
brizental: It is convenient for testing, but that is the only advantage. I thought it was weird as well upon implementing.

brizental: > Yup, but for the public API you should really look at the RLB :-)

Like I mention in https://github.com/mozilla/glean.js/pull/26#discussion_r560815291 this is not _really_ the public API.

Dexterp37 (approved):

PR #27, opened 2021-01-21
Bug 1687644 - Implement the external API for consumers

brizental:
Now that we can acutally measure it, for reference:

- The webext build is 46K.
- The qt build is 44K.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,9 +9,9 @@
     "lint": "eslint . --ext .ts,.js,.json --max-warnings=0",
     "fix": "eslint . --ext .ts,.js,.json --fix",
     "build:webext": "webpack --config webpack.config.webext.js --mode production",
-    "dev:webext": "webpack --watch --config webpack.config.webext.js --mode development",
+    "dev:webext": "webpack --watch --config webpack.config.webext.js --mode development --devtool inline-source-map",
     "build:qt": "webpack --config webpack.config.qt.js --mode production",
-    "dev:qt": "webpack --watch --config webpack.config.qt.js --mode development",
+    "dev:qt": "webpack --watch --config webpack.config.qt.js --mode development  --devtool inline-source-map",
     "build:test-webext": "cd tests/utils/webext/sample/ && npm install && npm run build:xpi"
   },
   "repository": {
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,25 +2,81 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+import Glean from "glean";
+import Configuration from "config";
+
+// Private stuff to export.
+import PingType from "pings";
+import BooleanMetricType from "metrics/types/boolean";
+import CounterMetricType from "metrics/types/counter";
+import DatetimeMetricType from "metrics/types/datetime";
+import StringMetricType from "metrics/types/string";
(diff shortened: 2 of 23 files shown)


brizental:
The Qt error seems unrelated. I will verify, but don't feel that it blocks the review :)

package.json
```diff
@@ -9,9 +9,9 @@
     "lint": "eslint . --ext .ts,.js,.json --max-warnings=0",
     "fix": "eslint . --ext .ts,.js,.json --fix",
     "build:webext": "webpack --config webpack.config.webext.js --mode production",
-    "dev:webext": "webpack --watch --config webpack.config.webext.js --mode development",
+    "dev:webext": "webpack --watch --config webpack.config.webext.js --mode development --devtool inline-source-map",
```
Dexterp37, line 12: Nice move :) I'll borrow this for Rally.

src/index.ts
```diff
@@ -2,25 +2,81 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+import Glean from "glean";
+import Configuration from "config";
+
+// Private stuff to export.
```
Dexterp37, line 8:
```suggestion
// Private Glean types to export.
```

src/storage/utils.ts
```diff
@@ -23,7 +23,10 @@ export function getValueFromNestedObject(obj: JSONObject, index: StorageIndex):
   let target: JSONValue = obj;
   for (const key of index) {
     if (isObject(target) && key in target) {
-      target = target[key];
+      const temp: unknown = target[key];
+      if (isJSONValue(temp)) {
```
Dexterp37, line 27: Did you stumble upon non-JSON values being stored?

src/upload/uploader/index.ts
```diff
@@ -34,7 +34,11 @@ export abstract class Uploader {
 //
 // TODO: remove this comment once Bug 1687516 is resolved.
 class MockUploader extends Uploader {
-  post(): Promise<UploadResult> {
+  // If we don't include these arguments here, the typescript checker will raise
+  // `TS2554: Expected 0 arguments, but got 3.`.
+  //
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
Dexterp37, line 40: It makes sense since the abstract class has them. Do we need to specify this?

src/storage/weak.ts
```diff
@@ -14,7 +14,11 @@ import { JSONObject, JSONValue } from "utils";
 class WeakStore implements Store {
   private store: JSONObject;
 
-  constructor() {
+  // If we don't include these arguments here, the typescript checker will raise
+  // `TS2554: Expected 0 arguments, but got 1.`.
+  //
```
Dexterp37, line 19: It makes sense since the abstract class has it. Do we need to specify this?

Dexterp37 (changes requested):

src/upload/uploader/index.ts
brizental: Eh, you are right. I'll remove the comment.

src/index.ts
brizental: Heh, much nicer.

brizental:
The Qt problems were actually related. We were using some browser APIs namely URL and crypto, which I didn't realize before. That was fixed. This is ready for a new look.

Dexterp37 (approved):

PR #28, opened 2021-01-21
Bug 1687644 - Use the new external API on the web ext sample

brizental:
This is a continuation of #27.

Glean.js works :) See the pings sent from the sample: https://glean-debug-view-dev-237806.firebaseapp.com/pings/glean.js-test. Chrome and Firefox, but you'll need to take my word for it, that information is not in the metrics yet.

I did encounter some bugs upon testing in Chrome, which were not major so I fixed them. I think it will be important to have a kinda real testing playground for the next task I'll tackle which is [Bug 1687491](https://bugzilla.mozilla.org/show_bug.cgi?id=1687491).

It is also clear that in the future we will need to have automated testing in both browsers, the errors I fixed in the storage I only found because I tested in Chrome.

### DIFF
--- a/samples/web-extension/index.js
+++ b/samples/web-extension/index.js
@@ -5,25 +5,11 @@
 "use strict";
 
 import Glean from "glean";
+import { samplePing } from "./generatedPings";
+import { webExtStarted, popupOpened } from "./generatedMetrics";
 
-// TODO: Once we have glean_parser generating pings and metrics,
-// remove this code.
-const samplePing = new Glean._private.PingType("sample", true, false);
-const webExtStarted = new Glean._private.DatetimeMetricType({
-  category: "sample",
-  name: "webext_installed",
-  sendInPings: ["sample"],
-  lifetime: "ping",
-  disabled: false
-}, "millisecond");
-const popupOpened = new Glean._private.CounterMetricType({
-  category: "sample",
-  name: "popup_opened",
-  sendInPings: ["sample"],
-  lifetime: "ping",
-  disabled: false
-}, "millisecond");
-
+// TODO: Do not wait for Glean to be initialized before recording metrics
+// once Bug 1687491 is resolved.
 Glean.initialize("web-extension", true)
   .then(() => {
     console.log("Glean has been succesfully initialized.");
--- a/samples/web-extension/index.js
+++ b/samples/web-extension/index.js
@@ -6,4 +6,46 @@
 
 import Glean from "glean";
 
-Glean.initialize("web-extension", true);
+// TODO: Once we have glean_parser generating pings and metrics,
+// remove this code.
+const samplePing = new Glean._private.PingType("sample", true, false);
+const webExtStarted = new Glean._private.DatetimeMetricType({
+  category: "sample",
+  name: "webext_installed",
+  sendInPings: ["sample"],
+  lifetime: "ping",
+  disabled: false
+}, "millisecond");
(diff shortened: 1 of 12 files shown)


samples/web-extension/index.js
```diff
@@ -6,4 +6,46 @@
 
 import Glean from "glean";
 
-Glean.initialize("web-extension", true);
+// TODO: Once we have glean_parser generating pings and metrics,
+// remove this code.
+const samplePing = new Glean._private.PingType("sample", true, false);
```
Dexterp37, line 11: Can we move this to a file called `generatedPings.js`?

samples/web-extension/index.js
```diff
@@ -6,4 +6,46 @@
...
+  lifetime: "ping",
+  disabled: false
+}, "millisecond");
+const popupOpened = new Glean._private.CounterMetricType({
+  category: "sample",
+  name: "popup_opened",
+  sendInPings: ["sample"],
+  lifetime: "ping",
+  disabled: false
+}, "millisecond");
```
Dexterp37, line 25: Can we move these to a file called `generatedMetrics.js`?

samples/web-extension/index.js
```diff
@@ -6,4 +6,46 @@
...
+
+Glean.initialize("web-extension", true)
+  .then(() => {
+    console.log("Glean has been succesfully initialized.");
+    webExtStarted.set()
```
Dexterp37, line 30: Mh, let's note down in the dispatcher bug that we'll need to change this line as well. Please kindly also add a comment here about it: our examples should not recommend "bad practices". In our case the bad practice is waiting for glean to be initialized to perform the collection :-)

samples/web-extension/popup.js
```diff
@@ -4,4 +4,14 @@
 
 "use strict";
 
-console.info("Nothing will happen. Glean.js is not yet implemented.");
+const env = (typeof browser !== "undefined")
+  ? browser : (typeof chrome !== "undefined" ? chrome : null);
```
Dexterp37, line 8: Why do you need this? You're now using web-extension polyfill, so you can always just call `browser`.

src/glean.ts
```diff
@@ -228,18 +228,10 @@ class Glean {
   }
 
   static get applicationId(): string | undefined {
-    if (!Glean.instance._initialized) {
-      console.warn("Attempted to access the Glean.applicationId before Glean was initialized.");
-    }
-
```
Dexterp37, line 234: Why are you removing the init check here and below?

Dexterp37 (changes requested):

samples/web-extension/popup.js
Dexterp37: Ah, good point. If that is an issue just leave a comment before this line explaining why webextensions-polyfill is not usable here.

samples/web-extension/index.js
brizental: Hm, I wonder if initialize should even return a promise...

Dexterp37: Good thinking :) I believe it shouldn't, even though it's internally async, exactly to discourage depending on the initialization status. Can we tackle that as a separate PR?

Dexterp37 (approved):

samples/web-extension/index.js
brizental: Filed a bug https://bugzilla.mozilla.org/show_bug.cgi?id=1688223

PR #31, opened 2021-01-26
Bug 1687491 - Implement a dispatcher and add an instance to the Glean singleton

brizental:
This is part 1. 

Other PRs will come for actually using this dispatcher on the external APIs.

### DIFF
--- a/src/dispatcher.ts
+++ b/src/dispatcher.ts
@@ -112,8 +112,7 @@ class Dispatcher {
    *
    * # Note
    *
-   * Will not enqueue in case the dispatcher has not been initialized yet (i.e. `flushInit` has not been called yet)
-   * and the queues length will exceed `maxPreInitQueueSize`.
+   * Will not enqueue in case the dispatcher has not been initialized yet and the queues length exceeds `maxPreInitQueueSize`.
    *
    * @param task The task to enqueue.
    *
@@ -136,7 +135,7 @@ class Dispatcher {
   }
 
   /**
-   * Flushes the tasks enqueued while the dispatcher was uuninitialized.
+   * Flushes the tasks enqueued while the dispatcher was uninitialized.
    *
    * This is a no-op in case the dispatcher is not in an uninitialized state.
    */
--- a/src/dispatcher.ts
+++ b/src/dispatcher.ts
@@ -1,0 +1,175 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// The possible states a dispatcher instance can be in.
+const enum DispatcherState {
+  // The dispatcher has not been initialized yet.
+  //
+  // When the dispatcher is in this state it will not enqueue
+  // more than `maxPreInitQueueSize` tasks.
+  Uninitialized,
+  // There are no commands queued and the dispatcher is idle.
+  Idle,
+  // The dispatcher is currently processing queued tasks.
+  Processing,
+  // The dispatcher is blocked, tasks queued will not be immediatelly processed.
+  Blocked,
+}
+
+// The possible commands to be processed by the dispatcher.
+const enum Commands {
+  Task,
+  Block,
+}
+
+// A task the dispatcher knows how to execute.
+type Task = () => Promise<void>;
+
+// An executable command.
+type Command = {
+  task: Task,
+  command: Commands.Task
+} | {
+  command: Commands.Block
+};
+
+/**
+ * A task dispatcher for async tasks.
+ *
+ * Will make sure tasks are execute in order.
+ */
+class Dispatcher {
+  // A FIFO queue of tasks to execute.
+  private queue: Command[];
+  // The current state of this dispatcher.
+  private state: DispatcherState;
+  // A promise contianing the current execution promise.
+  //
+  // This is `undefined` in case there is no ongoing execution of tasks.
+  private currentJob?: Promise<void>;
+
+  constructor(readonly maxPreInitQueueSize = 100) {
+    this.queue = [];
+    this.state = DispatcherState.Uninitialized;
+  }
+
+  /**
+   * Gets the oldest command added to the queue.
+   *
+   * @returns The oldest command or `undefined` if the queue is empty.
+   */
+  private getNextCommand(): Command | undefined {
+    return this.queue.shift();
+  }
+
+  /**
+   * Enqueues a Block command at the start of the queue.
+   */
+  private block(): void {
+    this.queue.unshift({ command: Commands.Block });
+  }
+
+  /**
+   * Executes all the commands in the queue, from oldest to newest.
+   *
+   * Stops on case a `Block` command is encountered.
+   */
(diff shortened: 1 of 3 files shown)


src/dispatcher.ts
```diff
@@ -0,0 +1,175 @@
...
+  // The dispatcher is currently processing queued tasks.
+  Processing,
+  // The dispatcher is blocked, tasks queued will not be immediatelly processed.
+  Blocked,
+}
+
+// The possible commands to be processed by the dispatcher.
+const enum Commands {
+  Task,
+  Block,
```
brizental, line 23: `Block` may have been a bad terminology choice. `Stop` might be better.

Also, I missed adding docs to each variant here...

src/dispatcher.ts
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  /**
+   * Launches a task on this dispatchers queue.
+   * Kickstarts the execution of the queue in case it is currently Idle.
+   *
+   * # Note
+   *
+   * Will not enqueue in case the dispatcher has not been initialized yet (i.e. `flushInit` has not been called yet)
+   * and the queues length will exceed `maxPreInitQueueSize`.
```
brizental, line 116:
```suggestion
   * Will not enqueue in case the dispatcher has not been initialized yet and the queues length exceeds `maxPreInitQueueSize`.
```

What a confusing little comment I wrote there.

src/glean.ts
```diff
@@ -304,6 +333,9 @@ class Glean {
     // Get back to an uninitialized state.
     Glean.instance._initialized = false;
 
+    // Clear the dispatcher queue.
+    await Glean.dispatcher.clear();
```
brizental, line 337: We should probably also clear the dispatcher `onUploadDisabled`, I need to check what is done on the bindings.

src/dispatcher.ts
```diff
@@ -0,0 +1,175 @@
...
+    this.queue = [];
+    this.state = DispatcherState.Idle;
+  }
+
+  /**
+   * Returns a promise that resolves once the current task execution in finished.
+   *
+   * @returns The promise.
+   */
+  async blockOnQueue(): Promise<void> {
```
brizental, line 170: Should probably rename this to `testBlockOnQueue`.

src/dispatcher.ts
```diff
@@ -0,0 +1,175 @@
...
+    this.queue.push({
+      task,
+      command: Commands.Task
+    });
+    this.triggerExecution();
+    return true;
+  }
+
+  /**
+   * Flushes the tasks enqueued while the dispatcher was uuninitialized.
```
brizental, line 139:
```suggestion
   * Flushes the tasks enqueued while the dispatcher was uninitialized.
```

src/glean.ts
```diff
@@ -227,14 +237,33 @@ class Glean {
...
+   * Gets this Glean's instance server endpoint.
+   *
+   * @returns The server endpoint or `undefined` in case Glean has not been initialized yet.
+   */
   static get serverEndpoint(): string | undefined {
     return Glean.instance._serverEndpoint;
   }
 
+  /**
+   * Gets this Gleans's instance dispatcher instance.
```
brizental, line 259:
```suggestion
   * Gets this Gleans's instance dispatcher.
```

src/glean.ts
```diff
@@ -27,6 +28,11 @@ class Glean {
   private _coreMetrics: CoreMetrics;
   // The ping uploader.
   private _pingUploader: PingUploader
+  // A task dispatcher to help order asynchronous external API calls.
+  //
+  // The singleton will manage the dispatcher, but it is up to each developer
+  // to use this dispatcher to launch _any_ user facing API call.
```
brizental, line 34:
```suggestion
    // A task dispatcher to execute asynchronous external API calls.
```

src/glean.ts
```diff
@@ -27,6 +28,11 @@ class Glean {
...
+  private _dispatcher: Dispatcher;
```
brizental, line 35:
```suggestion
  // A task dispatcher to help execute in order asynchronous external API calls.
  private _dispatcher: Dispatcher;
```

src/dispatcher.ts
Dexterp37: Yes, good idea.

src/dispatcher.ts
Dexterp37: Yes, let's rename and add docs, please :-)

src/dispatcher.ts
```diff
@@ -0,0 +1,174 @@
...
+        console.warn("Unable to enqueue task, pre init queue is full.");
+        return false;
+      }
+    }
+
+    this.queue.push({
+      task,
+      command: Commands.Task
+    });
+    this.triggerExecution();
```
Dexterp37, line 133: Can you add a comment before this line stating that while we know the `triggerExecution` is async, we don't want to block on it because of reasons XYZ. (please add that to the other call sites as well)

src/dispatcher.ts
```diff
@@ -0,0 +1,174 @@
...
+   *
+   * # Note
+   *
+   * Will not enqueue in case the dispatcher has not been initialized yet and the queues length exceeds `maxPreInitQueueSize`.
+   *
+   * @param task The task to enqueue.
+   *
+   * @returns Whether or not the task was succesfully enqueued.
+   */
+  launch(task: Task): boolean {
```
Dexterp37, line 121: The return value is not being used anywhere (AFAICT). Do we need it or can it just be `void`?

Dexterp37 (changes requested):

src/dispatcher.ts
brizental: Yeah you are right, at least for now we are not using it. I'll remove.

src/glean.ts
```diff
@@ -131,11 +140,8 @@ class Glean {
     // We need to briefly set upload_enabled to true here so that `set`
     // is not a no-op.
     //
-    // Note that we can't provide the same guarantees as glean-core here.
-    // If by any change another actor attempts to record a metric while
-    // we are setting the known client id and first run date, they will be allowed to.
-    //
-    // TODO: Bug 1687491 might resolve this issue.
+    // This is safe.
+    // Since the dispatcher is stopped, no external API calls will be executed.
```
brizental, line 144: Isn't this nice? The dispatcher actually fixed the issue. Now, it will only really be fixed after my next PR when the recording calls are indeed dispatched.

tests/dispatcher.spec.ts
```diff
@@ -0,0 +1,177 @@
...
+import sinon from "sinon";
+
+import Dispatcher from "dispatcher";
+
+const sandbox = sinon.createSandbox();
+
+/**
+ * A sample task which returns a promise that takes between 0 and 10 ms to resolve.
+ *
+ * @returns The promise.
```
Dexterp37, line 15: nit: well, this could use some more explanation :D 
```suggestion
 * @returns The promise resolved when the timeout expires.
```

Dexterp37 (approved):

tests/dispatcher.spec.ts
brizental: Heh, for sure.

PR #32, opened 2021-01-27
Bug 1687491 - Use the dispatcher on the boolean metric type

brizental:
Opening this as a draft as I would like some feedback before using the pattern chosen here on all the other metrics :)

Note to self: Interestingly enough the boolean metric type is _not_ used by Glean internally. Good candidate for a future metric type plugin test.

### DIFF
--- a/src/metrics/types/boolean.ts
+++ b/src/metrics/types/boolean.ts
@@ -14,33 +14,40 @@ export class BooleanMetric extends Metric<boolean, boolean> {
   validate(v: unknown): v is boolean {
     return isBoolean(v);
   }
+
   payload(): boolean {
     return this._inner;
   }
 }
 
 /**
+ *
  *  A boolean metric.
  *
  * Records a simple flag.
+ *
  */
-class BooleanMetricType extends MetricType {
+abstract class BooleanMetricType extends MetricType {
   constructor(meta: CommonMetricData) {
     super("boolean", meta);
   }
 
   /**
-   * Sets to the specified boolean value.
+   * Simulates setting this boolean metric
+   * and returns the resulting boolean metric that would be recorded.
    *
    * @param value the value to set.
+   *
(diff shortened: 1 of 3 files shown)


src/metrics/types/boolean.ts
```diff
@@ -14,33 +14,40 @@ export class BooleanMetric extends Metric<boolean, boolean> {
...
     return isBoolean(v);
   }
+
   payload(): boolean {
     return this._inner;
   }
 }
 
 /**
+ *
```
Dexterp37, line 24: nit: is this intentional?

src/metrics/types/boolean.ts
```diff
@@ -61,4 +68,38 @@ class BooleanMetricType extends MetricType {
   }
 }
 
-export default BooleanMetricType;
+/**
+ * A boolean metric instance to be used only by Glean.
+ *
+ * This implementation will allow callers to wait on the metrics recording
+ */
+export class BooleanMetricTypeInternal extends BooleanMetricType {
```
Dexterp37, line 76: Do we need this exclusively because of the testing?

brizental: Initially, yes. But after working on this, I realized it is nice that _only_ external API calls go to the dispatcher and the rest is all synchronous, makes it much easier to follow the code. I am not sure this is how glean-core / bindings work.

src/metrics/types/boolean.ts
brizental: Sheesh, nice catch. No it isn't.

src/metrics/types/boolean.ts
brizental: Note that the idea is that Glean internal metrics will all be the "internal" type. And the exported metrics that users will use will be the "external" type.

PR #34, opened 2021-01-28
Bug 1687491 - Provide a way to test dispatched recording functions

brizental:
A different approach for using and testing dispatched functions, taking into account what we discussed offline eariler today.

This is a draft because I want to get feeback on the approach before using it on the other metrics.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -11,7 +11,11 @@ interface Configuration {
   // The user visible version string fro the application running Glean.js.
   readonly appDisplayVersion?: string,
   // The server pings are sent to.
-  readonly serverEndpoint?: string
+  readonly serverEndpoint?: string,
+  // Whether or not Glean should be started in testing mode.
+  //
+  // This unlocks features that allows dispatcher tasks to be waited on.
+  readonly testing?: boolean
 }
 
 export default Configuration;
--- a/src/glean.ts
+++ b/src/glean.ts
@@ -39,6 +39,8 @@ class Glean {
   private _serverEndpoint?: string;
   // Whether or not to record metrics.
   private _uploadEnabled?: boolean;
+  // Whether or this Glean instance is in testing mode.
+  private _testing?: boolean;
 
   private constructor() {
     if (!isUndefined(Glean._instance)) {
@@ -185,6 +187,10 @@ class Glean {
       return;
     }
 
+    if (config?.testing) {
+      Glean.instance._testing = true;
+    }
+
     if (applicationId.length === 0) {
       throw new Error("Unable to initialize Glean, applicationId cannot be an empty string.");
     }
@@ -267,6 +273,15 @@ class Glean {
   }
 
   /**
+   * Whether or not Glean is currently in testing mode.
+   *
+   * @returns Whether or not Glean is currently in testing mode.
+   */
+  static get testing(): boolean {
+    return Glean.instance._testing || false;
+  }
+
+  /**
    * Determines whether upload is enabled.
    *
    * When upload is disabled, no data will be recorded.
@@ -346,7 +361,10 @@ class Glean {
     await Glean.pingsDatabase.clearAll();
 
     // Re-Initialize Glean.
-    await Glean.initialize(applicationId, uploadEnabled, config);
+    await Glean.initialize(applicationId, uploadEnabled, {
+      ...config,
+      testing: true,
+    });
   }
 }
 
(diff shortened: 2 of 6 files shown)


src/config.ts
```diff
@@ -11,7 +11,11 @@ interface Configuration {
   // The user visible version string fro the application running Glean.js.
   readonly appDisplayVersion?: string,
   // The server pings are sent to.
-  readonly serverEndpoint?: string
+  readonly serverEndpoint?: string,
+  // Whether or not Glean should be started in testing mode.
+  //
+  // This unlocks features that allows dispatcher tasks to be waited on.
+  readonly testing?: boolean
```
brizental, line 18: More robustness on the testing flag should come with [Bug 1682771](https://bugzilla.mozilla.org/show_bug.cgi?id=1682771).

src/metrics/index.ts
```diff
@@ -149,4 +157,38 @@ export abstract class MetricType implements CommonMetricData {
...
+
+  /**
+   * **Test-only API**
+   *
+   * Allows the caller to block until all recording tasks related to this metric are completed.
+   *
+   * @returns A promise that returns once all ongoing recording tasks are completed.
+   */
+  async testBlockOnRecordingTasks(): Promise<void> {
+    if (!Glean.testing) {
```
brizental, line 187: In my dreams this is extracted into a decorator, but those are still in experimental mode in Typescript 😭 . Anyways also something to worry about in [Bug 1682771](https://bugzilla.mozilla.org/show_bug.cgi?id=1682771).

src/metrics/index.ts
```diff
@@ -115,6 +116,11 @@ export abstract class MetricType implements CommonMetricData {
   readonly lifetime: Lifetime;
   readonly disabled: boolean;
 
+  // Private list of ongoing dispatched recording tasks.
+  //
+  // This array will only be filled in testing mode.
+  #recording: Promise<void>[];
```
brizental, line 122: BTW, the `#` is a way to have _actually_ private attributes in JS. There is [a stage 3 proposal](https://github.com/tc39/proposal-private-methods) for this to be part of the language someday, but also typescript will comple this is a way to make it really inaccesible see: https://www.typescriptlang.org/play?#code/MYGwhgzhAEAqCmEAu0DeAoa0DETFIC5pkAnASwDsBzdTaYAewtIFdgkGSAKASjTqxIAFmQgA6XPmgBeaACIEyOQG46AX3RqgA

Dexterp37: Oh neat. It makes it very unpleasant to find, I agree. :)

src/config.ts
Dexterp37, line 18:
```suggestion
  readonly testing?: boolean,
```

Does the ts compiler like the trailing comma? If so, we should rather use this, it makes diffs clearer.

src/glean.ts
```diff
@@ -39,6 +39,8 @@ class Glean {
   private _serverEndpoint?: string;
   // Whether or not to record metrics.
   private _uploadEnabled?: boolean;
+  // Whether or this Glean instance is in testing mode.
+  private _testing?: boolean;
```
Dexterp37, line 43: Why don't we store the full configuration object? looks like we'll need it all anyway

src/metrics/index.ts
```diff
@@ -149,4 +157,38 @@ export abstract class MetricType implements CommonMetricData {
...
+
+  /**
+   * Dispatches a recording task to be performed asynchronously.
+   *
+   * If Glean is currently in testing mode the tasks are launched in canary mode
+   * and it's possible to use `testWaitOnRecordingTasks` to wait on them being completed.
+   *
+   * @param task The recording task.
+   */
+  dispatchRecordingTask(task: Task): void {
```
Dexterp37, line 169: Why `Recording`? Will we use a different task/dispatching function for pings?

Dexterp37 (commented):
I like this approach. I left a few comments. I'll take a more in depth review once this gets out of draft mode :)

src/metrics/index.ts
brizental: Yes, this is a method on the MetricType class.

src/config.ts
brizental: Honestly need to read up on this, the ts compiler doesn't care. There times in the code I used `;` for this too. Need to check if there are differences and if not, decide on a standard and add a eslint rule if possible.

Dexterp37: Let's stick to the trailing comma, that's what is being used in other Mozilla projects, too.

brizental:
Closing in favour of #35. Not that I applied the review comments made here, there.

PR #35, opened 2021-02-02
Bug 1687491 - Use the dispatcher on all external API calls

brizental:


brizental:
I am closing this, as yet again I have found out that the approach here is not correct. Instead of putting this in draft mode and closing it at a later time, I will simply close it now and hope that the next PR I open under this bug contains a working solution. 🤞

PR #36, opened 2021-02-02
Bug 1688902 - Create the events metric type

Dexterp37:
See the individual commit messages descriptions for easier review.

This exclusively adds the metric type, it doesn't take care of adding the 'events' ping.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "An implementation of the Glean SDK, a modern cross-platform telemetry client, for Javascript environments.",
   "main": "./dist/glean.js",
   "scripts": {
-    "test": "npm run build:test-webext && npm run test:unit --",
+    "test": "npm run build:test-webext && npm run test:unit",
     "test:debug": "ts-mocha \"tests/**/*.spec.ts\" --paths -p ./tsconfig.json --recursive --inspect-brk",
     "test:unit": "ts-mocha \"tests/**/*.spec.ts\" --paths -p ./tsconfig.json  --recursive --timeout 0",
     "lint": "eslint . --ext .ts,.js,.json --max-warnings=0",
--- a/package.json
+++ b/package.json
@@ -4,8 +4,9 @@
   "description": "An implementation of the Glean SDK, a modern cross-platform telemetry client, for Javascript environments.",
   "main": "./dist/glean.js",
   "scripts": {
-    "test": "npm run build:test-webext && ts-mocha \"tests/**/*.spec.ts\" --paths -p ./tsconfig.json  --recursive --timeout 0",
+    "test": "npm run build:test-webext && npm run test:unit --",
     "test:debug": "ts-mocha \"tests/**/*.spec.ts\" --paths -p ./tsconfig.json --recursive --inspect-brk",
+    "test:unit": "ts-mocha \"tests/**/*.spec.ts\" --paths -p ./tsconfig.json  --recursive --timeout 0",
     "lint": "eslint . --ext .ts,.js,.json --max-warnings=0",
     "fix": "eslint . --ext .ts,.js,.json --fix",
     "build:webext": "webpack --config webpack.config.webext.js --mode production",
--- a/src/metrics/events_database.ts
+++ b/src/metrics/events_database.ts
@@ -1,0 +1,230 @@
+// /* This Source Code Form is subject to the terms of the Mozilla Public
+//  * License, v. 2.0. If a copy of the MPL was not distributed with this
+//  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { Store } from "storage";
+import WeakStore from "storage/weak";
+import { MetricType } from "metrics";
+import { isUndefined, JSONArray, JSONObject, JSONValue } from "utils";
+
+export interface Metrics {
+  [aMetricType: string]: {
+    [aMetricIdentifier: string]: JSONValue
+  }
+}
+
+// An helper type for the 'extra' map.
+export type ExtraMap = { [name: string]: string };
+
+// Represents the recorded data for a single event.
+export class RecordedEvent {
+  constructor(
+    category: string,
+    name: string,
+    timestamp: number,
+    extra?: ExtraMap,
+  ) {
+    this.category = category;
+    this.name = name;
+    this.timestamp = timestamp;
+    this.extra = extra;
+  }
+
+  static toJSONObject(e: RecordedEvent): JSONObject {
+    return {
+      "category": e.category,
+      "name": e.name,
+      "timestamp": e.timestamp,
+      "extra": e.extra,
+    };
+  }
+
+  static fromJSONObject(e: JSONObject): RecordedEvent {
+    return new RecordedEvent(
+      e["category"] as string,
+      e["name"] as string,
+      e["timestamp"] as number,
+      e["extra"] as ExtraMap | undefined
+    );
+  }
+
+  // The event's category.
+  //
+  // This is defined by users in the metrics file.
+  readonly category: string;
+  // The event's name.
+  //
+  // This is defined by users in the metrics file.
+  readonly name: string;
+  // The timestamp of when the event was recorded.
+  //
+  // This allows to order events.
+  readonly timestamp: number;
+  // A map of all extra data values.
+  //
+  // The set of allowed extra keys is defined by users in the metrics file.
+  readonly extra?: ExtraMap;
+}
+
+/**
+ * The events database is an abstraction layer on top of the underlying storage.
+ *
+ * Event data is saved to the database in the following format:
+ *
+ * {
+ *  "pingName": {
+ *    [
+ *      {
+ *        "timestamp": 0,
+ *        "category": "something",
+ *        "name": "other",
+ *        "extra": {...}
+ *      },
+ *      ...
+ *    ]
+ *  }
+ * }
+ *
+ * Events only support `Ping` lifetime.
+ */
+class EventsDatabase {
+  private eventsStore: Store;
+
+  constructor() {
+    this.eventsStore = new WeakStore("unused");
+  }
+
+  /**
+   * Records a given event.
+   *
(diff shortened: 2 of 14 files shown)


src/metrics/events_database.ts
```diff
@@ -0,0 +1,230 @@
...
+ *  }
+ * }
+ *
+ * Events only support `Ping` lifetime.
+ */
+class EventsDatabase {
+  private eventsStore: Store;
+
+  constructor() {
+    this.eventsStore = new WeakStore("unused");
```
Dexterp37, line 94: This should probably use the persistent storage :-P

src/metrics/events_database.ts
```diff
@@ -0,0 +1,230 @@
...
+    // Make all the events relative to the first one.
+    const firstTimestamp =
+      (sortedData[0] as unknown as RecordedEvent)["timestamp"];
+
+    return sortedData.map((e) => {
+      const objE = e as JSONObject;
+      const timestamp = (objE["timestamp"] as number) ?? 0;
+      objE["timestamp"] = timestamp - firstTimestamp;
+      return objE; 
+    });
```
Dexterp37, line 219: I wonder if there's any less ugly way to express this. Thoughts?

src/metrics/events_database.ts
```diff
@@ -0,0 +1,230 @@
...
+  ): Promise<RecordedEvent[] | undefined> {
+    const value = await this.eventsStore.get([ping]);
+    if (!value) {
+      return undefined;
+    }
+
+    const rawEvents = value as JSONArray;
+    return rawEvents
+      // Only report events for the requested metric.
+      .filter((e) => {
```
brizental, line 142: nit: You could spare an iteration round here by using `reduce`.

```ts
return rawEvents.reduce((result, elem) => {
  const rawEventObj = elem as JSONObject;
  if (rawEventObj["category"] === metric.category && rawEventObj["name"] === metric.name) {
    result.push(RecordedEvent.fromJSONObject(e as JSONObject))
  }
}, []);
```

src/metrics/events_database.ts
```diff
@@ -0,0 +1,230 @@
...
+  /**
+   * Gets all of the persisted metrics related to a given ping.
+   *
+   * @param ping The name of the ping to retrieve.
+   * @param clearPingLifetimeData Whether or not to clear the ping lifetime metrics retrieved.
+   *
+   * @returns An object containing all the metrics recorded to the given ping,
+   *          `undefined` in case the ping doesn't contain any recorded metrics.
+   */
+  async getPingMetrics(ping: string, clearPingLifetimeData: boolean): Promise<JSONArray | undefined> {
```
brizental, line 192: nit: `getPingEvents`

src/metrics/events_database.ts
```diff
@@ -0,0 +1,230 @@
...
+      await this.eventsStore.delete([ping]);
+    }
+
+    if (pingData.length === 0) {
+      return;
+    }
+
+    // Sort the events by their timestamp.
+    const sortedData = pingData.sort((a, b) => {
+      const objA = a as unknown as RecordedEvent;
```
brizental, line 205: Why do you need the `as unknown` here? Ideally `getAndValidatePingData` should already return an array of `RecordedEvent`s in my mind. Then you wouldn't need so many castings here.

src/metrics/events_database.ts
brizental: What exactly are you finding ugly here? Heh. 

I guess for me the liberal use of `as` is concerning, but it also make it a bit "ugly", specially here where you have `as unknown as RecordedEvent`.

In glean-core we have a [`serialize_relative`](https://github.com/mozilla/glean/blob/main/glean-core/src/event_database/mod.rs#L49-L56) method in the RecordedEvent directly, that might separate things a bit and make it nicer.

Let me know if you meant that something else was ugly hehe.

src/metrics/types/event.ts
```diff
@@ -0,0 +1,100 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { MetricType, CommonMetricData } from "metrics";
+import Glean from "glean";
+import { ExtraMap, RecordedEvent } from "metrics/events_database";
+
+const MAX_LENGTH_EXTRA_KEY_VALUE = 100;
+
+/**
+ * A datetime metric.
```
brizental, line 12: Oops, 🍝

src/metrics/types/event.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  /**
+   * An helper function to aid mocking the time in tests.
+   *
+   * This is only meant to be overridden in tests.
+   * 
+   * @returns the number of milliseconds since the time origin.
+   */
+  protected getMonotonicNow(): number {
+    return Math.round(performance.now() / 1000);
```
brizental, line 33: `performance.now` is a browser API. Would you mind providing a fallback here? No need to overthink it, it can be a Date.now() and we add a nove about the fallback not being monotonic for now. This is probably why the Qt build failed.

src/metrics/types/event.ts
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  /**
+   * Record an event by using the information
+   * provided by the instance of this class.
+   *
+   * @param extra optional. This is a map, both keys and values need to be
+   *              strings, keys are identifiers. This is used for events where
```
brizental, line 41: uber nit: identation shoud align with `extra`. At least that is how I've been aligning all this time.

brizental (changes requested):
Approach looks fine for me overall. Left a bunch of comments but most of them are nits. 

My main concern is the liebral usage of `as` to bypass the type checker.

src/metrics/events_database.ts
Dexterp37: Yeah, the usage of `as unknown` was my concern :(

package.json
```diff
@@ -4,8 +4,9 @@
   "description": "An implementation of the Glean SDK, a modern cross-platform telemetry client, for Javascript environments.",
   "main": "./dist/glean.js",
   "scripts": {
-    "test": "npm run build:test-webext && ts-mocha \"tests/**/*.spec.ts\" --paths -p ./tsconfig.json  --recursive --timeout 0",
+    "test": "npm run build:test-webext && npm run test:unit --",
```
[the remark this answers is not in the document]
Dexterp37: Good catch, it can be removed.

src/metrics/events_database.ts
```diff
@@ -0,0 +1,230 @@
...
+   *
+   * This doesn't clear the stored value.
+   *
+   * @param ping the ping from which we want to retrieve this metrics value from.
+   * @param metric the metric we're looking for.
+   * 
+   * @returns an array of `RecordedEvent` containing the found events or `undefined`
+   *          if no recorded event was found.
+   */
+  async testGetValue(
```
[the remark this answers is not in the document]
Dexterp37: Sounds good. I changed this to `getEvents` (trailing 's') though, because it can return multiple events for the same id.

src/metrics/events_database.ts
```diff
@@ -0,0 +1,230 @@
...
+    timestamp: number,
+    extra?: ExtraMap,
+  ) {
+    this.category = category;
+    this.name = name;
+    this.timestamp = timestamp;
+    this.extra = extra;
+  }
+
+  static toJSONObject(e: RecordedEvent): JSONObject {
```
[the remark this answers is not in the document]
Dexterp37: In an attempt to have an easy optimization: we don't need them on each instance.

src/metrics/events_database.ts
Dexterp37: > Warning: untested code.

```
Type 'JSONValue' is not assignable to type 'RecordedEvent[] | undefined'.
  Type 'string' is not assignable to type 'RecordedEvent[] | undefined'.ts(2322)
```

:-D

src/metrics/types/event.ts
Dexterp37: @brizental we should probably have a "platform" level abstraction somwhere. WDYT?

src/metrics/events_database.ts
brizental: If the validation function returns the correct type that should make things nicer.

src/metrics/events_database.ts
brizental: https://github.com/mozilla/glean.js/pull/36#discussion_r569388605 This should fix that though.

src/metrics/types/event.ts
brizental: Honestly, for now I'd rather leave this here, if this function gets bigger we can break it into smaller parts. WDYT?

src/metrics/events_database.ts
brizental: Hey, I think this one was missed :)

src/metrics/events_database.ts
```diff
@@ -0,0 +1,209 @@
...
+   * @returns an array of `RecordedEvent` containing the found events or `undefined`
+   *          if no recorded event was found.
+   */
+  async getEvents(
+    ping: string,
+    metric: EventMetricType
+  ): Promise<RecordedEvent[] | undefined> {
+    const events = await this.getAndValidatePingData(ping);
+    if (events.length === 0) {
+      return undefined;
```
brizental, line 126: uber nit: If you change the return type to `Promise<RecordedEvent[] | void>` you can do just `return;` here.

brizental (changes requested):

src/metrics/events_database.ts
```diff
@@ -91,7 +81,7 @@ class EventsDatabase {
   private eventsStore: Store;
 
   constructor() {
-    this.eventsStore = new WeakStore("unused");
+    this.eventsStore = new PersistentStore("unused");
```
brizental, line 84: `"unused"` may not the most descriptive storage name. Maybe `"events"`?

brizental (approved):
r+wc!

PR #37, opened 2021-02-03
Bug 1690496 - Make init works _exactly_ like in glean-core

brizental:


src/glean.ts
```diff
@@ -189,17 +200,43 @@ class Glean {
...
+      if (clientId) {
+        if (clientId !== KNOWN_CLIENT_ID) {
+          // Temporarily enable uploading so we can submit a
+          // deletion request ping.
+          Glean.uploadEnabled = true;
+          await Glean.onUploadDisabled();
+        }
+      } else {
+        await Glean.clearMetrics();
+      }
```
Dexterp37, line 239: Should this logic be refactored out?

Dexterp37 (approved):
r+wc

src/glean.ts
brizental: Honestly, I'd rather leave it here for now. If init starts getting too big I can break it off. Would that be alright?

Dexterp37: Works for me

PR #40, opened 2021-02-05
Bug 1687491 - Dispatch all external API calls

brizental:
Not only does this change the code to dispatch all external API calls but also it adequates the dispatcher and other parts of the code for all to work as expected. 

I attempted to separate the changes as well as possible in specific commits and be very descriptive in my commit messages. Sincerely hope that makes review clearer.

It seems this could have been broken in multiple PRs, but either tests would fail or the change wouldn't make sense by itself.


src/dispatcher.ts
```diff
@@ -276,6 +347,72 @@ class Dispatcher {
...
+   * This is important in order not to hang forever in case the dispatcher is stopped.
+   *
+   * @param task The task to launch.
+   *
+   * @returns A promise which only resolves once the task is done being executed
+   *          or is guaranteed to not be executed ever i.e. if the queue gets cleared.
+   *          This promise is rejected if the dispatcher takes more than 1s
+   *          to execute the current task i.e. if the dispatcher is uninitialized.
+   */
+  testLaunch(task: Task): Promise<void> {
```
brizental, line 368: The best way that I found to make sure that `testGetValue` function were called for sure only after `set`/`add` function was to dispatch the testing functions themselves.

samples/qt-qml-app/main.qml
```diff
@@ -48,10 +48,6 @@ Rectangle {
 
   Component.onCompleted: {
     // Initialize Glean when the application starts.
-    Glean.Glean.initialize("qt-qml-app", true)
-      // Note: If you change this console messages you should change
-      // the console message verifications on `bin/qt-js-check.sh.
-      .then(() => console.log("Called Glean.initialize succesfully."))
-      .catch(err => console.error(`Some Javascript error occured.\n${err}`));
+    Glean.Glean.initialize("qt-qml-app", true);
```
Dexterp37, line 51: `Glean.Glean` is so... eugh. But not new in this PR :)

src/glean.ts
```diff
@@ -190,11 +188,11 @@ class Glean {
    * - If config.serverEndpoint is an invalid URL;
    * - If the application if is an empty string.
    */
-  static async initialize(
+  static initialize(
```
Dexterp37, line 191: <3

src/glean.ts
```diff
@@ -203,54 +201,66 @@ class Glean {
     if (applicationId.length === 0) {
       throw new Error("Unable to initialize Glean, applicationId cannot be an empty string.");
     }
-    Glean.instance._applicationId = sanitizeApplicationId(applicationId);
-    Glean.instance._config = new Configuration(config);
 
-    // Clear application lifetime metrics.
+    // The configuration constructor will throw in case config hany any incorrect prop.
```
Dexterp37, line 205:
```suggestion
    // The configuration constructor will throw in case config has any incorrect prop.
```

Dexterp37: But.. it should probably not throw. The Glean SDK never throws, since it may break the consuming product.

src/metrics/types/datetime.ts
```diff
@@ -162,20 +162,27 @@ export class DatetimeMetric extends Metric<DatetimeInternalRepresentation, strin
...
+   * This is absolutely not meant to be used outside of Glean itself.
+   * It may cause multiple issues because it cannot guarantee
+   * that the recording of the metric will happen in order with other Glean API calls.
+   *
+   * @param instance The metric instance to record to.
+   * @param value The date we want to set to.
    */
-  async set(value?: Date): Promise<void> {
-    if (!this.shouldRecord()) {
+  static async _private_setSync(instance: DatetimeMetricType, value?: Date): Promise<void> {
```
Dexterp37, line 184: You know, it feels weird for `_private_setSync` to be marked as `async` :-D

Dexterp37: Maybe `_private_setUndispatched`?

src/storage/persistent/webext.ts
```diff
@@ -146,7 +146,7 @@ class WebExtStore implements Store {
       );
       return this.store.set(query);
     } catch(e) {
-      console.warn(e.message, "Ignoring");
+      console.warn((e as Error).message, "Ignoring");
```
Dexterp37, line 149:
```suggestion
      console.warn("Ignoring key", e);
```

This is another option. The whole exception object will be accessible in the console.

src/storage/weak.ts
```diff
@@ -5,42 +5,52 @@
 import { Store, StorageIndex } from "storage";
 import { updateNestedObject, getValueFromNestedObject, deleteKeyFromNestedObject } from "storage/utils";
 import { JSONObject, JSONValue } from "utils";
-
+
```
Dexterp37, line 8: nit :) trailing whitespace

src/storage/weak.ts
```diff
@@ -5,42 +5,52 @@
...
       this.store = deleteKeyFromNestedObject(this.store, index);
     } catch (e) {
-      console.warn(e.message, "Ignoring.");
+      console.warn((e as Error).message, "Ignoring.");
     }
+    return Promise.resolve();
   }
 }
-
+
```
Dexterp37, line 55: nit :) trailing whitespace

Dexterp37 (changes requested):

samples/qt-qml-app/main.qml
brizental: Yup, not sure how to bypass that though. Also didn't look into it too much hehe.

src/storage/weak.ts
brizental, line 8:
```suggestion

```

src/storage/weak.ts
brizental, line 55:
```suggestion

```

src/dispatcher.ts
```diff
@@ -118,12 +185,19 @@ class Dispatcher {
...
+      this.currentJob
+        .then(() => {
+          this.currentJob = undefined;
+          if (this.state === DispatcherState.Processing) {
+            this.state = DispatcherState.Idle;
+          }
+        })
+        .catch(error => {
+          console.error(
+            "IMPOSSIBLE: Something went wrong while the dispatcher was executing the tasks queue.",
```
Dexterp37, line 197: Bets on when we'll see this message?

tests/upload/index.spec.ts
```diff
@@ -112,7 +112,7 @@ describe("PingUploader", function() {
 
     // Trigger uploading, but don't wait for it to finish,
     // so that it is ongoing when we cancel.
-    uploader.triggerUpload();
+    void uploader.triggerUpload();
```
Dexterp37, line 115: What's this `void` here for?

Dexterp37 (approved):
r+wc

src/dispatcher.ts
brizental: no bets, unless we change the `executeTask` to not catch errors or use the observers for something else. right now it should be impossible for this to error. i only added the catch here because the linter was complaining :D

tests/upload/index.spec.ts
brizental: ah, this is also a complaint from the linter, about not doing anything with a statement that returns a promise. but the `void` is pretty cool. it is not a typescript operator, it is a javacript operator and what it does is evalute whatever expression comes after it and return undefined.

PR #42, opened 2021-02-08
Bug 1686685 -  Implement the deletion request ping

brizental:
The approach here is a bit different than the approach in glean-core.

I ported all the tests I found from the core and the android bindings and they all passed. 

**What's different?**

I skipped having a different storage for deletion requests and instead treat them as any other ping. I also added a special case, so that the deletion-request ping is sent even if upload is disabled.


Dexterp37 (approved):
Nice one :)

PR #45, opened 2021-02-10
Bug 1691365 - Use glean_parser for generating metrics and pings code in samples

brizental:
This depends on https://github.com/mozilla/glean_parser/pull/274 being merged and new version of glean_parser being released after that. Before that happens CI should fail for Qt because we need to add a glean_parser step to that script.


samples/qt-qml-app/metrics.yaml
```diff
@@ -0,0 +1,48 @@
...
+# This file defines the metrics that are recorded by the Glean SDK.
+# APIs to use these pings are automatically generated at build time using
+# the `glean_parser` PyPI package.
+
+# Metrics in this file may make use of SDK reserved ping names. See
+# https://mozilla.github.io/glean/book/dev/core/internal/reserved-ping-names.html
+# for additional information.
+
+---
+$schema: moz://mozilla.org/schemas/glean/metrics/1-0-0
```
Dexterp37, line 14: Let's use the schema 2-0-0 :-)

samples/qt-qml-app/pings.yaml
```diff
@@ -0,0 +1,24 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# This file defines the pings that are recorded by the Glean SDK.
+# Their code APIs is automatically generated, at build time using,
+# the `glean_parser` PyPI package.
+
+---
+$schema: moz://mozilla.org/schemas/glean/pings/1-0-0
```
Dexterp37, line 10: Let's use the newer schema here, too!

samples/web-extension/README.md
```diff
@@ -16,14 +16,25 @@ Whenever this web extensions popup is opened it will trigger Glean.js events.
 npm run build:webext
 ```
 
-2. Build this sample. On this `web-extension` folder run:
+2. Generate metrics and pings files.
+
+```bash
+npm run glean_parser
```
Dexterp37, line 22: Mh, this is really nitpicking, but should we think of a better name, especially if we're planning on exposing commands to external products? `npm run glean` seems good enough, IMHO :)

For consideration, but not for this PR:

- Is there any way we could expose an NPM hook to run at specific points ? (e.g. always run when touching code?)
- How can we nicely integrate with webextension build systems?

samples/web-extension/metrics.yaml
```diff
@@ -0,0 +1,48 @@
...
+# This file defines the metrics that are recorded by the Glean SDK.
+# APIs to use these pings are automatically generated at build time using
+# the `glean_parser` PyPI package.
+
+# Metrics in this file may make use of SDK reserved ping names. See
+# https://mozilla.github.io/glean/book/dev/core/internal/reserved-ping-names.html
+# for additional information.
+
+---
+$schema: moz://mozilla.org/schemas/glean/metrics/1-0-0
```
Dexterp37, line 14: schema version :D

samples/web-extension/pings.yaml
```diff
@@ -0,0 +1,24 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# This file defines the pings that are recorded by the Glean SDK.
+# Their code APIs is automatically generated, at build time using,
+# the `glean_parser` PyPI package.
+
+---
+$schema: moz://mozilla.org/schemas/glean/pings/1-0-0
```
Dexterp37, line 10: Version :-P

src/internal_pings.ts
```diff
@@ -8,10 +8,8 @@ import PingType from "pings";
 /**
  * Glean-provided pings, all enabled by default.
  *
- * Pings initialized here should be defined in `pings.yaml`
- * and for now manually translated into JS code.
- *
- * TODO: This file should be auto-generated once Bug 1691365 is resolved.
+ * Pings initialized here should be defined in `./pings.yaml`
```
Dexterp37, line 11: What do you mean, here?

src/metrics.yaml
```diff
@@ -0,0 +1,277 @@
...
+# This file defines the metrics that are recorded by the Glean SDK.
+# APIs to use these pings are automatically generated at build time using
+# the `glean_parser` PyPI package.
+
+# Metrics in this file may make use of SDK reserved ping names. See
+# https://mozilla.github.io/glean/book/dev/core/internal/reserved-ping-names.html
+# for additional information.
+
+---
+$schema: moz://mozilla.org/schemas/glean/metrics/1-0-0
```
Dexterp37, line 14: Schema version !

src/pings.yaml
```diff
@@ -0,0 +1,27 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# This file defines the pings that are recorded by the Glean SDK.
+# Their code APIs is automatically generated, at build time using,
+# the `glean_parser` PyPI package.
+
+---
+$schema: moz://mozilla.org/schemas/glean/pings/1-0-0
```
Dexterp37, line 10: Guess what..? :-D

Dexterp37 (changes requested):

samples/qt-qml-app/metrics.yaml
brizental: Heh, this file is copied over from glean-core. Did you find a bug there? https://github.com/mozilla/glean/blob/9225273abbb7a3d91079145d202edcaf621bdf70/glean-core/pings.yaml#L10 Happy to file it if confirmed.

Dexterp37: It's not a bug. We're migrating things over. The glean parser version with the new schema was released [only 5 days ago](https://github.com/mozilla/glean_parser/blob/main/HISTORY.rst#200-2021-02-05).

samples/web-extension/README.md
brizental: > Is there any way we could expose an NPM hook to run at specific points ? (e.g. always run when touching code?)

We can run the script whenever we run `npm run dev` or `npm run build`. I thought that could get annoying because these files don't really need to be rebuilt everytime and I thought it was overkill to write some sort of script that checks if there were changes to the .yaml files before building, given that this is a sample app.

> How can we nicely integrate with webextension build systems?

I [filed a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1692149) for investigation on this. Like I put in the note here, right now you need to have glean_parser installed as an available CLI tool for the command to work. My hunch is that the easiest path would be to provide a glean_parser npm package that will install glean_parser (the python package) magically for the user.

src/internal_pings.ts
brizental: That the pings defined here should be defined in pings.yaml. Not sure what is the confusion.

Autogenerating the internal metrics and pings is not a good idea here, we need a more robust and specific class than glean_parser provides (at least for metrics) and we would need to generate these files all the time... anyways would not play well with the development workflow IMO. I think glean_parser should stay as a user tool, not a glean internal development tool.

brizental: > That the pings defined here should be defined in pings.yaml

We have this for glean_core, that is the only reason I also added .yaml files for internal metrics and pings here. There are really just for show, we should instantite manually for internal stuff.

samples/qt-qml-app/metrics.yaml
brizental: TIL!

src/internal_pings.ts
Dexterp37: > 
> 
> That the pings defined here should be defined in pings.yaml. Not sure what is the confusion.

Well, the sentence was lacking the context you're providing below :)

> Autogenerating the internal metrics and pings is not a good idea here, we need a more robust and specific class than glean_parser provides (at least for metrics) and we would need to generate these files all the time... anyways would not play well with the development workflow IMO. I think glean_parser should stay as a user tool, not a glean internal development tool.

I'm not sure what you mean here. Why can't we dogfood?

brizental: > I'm not sure what you mean here. Why can't we dogfood?

I tried it just now. The tl;dr; is: we can, but it is not so simple.

1. The `glean_parser` template I created was thought out to be used by a consumer of Glean
and not by Glean itself. Looking at an example:

```js
/* eslint-disable */

/* This Source Code Form is subject to the terms of the Mozilla Public

"use strict";

// AUTOGENERATED BY glean_parser.  DO NOT EDIT.

// Global variable to use when in Qt/QML environments,
// where do not have access to the global object (i.e. window);
var gleanInternalMetrics;

// Universal Module Definition (UMD) template based on:
// https://github.com/umdjs/umd/blob/master/templates/returnExports.js
(function (root, factory) {
    if (typeof define === "function" && define.amd) {
        // AMD. Register as an anonymous module.
        define(["Glean"], factory);
    } else if (typeof module === "object" && module.exports) {
        // Node. Does not work with strict CommonJS, but
        // only CommonJS-like environments that support module.exports, like Node.
        module.exports = factory(require("glean"));
    } else if (typeof root === "undefined") {
        // In Qt/QML environments we can't change the global object from Javascript.
        // We will simply assing to a global variable in this case.
        gleanInternalMetrics = factory(Glean.Glean);
    } else {
        // Browser globals (root is window)
        if (!root["gleanAssets"]) {
            root["gleanAssets"] = {};
        }
        root["gleanAssets"]["gleanInternalMetrics"] = factory(root.Glean);
    }
})(typeof self !== "undefined" ? self : this, function(Glean) {
    return {
        /**
        * The name of the operating system.
        * Possible values:
        * Android, iOS, Linux, Darwin, Windows,
        * FreeBSD, NetBSD, OpenBSD, Solaris, unknown
        *
        * Generated from `os`.
        */
        os: new Glean._private.StringMetricType({
            category: "",
            name: "os",
            sendInPings: ["glean-client-info"],
            lifetime: "application",
            disabled: false,
        }, ),
    };
});
```

This is the branch we would fall on inside Glean.js:

```js
else if (typeof module === "object" && module.exports) {
    // Node. Does not work with strict CommonJS, but
    // only CommonJS-like environments that support module.exports, like Node.
    module.exports = factory(require("glean"));
}
```

Note that is assumes Glean is an installed package, named "glean". When we are inside Glean.js,
this assumption fails and this `require` statement returns `undefined` which will break everything.

2. The glean_parser template is a JS template, which means when we want to import it in Typescript,
Typescript is not too happy that it doesn't have type definitions. I have opened [a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1692157) about that.

3. We would need to run the glean_parser command before all other commands, it is fast,
but even so it could get annoying. We could maybe write a script that checked if it is necessary
to run that command before doing that, but that is better as a follow up IMO (see also: https://github.com/mozilla/glean.js/pull/45#discussion_r574321319).

Dexterp37: Ok, let's file a bug about this and move on.

Dexterp37 (approved):

PR #47, opened 2021-02-11
Bug 1692002 - Enable dependabot JS updates

brizental:


.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+version: 2
+updates:
```
Dexterp37, line 2: We should probably also track the samples: https://github.com/mozilla/glean.js/tree/main/samples/web-extension

Dexterp37 (approved):

PR #72, opened 2021-02-16
Bug 1686678 - Reorganize code in core/ and platform/ submodules

brizental:
This is more or less related to Bug 1686678. I did a bit of reserach on that bug to figure out the best way to organize the Glean.js package. In this research I saw other packages that organize their code like I do in this PR and I liked the approach.

I think this way the platform specific and non platform specific code is more clearly separated. As a bonus it is clearer to separate tests that run on the browser and tests that don't.

I also moved all the code into a glean/ folder. That is necessary in order to be able to use `npm link` to link and use the future `@mozilla/glean` package in our samples. Turns out we can't use that command if the sample is inside the package folder. To bypass that I move the package into a dedicated folder.


glean/src/core/metrics/database.ts
```diff
@@ -148,7 +148,7 @@ class MetricsDatabase {
    *
    * This behaviour is not consistent with what the Glean SDK does, but this is on purpose.
    * On the Glean SDK we panic when we can't serialize the persisted value,
-   * that is because this is an extremely unlikely situation for that environment.
+   * that is because this is an extremely unlikely situation for that platformironment.
```
Dexterp37, line 151: `platformironment.`? :-D

Dexterp37 (approved):
r+wc

glean/src/core/metrics/database.ts
brizental: That is what you get when you search and replave `env` to `platform` in a million places at the same time. Good catch though, thanks!

PR #74, opened 2021-02-17
Bug 1693260 - Add documentation about making a new release

brizental:


docs/release.md
```diff
@@ -0,0 +1,35 @@
+# Making a new release
```
Dexterp37, line 1: I think the SDK model works quite well for cutting a new release and it was ironed out through out the last year. We've been quite successful in applying the same model to Rally. Do you think there's value in doing the same for glean.js?

bin/prepare-release.sh
```diff
@@ -0,0 +1,133 @@
...
+    -e "s/\.\.\.main/...v${NEW_VERSION}/" \
+    "${WORKSPACE_ROOT}/${FILE}"
+run rm "${WORKSPACE_ROOT}/${FILE}.bak"
+
+if [ "$DOIT" = y ]; then
+    CHANGELOG=$(cat "${WORKSPACE_ROOT}/${FILE}")
+    cat > "${WORKSPACE_ROOT}/${FILE}" <<EOL
+# Unreleased changes
+
+[Full changelog](https://github.com/mozilla/glean/compare/v${NEW_VERSION}...main)
```
Dexterp37, line 101:
```suggestion
[Full changelog](https://github.com/mozilla/glean.js/compare/v${NEW_VERSION}...main)
```

Dexterp37 (approved):
r+wc

bin/prepare-release.sh
brizental: Argh, I thought I got all of these. Thanks for catching it :D

PR #86, opened 2021-02-23
Bumped version to 0.2.0

brizental:
Opening this as a pull request just so that CI runs.

I have already updated CircleCI to run not only on pull requests, but apparently it will still not let me re-run this job...


brizental:
And CI still did not run...

PR #90, opened 2021-02-24
Fix errors when importing Glean.js in Typescript environments

brizental:
Currently when attempting to import the @mozilla/glean package in the Bergamot extension, we get the following error: 

```bash
Cannot find module '@mozilla/glean/webext' or its corresponding type declarations.
```

I am mostly sure that the problem is with our types definitions. If you look at [this thread ](https://github.com/microsoft/TypeScript/issues/33079), you'll notice that it is not ironed out exactly how [`exports`](https://nodejs.org/api/packages.html#packages_subpath_exports) and type declarations play together.

I have used the `typesVersions` workaround since the change to use `exports`, because I am basing myself off of how [`jose`](https://github.com/panva/jose/blob/main/package.json) does thigs.

I seem to be using it incorrectly. In this branch I made some experiements. You'll see that I included conditional imports for different environment, after [noticing](https://github.com/mozilla-extensions/bergamot-browser-extension/blob/main/extension/tsconfig.json#L10) that bergamot is expecting `commonjs` modules. That was not the issue, but since I've already done the work I decided to keep it (we had [a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1693039) for it anyways).

Playing around with the `typesVersions` field was more promising. The way that is is right now I finally don't get an error when importing Glean in in the Bergamot extension 🎉 

Screenshot 2021-02-24 at 12 25 55

I also don't get an error when running `yarn run build:production`. Yay!

Sadly the minute I use the imported `Glean` object, either through `Glean.initialize` or even `console.log(Glean)` I get back to the same `Cannot find module...`. 

Also this only solves the `/webext` import, when attempting to do something like `/webext/private/ping` for example, we get back to not being able to import.

### DIFF
--- a/glean/tsconfig/base.json
+++ b/glean/tsconfig/base.json
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "target": "ES2019",
-    "module": "commonjs",
+    "module": "CommonJS",
     "strict": true,
     "moduleResolution": "node",
     "esModuleInterop": true,
--- a/glean/tsconfig/base.json
+++ b/glean/tsconfig/base.json
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "target": "ES2019",
-    "module": "CommonJS",
+    "module": "commonjs",
     "strict": true,
     "moduleResolution": "node",
     "esModuleInterop": true,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 [Full changelog](https://github.com/mozilla/glean.js/compare/v0.2.0...main)
 
+* [#90](https://github.com/mozilla/glean.js/pull/90): BUGIFX: Accept lifetimes as strings when instantiating metric types.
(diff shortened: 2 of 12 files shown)


brizental:
Thought: Seems like the import cares about the path for the types definitions (notice that in the vscode screenshot I provided the module path points to its type defiitions). Maybe, it error only when I actually call Glean, but the actual code is only imported upon first call, up until then it only cares about types definitions.

I have no basis for this claim, just assumptions based on these experiences. Could be due to the usage of yarn maybe? Remember the import errors do not happen in our own sample web extension.

glean/tsconfig/base.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "target": "ES2019",
-    "module": "commonjs",
+    "module": "CommonJS",
```
Dexterp37, line 4: Is this really needed?

Dexterp37 (approved):
r+wc

glean/tsconfig/base.json
brizental: Yeah, nope. 
```suggestion
    "module": "commonjs",
```

PR #93, opened 2021-03-03
Fix the indentation in glean.ts from 3 to 2 spaces

Dexterp37:
This only changes the indentation. I think this was mistakenly changed, since it's only part of the file.


brizental (approved):
r+! Provided the lint issues in CI get fixed :D

PR #94, opened 2021-03-03
Bug 1689547 - Create plugin / events infrastructure and define `afterPingCollection` event

brizental:


glean/src/core/config.ts
```diff
@@ -25,6 +27,8 @@ export interface ConfigurationInterface {
   readonly serverEndpoint?: string,
   // Debug configuration.
   debug?: DebugOptions,
+  // Optional list of plugins to instrument the current Glean instance.
```
Dexterp37, line 30:
```suggestion
  // Optional list of plugins to enable in the current Glean instance.
```

glean/src/core/config.ts
```diff
@@ -3,6 +3,8 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import { DEFAULT_TELEMETRY_ENDPOINT } from "./constants";
+import GleanPlugin from "../plugins";
+import GleanEvents from "./events";
```
Dexterp37, line 7: Why do you need this?

glean/src/core/events.ts
```diff
@@ -0,0 +1,15 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { PingPayload } from "./pings/database";
+import { JSONObject } from "./utils";
+
+import { GleanEvent } from "../plugins/index";
+
+export const AfterPingCollection: GleanEvent<[PingPayload], Promise<JSONObject>> = {
```
Dexterp37, line 10: Can you document what's this? When is this called? What's it for?

glean/src/core/glean.ts
```diff
@@ -94,6 +98,30 @@ class Glean {
...
+  /**
+   * Registers a list of plugins.
+   *
+   * @param plugins The plugins list from the configuration.
+   */
+  private static registerPlugins(plugins: GleanPlugin<GleanEvents>[]): void {
+    for (const plugin of plugins) {
+      if (plugin.event in Glean.plugins) {
+        console.error(
+          `Attempted to register plugin '${plugin.name}', which instruments the event '${plugin.event}'.`,
```
Dexterp37, line 114: Mh, what do you mean by "instrument"? We usually use this term to mean "add something to measure".

glean/src/core/pings/maker.ts
```diff
@@ -206,20 +208,32 @@ function makePath(identifier: string, ping: PingType): string {
...
+  if (!collectedPayload) {
     return;
   }
 
   if (Glean.logPings) {
-    console.info(JSON.stringify(payload, null, 2));
+    console.info(JSON.stringify(collectedPayload, null, 2));
+  }
+
+  const modifiedPayload = await Glean.triggerEvent(AfterPingCollection, collectedPayload);
```
Dexterp37, line 231: We should consider (not in this PR) a different way to inject the dependency here, reducing the calls to the Glean object.

glean/src/core/pings/maker.ts
```diff
@@ -206,20 +208,32 @@ function makePath(identifier: string, ping: PingType): string {
...
+  let payload: JSONObject;
+  if (modifiedPayload) {
+    payload = modifiedPayload;
+  } else {
+    payload = collectedPayload;
```
Dexterp37, line 236:
```suggestion
  const modifiedPayload = await Glean.triggerEvent(AfterPingCollection, collectedPayload) ?? collectedPayload;
```

What about this?

glean/tests/core/pings/maker.spec.ts
```diff
@@ -100,4 +111,38 @@ describe("PingMaker", function() {
...
+        super(AfterPingCollection.name, "mockPlugin");
+      }
+
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      action(_payload: PingPayload): Promise<JSONObject> {
+        return Promise.resolve({ "you": "got mocked!" });
+      }
+    }
+
+    await Glean.testUninitialize();
```
Dexterp37, line 130: I wish we could just test  this in isolation. We might want to reconsider how we propagate our dependencies within Glean.js :(

Dexterp37 (changes requested):

glean/src/core/pings/maker.ts
brizental: Much better.

glean/src/core/glean.ts
brizental: Hum, maybe I am using this word wrong... _checks wikipedia_ ... 100% I am using it wrong 🆘 Maybe I want "listens to"?

glean/src/core/events.ts
brizental: So, this is here to define:

- The type of the context provided by this event (`[PingPayload]`);
- The type of the result expected from this event (`[Promise<JSONObject>]`);
- The name of this event.

This way, when we define a plugin like this `class MyPlugin extends GleanPlugin<typeof AfterPingCollection>` the `action` in this plugin will already expected the correct arguments and return type defined by the event.

glean/src/core/pings/maker.ts
brizental: I considered making events completely unaware of the Glean singleton, but didn't try that out. I do think if we are going to go that way it should be now, otherwise we will be a bit stuck.

Let me attempt that and open an alternate to this PR with the different approach. I did not go that way before, because it would be quite different from what I defined in the proposal for this, but there is no harm in testing.

glean/src/core/config.ts
brizental: They are types, we pass plugins through the `ConfigurationInterface` to `Glean.initialize`, and this is here to help define exactly _what_ we are passing through that property.

brizental: `GleanEvents` is a union of all the possible GleanEvents types (not a union right now since we only defined one event, but you get the gist). And `GleanPlugin` is the type of the accepted plugin class. Each plugin is always attached to a type of event. To ensure we only accept plugins that listen to valid Glean events, the final type for the `plugins` property is `GleanPlugin<GleanEvents>[]`.

glean/src/core/glean.ts
brizental: Thankfully I can always claim that English is not my first language 🥇

brizental:
Closing in favour of #95

PR #95, opened 2021-03-04
[ALTERNATE] Bug 1689547 -  Create plugin / events infrastructure and define `afterPingCollection` event

brizental:
Different approach for #94 (and my preferred approach, 1) Less cleverness on the types 2) Less relying on the Glean singleton).

Based off of [this thread](https://github.com/mozilla/glean.js/pull/94#discussion_r586471152).

### DIFF
--- a/glean/src/core/events/index.ts
+++ b/glean/src/core/events/index.ts
@@ -1,0 +1,79 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import CoreEvents from "../../../dist/webext/types/core/events";
+import Plugin from "../../plugins";
+
+import { PingPayload } from "../pings/database";
+import { JSONObject } from "../utils";
+
+export class CoreEvent<
+   // An array of arguments that the event will provide as context to the plugin action.
+   Context extends unknown[] = unknown[],
+   // The expected type of the action result. To be returned by the plugin.
+   Result extends unknown = unknown
+ > {
+   // The plugin to be triggered eveytime this even occurs.
+   private plugin?: Plugin<CoreEvent<Context, Result>>;
+
+   constructor(readonly name: string) {}
+
+   /**
+    * Registers a plugin that listens to this event.
+    *
+    * @param plugin The plugin to register.
+    */
+   registerPlugin(plugin: Plugin<CoreEvent<Context, Result>>): void {
+     if (this.plugin) {
+       console.error(
+         `Attempted to register plugin '${plugin.name}', which listens to the event '${plugin.event}'.`,
+         `That event is already watched by plugin '${this.plugin.name}'`,
+         `Plugin '${plugin.name}' will be ignored.`
+       );
+       return;
+     }
+ 
+     this.plugin = plugin;
+   }
+ 
+   /**
+    * Deregisters the currently registered plugin.
+    *
+    * If no plugin is currently registered this is a no-op.
+    */
+   deregisterPlugin(): void {
+     this.plugin = undefined;
+   }
+
+   /**
+    * Triggers this event.
+    *
+    * Will execute the action of the registered plugin, if there is any.
+    *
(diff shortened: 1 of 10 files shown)


glean/src/core/events/index.ts
```diff
@@ -0,0 +1,79 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import CoreEvents from "../../../dist/webext/types/core/events";
+import Plugin from "../../plugins";
+
+import { PingPayload } from "../pings/database";
+import { JSONObject } from "../utils";
+
+export class CoreEvent<
```
brizental, line 11: It's not ideal to call this `Event` because that clashes with the lib DOM `Event` type. As you may remember from my other PR, I was calling this `GleanEvent`, but I thought it was weird, obviously it is a Glean event this is Glean! `CoreEvent` goes along with the `CorePings` and `CoreMetrics` naming choices.

glean/src/core/glean.ts
```diff
@@ -231,6 +232,12 @@ class Glean {
     // The configuration constructor will throw in case config has any incorrect prop.
     const correctConfig = new Configuration(config);
 
+    if (config?.plugins) {
+      for (const plugin of config.plugins) {
+        registerPluginToEvent(plugin);
+      }
+    }
```
brizental, line 239: The only responsability of the singleton is to register these plugins and reset them when in testing mode.

glean/src/core/pings/maker.ts
```diff
@@ -206,27 +208,35 @@ function makePath(identifier: string, ping: PingType): string {
...
 
   if (Glean.logPings) {
-    console.info(JSON.stringify(payload, null, 2));
+    console.info(JSON.stringify(collectedPayload, null, 2));
   }
 
   const headers = getPingHeaders();
+
+  const modifiedPayload = await CoreEvents.afterPingCollection.trigger(collectedPayload);
+  const finalPayload = modifiedPayload ? modifiedPayload : collectedPayload;
```
brizental, line 234: Need to check exactly why, by `modifiedPayload ?? collectedPayload` did not work, while this worked 🤔

Dexterp37: Mh, maybe because of the `null` vs `undefined` difference? see [the docs](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing)

glean/src/core/events/index.ts
Dexterp37: Yes, this seems like a good rationale and a good choice.

glean/src/core/glean.ts
```diff
@@ -50,7 +51,7 @@ class Glean {
     metrics: MetricsDatabase,
     events: EventsDatabase,
     pings: PingsDatabase
-  }
+  };
```
Dexterp37, line 54: Mh, was this from the linter? If yes, why wasn't it failing before? If not, why are we adding it? :-) Do we need a new linting rule?

Dexterp37 (approved):
I like this approach more than the previous one, thanks for taking time to go through this again. I appreciate the effort!

glean/src/core/glean.ts
brizental: YOU CAUGHT ME! Yeah I think we need a lint rule, we _have_ a lint rule for enforcing semi colons but it doesn't seem to have effect in class properties.

brizental: Eslint doesn't have such a rule, sadly. I found this plugin https://github.com/marudor/eslint-plugin-class-property/blob/master/docs/rules/class-property-semicolon.md but it is archived and I'd rather not go that way...

glean/src/core/pings/maker.ts
brizental: Hm, interesting. I thought `??` was just sugar for `something ? something : somethingElse`. Thanks for the link :D

glean/src/core/glean.ts
Dexterp37: Yes, makes sense. Ignore this for now. Maybe file a bug in case things change in the future!

PR #96, opened 2021-03-04
Bug 1689547 - First attempt at providing a ping encryption plugin

brizental:
Opening this as a draft to get feedback on the encryption implementation.

### DIFF
--- a/glean/package.json
+++ b/glean/package.json
@@ -18,6 +18,11 @@
       "browser": "./dist/webext/browser/core/pings/index.js",
       "import": "./dist/webext/esm/core/pings/index.js",
       "require": "./dist/webext/cjs/core/pings/index.js"
+    },
+    "./webext/plugins/*": {
+      "browser": "./dist/webext/browser/plugins/*.js",
+      "import": "./dist/webext/esm/plugins/*.js",
+      "require": "./dist/webext/cjs/plugins/*.js"
     }
   },
   "typesVersions": {
@@ -30,6 +35,9 @@
       ],
       "webext/private/metrics/*": [
         "./dist/webext/types/core/metrics/types/*"
+      ],
+      "webext/plugins/*": [
+        "./dist/webext/types/plugins/*"
       ]
     }
   },
--- a/glean/package.json
+++ b/glean/package.json
@@ -96,6 +96,7 @@
     "webpack-cli": "^4.5.0"
   },
   "dependencies": {
+    "jose": "^3.7.0",
     "uuid": "^8.3.2"
   }
 }
--- a/glean/package.json
+++ b/glean/package.json
@@ -39,7 +39,7 @@
     "dist/**/*"
   ],
   "scripts": {
-    "test": "npm run test:core && npm run test:platform",
+    "test": "npm run test:core && npm run test:platform && npm run test:plugins",
     "test:core": "ts-mocha \"tests/core/**/*.spec.ts\" --recursive",
     "test:plugins": "ts-mocha \"tests/plugins/**/*.spec.ts\" --recursive",
     "test:platform": "npm run build:test-webext && ts-mocha \"tests/platform/**/*.spec.ts\" --recursive --timeout 0",
--- a/glean/package.json
+++ b/glean/package.json
@@ -41,6 +41,7 @@
   "scripts": {
     "test": "npm run test:core && npm run test:platform",
     "test:core": "ts-mocha \"tests/core/**/*.spec.ts\" --recursive",
+    "test:plugins": "ts-mocha \"tests/plugins/**/*.spec.ts\" --recursive",
     "test:platform": "npm run build:test-webext && ts-mocha \"tests/platform/**/*.spec.ts\" --recursive --timeout 0",
     "build:test-webext": "cd tests/platform/utils/webext/sample/ && npm install && npm run build:xpi",
     "lint": "eslint . --ext .ts,.js,.json --max-warnings=0",
--- a/glean/src/core/events/index.ts
+++ b/glean/src/core/events/index.ts
@@ -12,53 +12,57 @@ export class CoreEvent<
    Context extends unknown[] = unknown[],
    // The expected type of the action result. To be returned by the plugin.
    Result extends unknown = unknown
- > {
-   // The plugin to be triggered eveytime this even occurs.
-   private plugin?: Plugin<CoreEvent<Context, Result>>;
+> {
+  // The plugin to be triggered eveytime this even occurs.
+  private plugin?: Plugin<CoreEvent<Context, Result>>;
 
-   constructor(readonly name: string) {}
+  constructor(readonly name: string) {}
 
-   /**
-    * Registers a plugin that listens to this event.
-    *
-    * @param plugin The plugin to register.
-    */
-   registerPlugin(plugin: Plugin<CoreEvent<Context, Result>>): void {
-     if (this.plugin) {
-       console.error(
-         `Attempted to register plugin '${plugin.name}', which listens to the event '${plugin.event}'.`,
-         `That event is already watched by plugin '${this.plugin.name}'`,
-         `Plugin '${plugin.name}' will be ignored.`
-       );
-       return;
-     }
- 
-     this.plugin = plugin;
-   }
- 
-   /**
-    * Deregisters the currently registered plugin.
-    *
-    * If no plugin is currently registered this is a no-op.
-    */
-   deregisterPlugin(): void {
-     this.plugin = undefined;
-   }
+  get registeredPluginIdentifier(): string | undefined {
+    return this.plugin?.name;
+  }
 
-   /**
-    * Triggers this event.
-    *
-    * Will execute the action of the registered plugin, if there is any.
-    *
-    * @param args The arguments to be passed as context to the registered plugin.
-    *
-    * @returns The result from the plugin execution.
-    */
-   trigger(...args: Context): Result | void {
-     if (this.plugin) {
-       return this.plugin.action(...args);
-     }
-   }
+  /**
+   * Registers a plugin that listens to this event.
+   *
+   * @param plugin The plugin to register.
+   */
+  registerPlugin(plugin: Plugin<CoreEvent<Context, Result>>): void {
+    if (this.plugin) {
+      console.error(
+        `Attempted to register plugin '${plugin.name}', which listens to the event '${plugin.event}'.`,
+        `That event is already watched by plugin '${this.plugin.name}'`,
+        `Plugin '${plugin.name}' will be ignored.`
+      );
+      return;
+    }
+
+    this.plugin = plugin;
+  }
+
+  /**
+   * Deregisters the currently registered plugin.
+   *
+   * If no plugin is currently registered this is a no-op.
+   */
+  deregisterPlugin(): void {
+    this.plugin = undefined;
+  }
+
+  /**
+   * Triggers this event.
+   *
+   * Will execute the action of the registered plugin, if there is any.
+   *
+   * @param args The arguments to be passed as context to the registered plugin.
+   *
+   * @returns The result from the plugin execution.
+   */
+  trigger(...args: Context): Result | void {
+    if (this.plugin) {
+      return this.plugin.action(...args);
+    }
+  }
 }
 
 /**
--- a/glean/src/core/pings/maker.ts
+++ b/glean/src/core/pings/maker.ts
@@ -209,7 +209,7 @@ export async function collectPing(ping: PingType, reason?: string): Promise<Ping
  *
  * @returns The final submission path.
  */
-function makePath(identifier: string, ping: PingType): string {
+export function makePath(identifier: string, ping: PingType): string {
   // We are sure that the applicationId is not `undefined` at this point,
   // this function is only called when submitting a ping
   // and that function return early when Glean is not initialized.
--- a/glean/src/core/pings/maker.ts
+++ b/glean/src/core/pings/maker.ts
@@ -237,15 +237,25 @@ export async function collectAndStorePing(identifier: string, ping: PingType, re
     return;
   }
 
+  let modifiedPayload;
+  try {
+    modifiedPayload = await CoreEvents.afterPingCollection.trigger(collectedPayload);
+  } catch(e) {
+    console.error(
+      `Error while attempting to modify ping payload for the "${ping.name}" ping using`,
+      `the ${JSON.stringify(CoreEvents.afterPingCollection.registeredPluginIdentifier)} plugin.`,
+      "Ping will not be submitted. See more logs below.\n\n",
+      e
+    );
+
+    return;
+  }
+
   if (Glean.logPings) {
     console.info(JSON.stringify(collectedPayload, null, 2));
   }
-
-  const headers = getPingHeaders();
-
-  const modifiedPayload = await CoreEvents.afterPingCollection.trigger(collectedPayload);
   const finalPayload = modifiedPayload ? modifiedPayload : collectedPayload;
-
+  const headers = getPingHeaders();
   return Glean.pingsDatabase.recordPing(
     makePath(identifier, ping),
     identifier,
--- a/glean/src/plugins/encryption.ts
+++ b/glean/src/plugins/encryption.ts
@@ -11,6 +11,9 @@ import { PingPayload } from "../core/pings/database";
 import { JSONObject } from "../core/utils";
 import CoreEvents from "../core/events";
 
+const JWE_ALGORTIHM = "ECDH-ES";
+const JWE_CONTENT_ENCODING = "A256GCM";
+
 /**
  * A plugin that listens for the `afterPingCollection` event and encrypts **all** outgoing pings
  * with the JWK provided upon initialization.
@@ -28,25 +31,19 @@ class PingEncryptionPlugin extends Plugin<typeof CoreEvents["afterPingCollection
    * Creates a new PingEncryptionPlugin instance.
    *
    * @param jwk The JWK that will be used to encode outgoing ping payloads.
-   * @param alg The algorithm this plugin will use for parsing the JWK. If this argument is not present,
-   *            we will look for the `alg` key in the JWK. If neither is present we default to "ECDH-ES".
    */
-  constructor(private jwk: JWK, private alg?: string) {
+  constructor(private jwk: JWK) {
     super(CoreEvents["afterPingCollection"].name, "pingEncryptionPlugin");
-
-    if (!alg) {
-      this.alg = jwk.alg ? jwk.alg : "ECDH-ES";
-    }
   }
 
   async action(payload: PingPayload): Promise<JSONObject> {
-    const key = await parseJwk(this.jwk, this.alg);
+    const key = await parseJwk(this.jwk, JWE_ALGORTIHM);
     const encoder = new TextEncoder();
     const encodedPayload = await new CompactEncrypt(encoder.encode(JSON.stringify(payload)))
       .setProtectedHeader({
         kid: this.jwk.kid,
-        alg: this.alg,
-        enc: "A256GCM",
+        alg: JWE_ALGORTIHM,
+        enc: JWE_CONTENT_ENCODING,
         typ: "JWE",
       })
       .encrypt(key);
--- a/glean/src/plugins/encryption.ts
+++ b/glean/src/plugins/encryption.ts
@@ -4,6 +4,7 @@
 
 import CompactEncrypt from "jose/jwe/compact/encrypt";
 import parseJwk from "jose/jwk/parse";
+import calculateThumbprint from "jose/jwk/thumbprint";
 import { JWK } from "jose/types";
 
 import Plugin from "./index";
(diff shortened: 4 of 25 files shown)


glean/src/plugins/encryption.ts
```diff
@@ -0,0 +1,32 @@
...
+ */
+class PingEncryptionPlugin extends Plugin<typeof CoreEvents["afterPingCollection"]> {
+  constructor(readonly jwk: JWK) {
+    super(CoreEvents["afterPingCollection"].name, "pingEncryptionPlugin");
+  }
+
+  async action(payload: PingPayload): Promise<JSONObject> {
+    const key = await parseJwk(this.jwk);
+    const encoder = new TextEncoder();
+    const encodedPayload = await new CompactEncrypt(encoder.encode(JSON.stringify(payload)))
```
brizental, line 25: Question 1: `payload` comes as an object, it is correct that I stringify it?

glean/src/plugins/encryption.ts
```diff
@@ -0,0 +1,32 @@
...
+      .setProtectedHeader({ kid: this.jwk.kid })
```
brizental, line 26: Question 2: I am setting this header as described in the proposal:
> Since JSON is the supported transport for our data ingestion platform, the encrypted payload will be in the JWE Compact Serialization format and must include a key id (kid) header identifying the public key used for encryption. 

I saw in the example keys you sent me that the keys don't have this headers though. If this key is required, should I validate that the JWK contains all the expected keys upon instantiation?

glean/src/plugins/encryption.ts
```diff
@@ -0,0 +1,32 @@
...
+import Plugin from "./index";
+import { PingPayload } from "../core/pings/database";
+import { JSONObject } from "../core/utils";
+import CoreEvents from "../core/events";
+
+/**
+ * A plugin that encrypts the payload of a ping before we store it.
+ */
+class PingEncryptionPlugin extends Plugin<typeof CoreEvents["afterPingCollection"]> {
+  constructor(readonly jwk: JWK) {
```
brizental, line 18: I am using the provided JWK type from jose (https://github.com/panva/jose/blob/main/src/types.d.ts#L10-L68), all the properties are optioonal on it. Are we requiring any specific properties and should I validate that they exist before instantiation?

glean/src/plugins/encryption.ts
```diff
@@ -0,0 +1,32 @@
...
+    super(CoreEvents["afterPingCollection"].name, "pingEncryptionPlugin");
+  }
+
+  async action(payload: PingPayload): Promise<JSONObject> {
+    const key = await parseJwk(this.jwk);
```
brizental, line 23: This function will either gets the algorith as an argument of infers it from the `alg` property of the JWK. Do our JWKs contain this property? Or are we using the same algortihm always and I should just hard-code it here?

glean/src/plugins/encryption.ts
```diff
@@ -0,0 +1,37 @@
...
+    const encoder = new TextEncoder();
+    const encodedPayload = await new CompactEncrypt(encoder.encode(JSON.stringify(payload)))
+      .setProtectedHeader({
+        kid: this.jwk.kid,
+        alg: this.jwk.alg,
+        enc: "A256GCM",
+        typ: "JWE",
+      })
```
brizental, line 31: I am setting this header as described in the proposal:

> Since JSON is the supported transport for our data ingestion platform, the encrypted payload will be in the JWE Compact Serialization format and must include a key id (kid) header identifying the public key used for encryption.

I saw in the example keys you sent me that the keys don't have this header though. If this key is required, should I validate that the JWK contains all the expected keys upon instantiation?

glean/src/plugins/encryption.ts
```diff
@@ -0,0 +1,37 @@
...
+  constructor(readonly jwk: JWK) {
+    super(CoreEvents["afterPingCollection"].name, "pingEncryptionPlugin");
+  }
+
+  async action(payload: PingPayload): Promise<JSONObject> {
+    const key = await parseJwk(this.jwk);
+    const encoder = new TextEncoder();
+    const encodedPayload = await new CompactEncrypt(encoder.encode(JSON.stringify(payload)))
+      .setProtectedHeader({
+        kid: this.jwk.kid,
```
brizental, line 27: I am setting this header as described in the proposal:

> Since JSON is the supported transport for our data ingestion platform, the encrypted payload will be in the JWE Compact Serialization format and must include a key id (kid) header identifying the public key used for encryption.

I saw in the example keys you sent me that the keys don't have this header though. If this key is required, should I validate that the JWK contains all the expected keys upon instantiation?

glean/tests/plugins/encryption.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    await Glean.testUninitialize();
+
+    const plugin = new PingEncryptionPlugin({
+      "kid": "test",
+      "alg": "ECDH-ES",
+      "crv": "P-256",
+      "kty": "EC",
+      "x": "Qqihp7EryDN2-qQ-zuDPDpy5mJD5soFBDZmzPWTmjwk",
+      "y": "PiEQVUlywi2bEsA3_5D0VFrCHClCyUlLW52ajYs-5uc"
```
brizental, line 36: Key copied over from: https://github.com/mozilla/gcp-ingestion/blob/master/ingestion-beam/src/test/resources/pioneer/study-foo.public.json

Generated payload for reference: 
```
'eyJraWQiOiJ0ZXN0IiwiYWxnIjoiRUNESC1FUyIsImVuYyI6IkEyNTZHQ00iLCJ0eXAiOiJKV0UiLCJlcGsiOnsiY3J2IjoiUC0yNTYiLCJrdHkiOiJFQyIsIngiOiJocWRHSTlTcFRhX0FHSkhGdnVwa0xQcTZJUWZvdHZGdHBUYUEzeklyVEpZIiwieSI6IlRqRG9tVUpubU5XRVVZT3J3dV9XRmJLTVc0Q2Z3NldqQmVqbzQwUjVwMm8ifX0..Uka-v-MUTLfJ3o0I.-iYdgmD9pYfGmxxIyms4Ppij7AroTrB2CQZh6SIaV8mysZxa0PyNMAfKi3tU2ts8mGcW4SUz1oLCUt7G-p-JShAqNxj2iK_9myNNsvBiHPlKurik-djRq3T6GYvkmk960o21n0-zoZhpU2JRcnwFs09WMa5lCSsWmbURKGkiG1-Zqymyao4yFKv6ITpYFnce_4XtejtEr9aaiC12FQ0H1r7QCbujx2z1XYLGiNqYyYeb-42jvtZZQnoxG4WjNGpBhGuEkBDHqgC_ALmjx4scZtNtV-mTGxPS3CtQodfYtZHcs2PYRVQRzO8reSrnv40p9O0mcCQWpJLdFWhPTgUyhT3jQ0HgrgG7leRgksIj-L_ftjwGwPBDVKSrtAA5uv93iImDS4rHhHBPXAZyvqmr8fps9DB55yW094-IJrDrVimmVkzeoKTIOwvmc_6DwQDDcC5gAXPXgHknU_RoPYNeBN1kiSTmzNC9VL3LzyOG.AGPJZ5I2SutdWxttwRoyeA'
```

brizental: Included `alg` and `kid` of course.

glean/src/plugins/encryption.ts
```diff
@@ -0,0 +1,37 @@
...
+        alg: this.jwk.alg,
+        enc: "A256GCM",
+        typ: "JWE",
+      })
+      .encrypt(key);
+    return { payload: encodedPayload };
```
brizental, line 33: Self reminder: This generated the following payload:

```js
 payload: {
  "payload": "eyJraWQiO..."
}
```

Obviously incorrect, will probably need to accept `JSONValue` on the ping maker in order for the plugin to be allowed to return simply a string.

glean/src/plugins/encryption.ts
acmiyaguchi: This is what an example of a provisioned JWK looks like (https://bugzilla.mozilla.org/show_bug.cgi?id=1693305#c9). 

```
{
    "crv": "P-256",
    "kid": "citp-news-disinfo-two",
    "kty": "EC",
    "x": "Q20tsJdrryWJeuPXTM27wIPb_YbsdYPpkK2N9O6aXwM",
    "y": "1onW1swaCcN1jkmkIwhXpCm55aMP8GRJln5E8WQKLJk"
}
```

I don't think you should have to do anything special in order to parse the key correctly.

glean/src/plugins/encryption.ts
acmiyaguchi: The properties on the key depends on the type of encryption it's for, which could potentially change in the future. I don't think it necessary to validate the keys outside the encryption function actually succeeding.

glean/src/plugins/encryption.ts
acmiyaguchi: The keys don't need the header, but the encrypted payload does as per the spec: https://tools.ietf.org/html/rfc7516#page-8. Specifically:

```
  In the JWE Compact Serialization, a JWE is represented as the
   concatenation:

      BASE64URL(UTF8(JWE Protected Header)) || '.' ||
      BASE64URL(JWE Encrypted Key) || '.' ||
      BASE64URL(JWE Initialization Vector) || '.' ||
      BASE64URL(JWE Ciphertext) || '.' ||
      BASE64URL(JWE Authentication Tag)
```

The JWE encrypted key is the second element, while the ciphertext is the 3rd element. The library should take care of setting the parameters of encryption for the most part.  Looking at the JWE implemention in Firefox, it looks like the key is in the protected header (https://searchfox.org/mozilla-central/rev/eeb8cf278192d68b3977d0adb4d43f1463439269/services/crypto/modules/jwcrypto.jsm#129-136), so you might be right on this one. The decoder doesn't actually look at the kid atm and chooses a key based purely on the the routing parameters in the submission url.

acmiyaguchi: I think I mispoke, this is pretty much what my testing code does. https://github.com/mozilla/gcp-ingestion/blob/97619c801279670bddba8ce200799e8b8cda5e60/ingestion-beam/src/test/resources/pioneer.py#L53-L58

The algorithm is "ECDH-ES".

brizental:
@acmiyaguchi I attended to your review comments in 9b69d77 this commit :)

glean/src/plugins/encryption.ts
brizental: Yeah, this was not the case. I guess I got confused because we have the `payload` property inside the the payload now. I left a bit more comprehensive tests in 92d7825 just to be sure.

glean/src/plugins/encryption.ts
```diff
@@ -0,0 +1,57 @@
...
+ * }
+ * ```
+ */
+class PingEncryptionPlugin extends Plugin<typeof CoreEvents["afterPingCollection"]> {
+  /**
+   * Creates a new PingEncryptionPlugin instance.
+   *
+   * @param jwk The JWK that will be used to encode outgoing ping payloads.
+   * @param alg The algorithm this plugin will use for parsing the JWK. If this argument is not present,
+   *            we will look for the `alg` key in the JWK. If neither is present we defaut to "ECDH-ES".
```
acmiyaguchi, line 32:
```suggestion
   *            we will look for the `alg` key in the JWK. If neither is present we default to "ECDH-ES".
```

acmiyaguchi (commented):
Thanks for addressing my comments :) What do you still have left in this PR?

brizental:
> Thanks for addressing my comments :) What do you still have left in this PR?

If I have your green light on the encryption I'll just request Alessio's review for sanity check on the Glean.js specific parts and merge! Thanks @acmiyaguchi :D

glean/package.json
```diff
@@ -41,6 +41,7 @@
   "scripts": {
     "test": "npm run test:core && npm run test:platform",
     "test:core": "ts-mocha \"tests/core/**/*.spec.ts\" --recursive",
+    "test:plugins": "ts-mocha \"tests/plugins/**/*.spec.ts\" --recursive",
```
brizental, line 44: Forgot to add this command to the `test` command above.

glean/src/plugins/encryption.ts
```diff
@@ -0,0 +1,57 @@
...
+   *
+   * @param jwk The JWK that will be used to encode outgoing ping payloads.
+   * @param alg The algorithm this plugin will use for parsing the JWK. If this argument is not present,
+   *            we will look for the `alg` key in the JWK. If neither is present we default to "ECDH-ES".
+   */
+  constructor(private jwk: JWK, private alg?: string) {
+    super(CoreEvents["afterPingCollection"].name, "pingEncryptionPlugin");
+
+    if (!alg) {
+      this.alg = jwk.alg ? jwk.alg : "ECDH-ES";
```
Dexterp37, line 38: Why this default? Should we throw instead?

glean/tests/plugins/encryption.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+  beforeEach(async function() {
+    await Glean.testResetGlean("something something");
+  });
+
+  // eslint-disable-next-line mocha/no-hooks-for-single-case
+  afterEach(function () {
+    sandbox.restore();
+  });
+
+  it("collect and store triggers the AfterPingCollection and deals with possible result correctly", async function () {
```
Dexterp37, line 28: Can you add a test that provides an `alg`?

glean/tests/plugins/encryption.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+    await Glean.testUninitialize();
+    await Glean.testInitialize(
```
Dexterp37, line 30: Can't both this lines be summarized to `Glean.testResetGlean(...)`? If not, why?

glean/tests/plugins/encryption.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+      "something something",
```
Dexterp37, line 31: Let's try to be more specific here: `gleanjs.unit.test` (in case we see these in the wild)?

glean/src/plugins/encryption.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  async action(payload: PingPayload): Promise<JSONObject> {
+    const key = await parseJwk(this.jwk, this.alg);
+    const encoder = new TextEncoder();
+    const encodedPayload = await new CompactEncrypt(encoder.encode(JSON.stringify(payload)))
+      .setProtectedHeader({
+        kid: this.jwk.kid,
+        alg: this.alg,
+        enc: "A256GCM",
```
Dexterp37, line 49: Doesn't this depend on `alg`, too?

Dexterp37 (changes requested):
I left a few comments below. Did you check that "log pings" allow us to ... log pings with this?

glean/tests/plugins/encryption.spec.ts
brizental: Hah, sure. Should maybe do that with all tests though, I use this app id everywhere.

glean/src/plugins/encryption.ts
brizental: https://github.com/mozilla/glean.js/pull/96#discussion_r587880928

glean/src/plugins/encryption.ts
brizental: 🤔 Hm, yeap I think you are right. I'll just go ahead and support `"ECDH-ES"` do avoid confusion. That is the only alg we need at the moment IIUC. We cross the multiple algs bridge when we get there.

glean/tests/plugins/encryption.spec.ts
Dexterp37: Yes, that's fine to take it outside of this PR of course. But you might as well just with this file :)

brizental:
> Did you check that "log pings" allow us to ... log pings with this?

Nope, but I can check that and even go further and a test for it.

Dexterp37:
> 
> 
> > Did you check that "log pings" allow us to ... log pings with this?
> 
> Nope, but I can check that and even go further and a test for it.

Yes please :)

glean/src/plugins/encryption.ts
brizental: Well, I according to the RFC `enc` is related to the content encryption, so _I think_ it is unrelated to `alg`.

Dexterp37: Then the follow-up question is: why this default? :) Can you leave a comment about it in the code?

brizental: Ok I tried to understand why, but the truth is I copied from https://github.com/mozilla/gcp-ingestion/blob/97619c801279670bddba8ce200799e8b8cda5e60/ingestion-beam/src/test/resources/pioneer.py#L53-L58. @acmiyaguchi do you know the eact reason for this choice?

During my research I learned that `alg` and `enc` do seem to be related somehow, some pairs will fail. I will go ahead with what I said before and support only what we need for now.

brizental: If we are leaving a comment about this we will probably want to leave a comment on the reasoning behind choosing the `ECDH-ES` algorithm, so @acmiyaguchi if you know the reason for that as well, please do tell :)

glean/tests/plugins/encryption.spec.ts
brizental: https://github.com/mozilla/glean.js/pull/96/commits/0909b35e4b54aeb3ca467d799300b77bfc8153b6

glean/src/plugins/encryption.ts
```diff
@@ -11,6 +11,9 @@ import { PingPayload } from "../core/pings/database";
 import { JSONObject } from "../core/utils";
 import CoreEvents from "../core/events";
 
+const JWE_ALGORTIHM = "ECDH-ES";
+const JWE_CONTENT_ENCODING = "A256GCM";
+
 /**
  * A plugin that listens for the `afterPingCollection` event and encrypts **all** outgoing pings
  * with the JWK provided upon initialization.
```
Dexterp37, line 19: Let's mention that this offers a specific encryption mechanism in the comment here.

brizental: Yeah I left this open on purpose. I will add descriptions and more descriptive doc strings there once Anthony gets back  to us here https://github.com/mozilla/glean.js/pull/96#discussion_r592271956

glean/src/plugins/encryption.ts
acmiyaguchi: ECDH-ES is the only algorithm supported by Firefox's hand-rolled JWE implementation, and so :rhelmer ended up setting most of the parameters that we use. https://searchfox.org/mozilla-central/rev/eeb8cf278192d68b3977d0adb4d43f1463439269/services/crypto/modules/jwcrypto.jsm#58-74

glean/src/core/events/index.ts
```diff
@@ -12,53 +12,57 @@ export class CoreEvent<
    Context extends unknown[] = unknown[],
    // The expected type of the action result. To be returned by the plugin.
    Result extends unknown = unknown
- > {
-   // The plugin to be triggered eveytime this even occurs.
-   private plugin?: Plugin<CoreEvent<Context, Result>>;
+> {
+  // The plugin to be triggered eveytime this even occurs.
```
brizental, line 16: Since we are here...

```suggestion
  // The plugin to be triggered eveytime this event occurs.
```

glean/src/core/pings/maker.ts
```diff
@@ -223,15 +223,25 @@ export async function collectAndStorePing(identifier: string, ping: PingType, re
...
   }
 
+  let modifiedPayload;
+  try {
+    modifiedPayload = await CoreEvents.afterPingCollection.trigger(collectedPayload);
+  } catch(e) {
+    console.error(
+      `Error while attempting to modify ping payload for the "${ping.name}" ping using`,
+      `the ${JSON.stringify(CoreEvents.afterPingCollection.registeredPluginIdentifier)} plugin.`,
+      "Ping will not be submitted. See more logs below.\n\n",
```
brizental, line 233:
```suggestion
      "Ping will not be submitted. See more logs below.\n",
```

Why did I add two line breaks...

glean/src/plugins/encryption.ts
```diff
@@ -0,0 +1,60 @@
...
+   * Creates a new PingEncryptionPlugin instance.
+   *
+   * @param jwk The JWK that will be used to encode outgoing ping payloads.
+   */
+  constructor(private jwk: JWK) {
+    super(CoreEvents["afterPingCollection"].name, "pingEncryptionPlugin");
+  }
+
+  async action(payload: PingPayload): Promise<JSONObject> {
+    const key = await parseJwk(this.jwk, JWE_ALGORITHM);
```
brizental, line 46: Note: This will throw an error in case the provided `jwk` is incorrect for the default algorithm.

glean/src/core/pings/maker.ts
```diff
@@ -223,15 +223,25 @@ export async function collectAndStorePing(identifier: string, ping: PingType, re
     return;
   }
 
+  let modifiedPayload;
+  try {
```
brizental, line 227: I considered just adding the `try / catch` block inside the `trigger` method, but I think this is something to decide in an event by event basis. Here I thought it would be best to not send a ping if a plugin errors, for the encryption plugin, sending the unencrypted payload would probably be a terrible decision.

Dexterp37 (approved):

PR #97, opened 2021-03-09
Add labeled metric types

Dexterp37:
This introduces the labelled metric types for string, boolean and counter types. The implementation closely follows what's in the Glean SDK, with no major deviation.

Note that this required adding a couple more functions to the database.

TODO:

- [x] Port test coverage.
- [x] Enable dynamic labels.
- [ ] Add documentation.
- [x] Enforce label limits.
- [x] Enable different labelled metric types (instead of hardcoding)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * [#101](https://github.com/mozilla/glean.js/pull/101): BUGFIX: Only validate Debug View Tag and Source Tags when they are present.
 * [#101](https://github.com/mozilla/glean.js/pull/101): BUGFIX: Only validate Debug View Tag and Source Tags when they are present.
 * [#102](https://github.com/mozilla/glean.js/pull/102): BUGFIX: Include a Glean User-Agent header in all pings.
+* [#97](https://github.com/mozilla/glean.js/pull/97): Add support for labelled metric types (string, boolean and counter).
 
 # v0.4.0 (2021-03-10)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 * [#101](https://github.com/mozilla/glean.js/pull/101): BUGFIX: Only validate Debug View Tag and Source Tags when they are present.
 * [#101](https://github.com/mozilla/glean.js/pull/101): BUGFIX: Only validate Debug View Tag and Source Tags when they are present.
 * [#102](https://github.com/mozilla/glean.js/pull/102): BUGFIX: Include a Glean User-Agent header in all pings.
-* [#97](https://github.com/mozilla/glean.js/pull/97): Add support for labelled metric types (string, boolean and counter).
+* [#97](https://github.com/mozilla/glean.js/pull/97): Add support for labeled metric types (string, boolean and counter).
 
 # v0.4.0 (2021-03-10)
 
--- a/glean/src/core/metrics/database.ts
+++ b/glean/src/core/metrics/database.ts
@@ -5,7 +5,7 @@
 import Store from "../storage";
 import { MetricType, Lifetime, Metric } from "./";
 import { createMetric, validateMetricInternalRepresentation } from "./utils";
-import { isObject, isUndefined, JSONValue } from "../utils";
+import { isObject, isUndefined, JSONObject, JSONValue } from "../utils";
 import Glean from "../glean";
 
 export interface Metrics {
@@ -133,11 +133,48 @@ class MetricsDatabase {
     }
 
     const store = this._chooseStore(metric.lifetime);
-    const storageKey = metric.identifier;
+    const storageKey = await metric.getAsyncIdentifier();
     for (const ping of metric.sendInPings) {
       const finalTransformFn = (v?: JSONValue): JSONValue => transformFn(v).get();
       await store.update([ping, metric.type, storageKey], finalTransformFn);
     }
+  }
+
+  /**
+   * Checks if anything was stored for the provided metric.
+   *
+   * @param lifetime the metric `Lifetime`.
+   * @param ping the ping storage to search in.
+   * @param metricType the type of the metric.
+   * @param metricIdentifier the metric identifier.
+   *
+   * @returns `true` if the metric was found (regardless of the validity of the
+   *          stored data), `false` otherwise.
+   */
+  async hasMetric(lifetime: Lifetime, ping: string, metricType: string, metricIdentifier: string): Promise<boolean> {
+    const store = this._chooseStore(lifetime);
+    const value = await store.get([ping, metricType, metricIdentifier]);
+    return !isUndefined(value);
+  }
+
+  /**
+   * Counts the number of stored metrics with an id starting with a specific identifier.
+   *
+   * @param lifetime the metric `Lifetime`.
+   * @param ping the ping storage to search in.
+   * @param metricType the type of the metric.
+   * @param metricIdentifier the metric identifier.
+   *
+   * @returns the number of stored metrics with their id starting with the given identifier.
+   */
+  async countByBaseIdentifier(lifetime: Lifetime, ping: string, metricType: string, metricIdentifier: string): Promise<number> {
+    const store = this._chooseStore(lifetime);
+    const pingStorage = await store.get([ping, metricType]);
+    if (isUndefined(pingStorage)) {
+      return 0;
+    }
+
+    return Object.keys(pingStorage).filter(n => n.startsWith(metricIdentifier)).length;
   }
 
   /**
@@ -168,10 +205,10 @@ class MetricsDatabase {
     metric: MetricType
   ): Promise<T | undefined> {
     const store = this._chooseStore(metric.lifetime);
-    const storageKey = metric.identifier;
+    const storageKey = await metric.getAsyncIdentifier();
     const value = await store.get([ping, metric.type, storageKey]);
     if (!isUndefined(value) && !validateMetricInternalRepresentation<T>(metric.type, value)) {
-      console.error(`Unexpected value found for metric ${metric.identifier}: ${JSON.stringify(value)}. Clearing.`);
+      console.error(`Unexpected value found for metric ${storageKey}: ${JSON.stringify(value)}. Clearing.`);
       await store.delete([ping, metric.type, storageKey]);
       return;
     } else {
@@ -210,6 +247,30 @@ class MetricsDatabase {
     return data;
   }
 
+  private processLabeledMetric(snapshot: Metrics, metricType: string, metricId: string, metricData: JSONValue) {
+    const newType = `labeled_${metricType}`;
+    const idLabelSplit = metricId.split("/", 2);
+    const newId = idLabelSplit[0];
+    const label = idLabelSplit[1];
+
+    if (newType in snapshot && newId in snapshot[newType]) {
+      // Other labels were found for this metric. Do not throw them away.
+      const existingData = snapshot[newType][newId];
+      snapshot[newType][newId] = {
+        ...(existingData as JSONObject),
+        [label]: metricData
+      };
+    } else {
+      // This is the first label for this metric.
+      snapshot[newType] = {
+        ...snapshot[newType],
+        [newId]: {
+          [label]: metricData
+        }
+      };
+    }
+  }
+
   /**
    * Gets all of the persisted metrics related to a given ping.
    *
@@ -228,13 +289,21 @@ class MetricsDatabase {
       await this.clear(Lifetime.Ping, ping);
     }
 
-    const response: Metrics = { ...pingData };
-    for (const data of [userData, appData]) {
+    const response: Metrics = {};
+    for (const data of [userData, pingData, appData]) {
       for (const metricType in data) {
-        response[metricType] = {
-          ...response[metricType],
-          ...data[metricType]
-        };
+        for (const metricId in data[metricType]) {
+          if (metricId.includes("/")) {
+            // While labeled data is stored within the subtype storage (e.g. counter storage), it
+            // needs to live in a different section of the ping payload (e.g. `labeled_counter`).
+            this.processLabeledMetric(response, metricType, metricId, data[metricType][metricId]);
+          } else {
+            response[metricType] = {
+              ...response[metricType],
+              [metricId]: data[metricType][metricId]
+            };
+          }
+        }
       }
     }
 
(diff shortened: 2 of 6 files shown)


glean/src/core/metrics/types/labeled.ts
```diff
@@ -0,0 +1,92 @@
...
+import Glean from "../../glean";
+import CounterMetricType from "./counter";
+import BooleanMetricType from "./boolean";
+import { JSONValue } from "../../../../dist/webext/types/core/utils";
+
+const MAX_LABELS = 16;
+const OTHER_LABEL = "__other__";
+const MAX_LABEL_LENGTH = 61;
+
+export class LabeledMetric extends Metric<JSONValue, JSONValue> {
```
Dexterp37, line 16: Note for @brizental as you can see, this is not validating anything at all. That's because the internal representation is being handled by the sub-metric type. So we just copy stuff over with this.

glean/src/core/metrics/types/labeled.ts
```diff
@@ -0,0 +1,92 @@
...
+  constructor(v: unknown) {
+    super(v);
+  }
+
+  validate(v: unknown): v is JSONValue {
+    return true;
+  }
+
+  payload(): JSONValue {
```
brizental, line 25: Random thought: we should provide a default for this method that does this. We have other metrics that just return `this._inner`, I think most of them do this even. Anyways, no need to do that in this PR.

glean/src/core/metrics/types/labeled.ts
```diff
@@ -0,0 +1,92 @@
...
+    return this._inner;
+  }
+}
+
+class LabeledMetricType<T extends CounterMetricType> {
+  // Define an index signature (https://www.typescriptlang.org/docs/handbook/interfaces.html#indexable-types)
```
brizental, line 31: This link in deprecated. I think we can leave this comment out. This is a normal Typescript pattern and I have used in other parts of the code anyways :D (https://github.com/mozilla/glean.js/blob/main/glean/src/core/events/index.ts#L69). To make things clearer, maybe instead of `[key: string]` we use `[label: string]`?

glean/src/core/metrics/utils.ts
```diff
@@ -20,6 +21,7 @@ const METRIC_MAP: {
   "boolean": BooleanMetric,
   "counter": CounterMetric,
   "datetime": DatetimeMetric,
+  "labeled_counter": LabeledMetric,
```
brizental, line 24: Shouldn't this have `LabeledMetric<CounterMetricType>`?

glean/src/core/metrics/types/labeled.ts
brizental: Ok, after reading the code a bit more I understand that you are saving these metrics as if they were not labeled. That is counterintuitive for me, why can't we save them already as labeled metrics?

glean/src/core/metrics/types/labeled.ts
brizental: Or maybe `labels: Record<string, T>`?

glean/src/core/metrics/types/labeled.ts
```diff
@@ -0,0 +1,92 @@
...
+  // to make the Proxy aware of the expected return type. Note that this is
+  // required because TypeScript does not allow different input and output types
+  // in Proxy (https://github.com/microsoft/TypeScript/issues/20846). 
+  [key: string]: T;
+
+  constructor(
+    meta: CommonMetricData,
+    labels?: string[]
+  ) {
+    return new Proxy(this, {
```
brizental, line 41: Bear with me here as I have never used `Proxy` before, but after reading about it, I don't understand why we need it. Why can't `get` be a method of this class that returns `T`? We could do sorta what glean-core does and save a reference to the desired submetric constructor to this class. Like so:

```ts
constructor<A>(meta: CommonMetricData, submetric: new(...args: A) -> T, labels?: string[]) {
    this.submetric = submetric
}
```

Or we could even do _exactly_ like glean-core and provide an instance of the submetric and clone when necessary. There is apparently a crazy way to clone instances of classes in JS (https://stackoverflow.com/questions/41474986/how-to-clone-a-javascript-es6-class-instance), but we can also force metrics to define  `clone` method by making it an abstract method on the `MetricType` class.

I think something similar to this will be necessary anyways to do [this](https://github.com/mozilla/glean.js/pull/97/files#diff-25b869315615f9e430ce2ebcb8e11349dc38fdbab44706dac5d0fbaf7cb2486dR79).

glean/src/core/metrics/types/labeled.ts
brizental: Since we are currently saving metrics as their submetric type, this class is probably never used.

glean/src/core/metrics/types/labeled.ts
Dexterp37: > 
> 
> Or maybe `labels: Record<string, T>`?

This would break compilation :)

glean/src/core/metrics/database.ts
```diff
@@ -133,13 +133,50 @@ class MetricsDatabase {
...
 
+  /**
+   * Checks if anything was stored for the provided metric.
+   *
+   * @param lifetime the metric `Lifetime`.
+   * @param ping the ping storage to search in.
+   * @param metricType the type of the metric.
+   * @param metricIdentifier the metric identifier.
+   *
+   * @returns `true` if the metric was found (regardless of the validity of the
```
brizental, line 151: Why does the validity of the stored data not matter?

Dexterp37: Because this is used to count things to enforce the label limit. And in glean-core we enforce the limit regardless of the content.

brizental (changes requested):
It is Friday EOD and I probably missing something in this review. Please consider this a "first pass" and I'll take a closer look on Monday 🧇

glean/src/core/metrics/index.ts
```diff
@@ -147,3 +176,24 @@ export abstract class MetricType implements CommonMetricData {
...
 }
+
+/**
+ * This is no-op internal metric representation.
+ *
+ * This can be used to instruct the validators to simply report
+ * whatever is stored internally, without performing any specific
+ * validation.
+ */
+export class PassthroughMetric extends Metric<JSONValue, JSONValue> {
```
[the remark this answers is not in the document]
brizental: Reminder to self: I don't understand why we need the `PassthorughMetric` at all.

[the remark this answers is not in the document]
brizental: Yeah I get it now. I was wrong when I assumed Labeled metrics did not need a `Metric` declaration. Even though we don't care about their internal representation, they need a `Metric` representation for generating the ping payload. I am sorry for the back and forth here, but I would prefer if instead of `PassthrougMetric` you made this `LabeledMetric`.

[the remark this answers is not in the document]
Dexterp37: Mh, too bad :) Now, I could do that and move it back to the labeled metric implementation file. Unfortunately doing so will trigger a circular dependency that breaks the TS build :(

glean/src/core/metrics/types/labeled.ts
```diff
@@ -0,0 +1,174 @@
...
+// * `thisisfine`
+// * `_.is_fine`
+// * `this.is-fine`
+// * `this-is-fine`
+// Bad:
+// * `this.is.not_fine_due_tu_the_length_being_too_long_i_thing.i.guess`
+// * `1.not_fine`
+// * `this.$isnotfine`
+// * `-.not_fine`
+const LABEL_REGEX = /^[a-z_][a-z0-9_-]{0,29}(\.[a-z_][a-z0-9_-]{0,29})*$/;
```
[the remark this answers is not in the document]
Dexterp37: This is the regular expression I took from the comment, so I assume this covers the cases we care about. The label length is checked for in the code outside of this regex (see the MAX_LABEL_LENGTH const)

glean/src/core/metrics/database.ts
Dexterp37: > 
> 
> Hm, but this is counterintuitive for me, because a metric will be deleted in case it is invalid, so it's like it never existed. Also https://github.com/mozilla/glean/blob/main/glean-core/src/database/mod.rs#L414

From looking at the code the comment seems to refer to the fact that if reading fails, the count will be 0. We don't validate the data AFAICT! That deep we really have no proper way to know what to expect, unless we leak the type to the database, which is undesireable, IMHO, from an architectural POV

glean/src/core/metrics/index.ts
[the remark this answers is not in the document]
brizental: Argh, come on TS.

glean/src/core/metrics/database.ts
brizental: Still counter intuitive for me for the reasons I said above. BUT, you are right that glean-core doesn't validate. I confused reading with deserializing, so... Fine by me.

glean/src/core/metrics/types/labeled.ts
```diff
@@ -0,0 +1,174 @@
...
+   * Create an instance of the submetric type for the provided dynamic label.
+   *
+   * @param meta the `CommonMetricData` information for the metric.
+   * @param submetricClass the class type for the submetric.
+   * @param label the desired label to record to.
+   *
+   * @returns an instance of the submetric class type that allows to record data.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  private static createFromDynamicLabel<T extends new (...args: any) => InstanceType<T>>(
```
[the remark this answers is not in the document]
Dexterp37: The TS compiler disagrees :-) I guess that's because it's a static function

glean/src/core/metrics/database.ts
Dexterp37: > 
> 
> That would be a problem on the counting function below... I know it's annyoing...

What do you mean?

brizental: In case you were validating here it would make sense to validate below. Not the case in both functions from the previous discussion.

glean/src/core/metrics/types/labeled.ts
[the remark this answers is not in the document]
brizental: TIL! Indeed it's because it is static.

glean/src/core/metrics/index.ts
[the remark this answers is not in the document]
brizental: So, if you move [this line](https://github.com/mozilla/glean.js/pull/97/files/a70deda5c5666961b03b39ab1c894bc1d6070404#diff-92a923012f502164643de6a0da030333f187db543246ceddeb72afc3b6820d5cR7) to somewhere _after_ you define the `Metric` abstract class the error goes away.

glean/src/core/metrics/types/labeled.ts
```diff
@@ -0,0 +1,174 @@
...
+   *
+   * @returns an instance of the submetric class type that allows to record data.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  private static createFromStaticLabel<T extends new (...args: any) => InstanceType<T>>(
+    meta: CommonMetricData,
+    submetricClass: T,
+    allowedLabels: string[],
+    label: string
+  ): T {
```
brizental, line 97:
```suggestion
  ): InstanceType<T> {
```

glean/src/core/metrics/types/labeled.ts
```diff
@@ -0,0 +1,174 @@
...
+   * @param label the desired label to record to.
+   *
+   * @returns an instance of the submetric class type that allows to record data.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  private static createFromDynamicLabel<T extends new (...args: any) => InstanceType<T>>(
+    meta: CommonMetricData,
+    submetricClass: T,
+    label: string
+  ): T {
```
brizental, line 122:
```suggestion
  ): InstanceType<T> {
```

glean/src/core/metrics/types/labeled.ts
```diff
@@ -0,0 +1,174 @@
...
+
+  constructor(
+    meta: CommonMetricData,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    submetric: new (...args: any) => T,
+    labels?: string[],
+  ) {
+    return new Proxy(this, {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      get: (_target: LabeledMetricType<T>, label: string): any => {
```
brizental, line 56:
```suggestion
      get: (_target: LabeledMetricType<T>, label: string): T => {
```

We need to be mindful of returning `any` here specially because it is the user facing API.

glean/src/core/metrics/types/labeled.ts
```diff
@@ -0,0 +1,174 @@
...
+    label: string
+  ): T {
+    const newMeta: CommonMetricData = {
+      ...meta,
+      dynamicLabel: label
+    };
+    return new submetricClass(newMeta);
+  }
+
+  static async getValidDynamicLabel(metric: MetricType): Promise<string> {
```
brizental, line 130: nit: Missing documentation.

glean/tests/core/metrics/labeled.spec.ts
```diff
@@ -0,0 +1,384 @@
...
+import CounterMetricType from "../../../src/core/metrics/types/counter";
+import LabeledMetricType from "../../../src/core/metrics/types/labeled";
+import StringMetricType from "../../../src/core/metrics/types/string";
+import PingType from "../../../src/core/pings";
+
+const sandbox = sinon.createSandbox();
+
+describe("LabeledMetric", function() {
+  beforeEach(async function() {
+    await Glean.testResetGlean("gleanjs.unit.test");
```
brizental, line 21: https://github.com/mozilla/glean.js/blob/main/glean/tests/core/glean.spec.ts#L31-L34 Please create the test id the same as in other tests.

brizental (changes requested):

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 * [#101](https://github.com/mozilla/glean.js/pull/101): BUGFIX: Only validate Debug View Tag and Source Tags when they are present.
 * [#101](https://github.com/mozilla/glean.js/pull/101): BUGFIX: Only validate Debug View Tag and Source Tags when they are present.
```
brizental, line 11: Why do we have two entries here... Whathever, can be fixed when we make the next release.

brizental (approved):
Good to go, assuming the `PassthroughMetric` comments are addressed.

glean/src/core/metrics/index.ts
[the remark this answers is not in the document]
Dexterp37: I renamed this as requested and added a better comment.

PR #104, opened 2021-03-17
Bug 1698874 - Sever the circular dependencies in the database

Dexterp37:
This is the first part, taking care of the circular dependencies for the database. I decided to split the changes, because the next ones will be a bit more involved, and I'm not sure it's worth rabbit holing in it right now.

### DIFF
--- a/glean/src/core/glean.ts
+++ b/glean/src/core/glean.ts
@@ -223,9 +223,9 @@ class Glean {
 
     if (!Glean.instance._db) {
       Glean.instance._db = {
-        metrics: new MetricsDatabase(),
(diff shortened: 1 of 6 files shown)


brizental (approved):

brizental:
PlatformInfo and uploader don't have this same issue? Or is that ocming in the next PR?

Dexterp37:
> 
> 
> PlatformInfo and uploader don't have this same issue? Or is that coming in the next PR?

They do but, at least the uploader is much harder to crack. I concentrated on the lower hanging fruit here, other changes will come as separate PR (slower pace, sorry)

PR #106, opened 2021-03-18
Bumped version to 0.5.0

Dexterp37:


brizental (approved):
r+wc

Please deduplicate on the chagelog these lines:

> * [#101](https://github.com/mozilla/glean.js/pull/101): BUGFIX: Only validate Debug View Tag and Source Tags when they are present.
> * [#101](https://github.com/mozilla/glean.js/pull/101): BUGFIX: Only validate Debug View Tag and Source Tags when they are present.

PR #144, opened 2021-03-26
No bug - Add a test to verify Glean's behaviour against specific async + sync calls scenario

brizental:
Also found a small bug in `dispatcher.testUninitialze`.

### DIFF
--- a/glean/tests/core/glean.spec.ts
+++ b/glean/tests/core/glean.spec.ts
@@ -10,8 +10,9 @@ import CoreEvents from "../../src/core/events";
 import Glean from "../../src/core/glean";
 import { Lifetime } from "../../src/core/metrics";
 import StringMetricType from "../../src/core/metrics/types/string";
+import CounterMetricType from "../../src/core/metrics/types/counter";
 import PingType from "../../src/core/pings";
-import { JSONObject } from "../../src/core/utils";
+import { isObject, JSONObject } from "../../src/core/utils";
 import TestPlatform from "../../src/platform/qt";
 import Plugin from "../../src/plugins";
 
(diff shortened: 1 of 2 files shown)


glean/tests/core/glean.spec.ts
```diff
@@ -412,4 +413,72 @@ describe("Glean", function() {
...
+    });
+
+    // Synchoronous function that records and submits a ping.
+    //
+    // This is the catch, even if this is called inside an async function
+    // it will be executed to completion, guaranteeing one count per async action.
+    const recordMetricAndSendPing = () => {
+      counter.add();
+      custom.submit();
+    };
```
brizental, line 451: If we change this to be an async function like so:

```ts
const recordMetricAndSendPing = async () => {
  counter.add();
  await (new Promise<void>(resolve => setTimeout(() => resolve(), Math.random() * 100)));
  custom.submit();
};
```

We get non-deterministic values, like so `[5,  7, 10, 10, 10, 10, 10, 10, 10, 10]` or `[9, 9, 9, 9, 10, 10, 10, 10, 10, 10]`. This is expected, but might be something we want to document as a common pitfall.

At this moment it is not clear to me how common it is though, further discussion to come in the document I am writing up related to [Bug 1700902](https://bugzilla.mozilla.org/show_bug.cgi?id=1700902).

Dexterp37: That's an interesting behaviour. Yes, we should indeed document... but where? (for a followup PR)

Dexterp37 (approved):

glean/tests/core/glean.spec.ts
brizental: IMO, should be part of the userdocumentation on Glean.js' general API.

PR #155, opened 2021-03-29
Bug 1687287 - Support setting a custom uploader

Dexterp37:
This PR ended up being bigger than anticipated, sorry. It does:

- Make `Uploader` a proper interface rather than an abstract class.
- Inject instances of the uploader to `PingsDatabase` rather than having it directly call `Glean.platform.uploader`.
- Fixes `WeakStore` to behave like other stores.
- Adds `httpClient` as a configuration option (the naming is consistent with the other bindings).

See the individual commit messages for more information.

### DIFF
--- a/glean/src/core/glean.ts
+++ b/glean/src/core/glean.ts
@@ -30,8 +30,10 @@ class Glean {
   private _coreMetrics: CoreMetrics;
   // Instances of Glean's core pings.
   private _corePings: CorePings;
-  // The ping uploader.
-  private _pingUploader: PingUploader
+  // The ping uploader. Note that we need to use the definite assignment assertion
+  // because initialization will not happen in the constructor, but in the `initialize`
+  // method.
+  private _pingUploader!: PingUploader
   // A task dispatcher to help execute in order asynchronous external API calls.
   private _dispatcher: Dispatcher;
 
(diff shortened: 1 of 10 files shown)


glean/src/core/glean.ts
```diff
@@ -30,8 +30,10 @@ class Glean {
   private _coreMetrics: CoreMetrics;
   // Instances of Glean's core pings.
   private _corePings: CorePings;
-  // The ping uploader.
-  private _pingUploader: PingUploader
+  // The ping uploader. Note that we need to use the definite assignment assertion
+  // because initialization will not happen in the constructor, but in the `initialize`
+  // method.
+  private _pingUploader!: PingUploader
```
brizental, line 36: nit: Can you move this to the block of `// Properties that will only be set on initialize.` below?

glean/src/platform/test/storage.ts
```diff
@@ -5,25 +5,32 @@
...
+// behaviour of the other persistent storages.
+let globalStore: JSONObject = {};
+
 /**
  * A weak implementation for the Store interface.
  *
  * This means the data saved in this store does not persist throughout application runs.
+ * However, data can be shared across two instances of the store.
  */
 class WeakStore implements Store {
```
brizental, line 19: nit: Can we rename this to MockStore or something else? It is not "weak" anymore :D

brizental (approved):

PR #159, opened 2021-03-30
Use a custom uploader for the encryption test

Dexterp37:
This fixes a couple of small bugs in the configuration object (first commit) and then converts a test to use the uploader instead of spies.


glean/tests/plugins/encryption.spec.ts
```diff
@@ -67,45 +70,76 @@ describe("PingEncryptionPlugin", function() {
...
+    counter.add(37);
+    ping.submit();
 
-    const preEncryptionPayload = JSON.parse(consoleSpy.lastCall.args[0]) as JSONObject;
-    const encryptedPayload = (await Glean.pingsDatabase.getAllPings())[pingId]["payload"]["payload"];
+    const { url, body } = await uploadPromise;
 
     const { plaintext, protectedHeader } = await compactDecrypt(
-      encryptedPayload as string,
+      JSON.parse(body)!["payload"],
```
Dexterp37, line 126: The linter is not being nice here. Any suggestion @brizental ?

glean/src/core/config.ts
```diff
@@ -62,6 +64,7 @@ export class Configuration implements ConfigurationInterface {
     this.serverEndpoint = (config && config.serverEndpoint)
       ? config.serverEndpoint : DEFAULT_TELEMETRY_ENDPOINT;
 
+    this.plugins = config?.plugins;
```
brizental, line 67: It was on purpose that I did not include the plugins in the object but only in the configuration. The Glean singleton doesn't need to be aware of the plugins, each event is aware of the plugins related to it.

glean/tests/plugins/encryption.spec.ts
brizental: `body && body.payload ? payload : ""`?

brizental: Or, something uglier `const payload = body?.payload as string;`

glean/tests/plugins/encryption.spec.ts
```diff
@@ -81,31 +95,49 @@ describe("PingEncryptionPlugin", function() {
...
+      sendInPings: ["encryptedping"],
+      lifetime: Lifetime.Ping,
+      disabled: false
+    });
+
+    counter.add(37);
+    ping.submit();
 
-    const preEncryptionPayload = JSON.parse(consoleSpy.lastCall.args[0]) as JSONObject;
-    const encryptedPayload = (await Glean.pingsDatabase.getAllPings())[pingId]["payload"]["payload"];
```
brizental, line 99: We do this "get pings from database and mock upload" hack [all around](https://github.com/mozilla/glean.js/search?q=Disable+ping+uploading+for+it+not+to+interfere+with+this+tests.). Would you kindly file a bug for us to replace that with a propoer custom uploader?

glean/tests/plugins/encryption.spec.ts
```diff
@@ -67,11 +69,23 @@ describe("PingEncryptionPlugin", function() {
...
-
     const { publicKey, privateKey } = await generateKeyPair("ECDH-ES");
 
+    type UploadSignature = {url: string, body: string, headers?: Record<string, string>};
+    let uploadPromiseResolve: (value: UploadSignature) => void;
+    const uploadPromise = new Promise<UploadSignature>(r => uploadPromiseResolve = r);
+
+    class MockUploader implements Uploader {
+      post(url: string, body: string, headers?: Record<string, string>): Promise<UploadResult> {
+        uploadPromiseResolve({url, body, headers});
```
brizental, line 80: Why does this need to be a promise? Why not:

```ts
// Also, isn't there a clever Typescript way to 
// extract the function args instead of defining `UploadSignature`?
let interceptedArgs: UploadSignature;
...
post(args) { 
   interceptedArgs = args;
   ...
}
```

After writing this example, I realize you could probably go further and do something like:

```ts
const MockUploader: Uploader = {
   post: sinon.stub().returns(Promise.resolve({...}),
}
```

And then leverage sinon to do all the necessary assertions.

brizental (changes requested):

glean/tests/plugins/encryption.spec.ts
```diff
@@ -67,11 +69,23 @@ describe("PingEncryptionPlugin", function() {
...
-    // Disable ping uploading for it not to interfere with this tests.
-    sandbox.stub(Glean["pingUploader"], "triggerUpload").callsFake(() => Promise.resolve());
-
     const { publicKey, privateKey } = await generateKeyPair("ECDH-ES");
 
+    type UploadSignature = {url: string, body: string, headers?: Record<string, string>};
+    let uploadPromiseResolve: (value: UploadSignature) => void;
+    const uploadPromise = new Promise<UploadSignature>(r => uploadPromiseResolve = r);
+
+    class MockUploader implements Uploader {
```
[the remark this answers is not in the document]
Dexterp37: As per our discussion today I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1702206

glean/tests/plugins/encryption.spec.ts
Dexterp37: https://bugzilla.mozilla.org/show_bug.cgi?id=1702208

glean/tests/plugins/encryption.spec.ts
Dexterp37: As discussed, I've added a comment

Dexterp37: And filed https://bugzilla.mozilla.org/show_bug.cgi?id=1702251 :D

PR #165, opened 2021-04-01
Bug 1701578 (Part 1) - Remove some circular dependencies

Dexterp37:
This PR brings down the circular dependencies detected by `npm run lint:circular-deps` from 34 to 20.

These are the low hanging fruits. I'll follow up with other work next.

### DIFF
--- a/glean/src/core/pings/ping_payload.ts
+++ b/glean/src/core/pings/ping_payload.ts
@@ -1,0 +1,42 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { Metrics } from "../metrics/metrics_interface.js";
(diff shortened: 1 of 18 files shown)


glean/src/core/pings/ping_payload.ts
```diff
@@ -0,0 +1,42 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { Metrics } from "../metrics/metrics_interface.js";
```
brizental, line 5: I've seen multiple projects that have a `types/` folder with all the types. Not for this PR, but I wonder if we could benefit form that. WDYT?

brizental (approved):

glean/src/core/pings/ping_payload.ts
Dexterp37: Yeah, it might make sense :)

PR #174, opened 2021-04-06
Bug 1701578 - Remove a circular dependency in the upload module

Dexterp37:
This removes one more circular dependency, but it's quite a bit of work. We're now at 19!

I've got a few more PRs lined up.

### DIFF
--- a/glean/src/core/upload/index.ts
+++ b/glean/src/core/upload/index.ts
@@ -2,9 +2,12 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+import Platform from "../../platform/index.js";
+import { Configuration } from "../config.js";
 import { GLEAN_VERSION } from "../constants.js";
 import { Observer as PingsDatabaseObserver, PingInternalRepresentation } from "../pings/database.js";
-import Glean from "../glean.js";
+import type PingsDatabase from "../pings/database.js";
+import PlatformInfo from "../platform_info.js";
 import Uploader, { UploadResult, UploadResultStatus } from "./uploader.js";
 
 interface QueuedPing extends PingInternalRepresentation {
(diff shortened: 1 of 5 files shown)


glean/src/core/upload/index.ts
```diff
@@ -179,18 +199,16 @@ class PingUploader implements PingsDatabaseObserver {
    *
    * @returns Whether or not to retry the upload attempt.
    */
-  private async processPingUploadResponse(identifier: string, response: UploadResult): Promise<boolean> {
+  private processPingUploadResponse(identifier: string, response: UploadResult): boolean {
     const { status, result } = response;
     if (status && status >= 200 && status < 300) {
       console.info(`Ping ${identifier} succesfully sent ${status}.`);
-      await Glean.pingsDatabase.deletePing(identifier);
```
Dexterp37, line 186: @brizental this is the reason why we need to return ping ids for deletion from the `Observer`: we need to remove this call but we cannot directly inject `pingsDatabase` from the constructor, because creating a `PingsDatabase` requires an uploader (and vice versa!)

glean/tests/core/upload/index.spec.ts
```diff
@@ -66,21 +66,12 @@ describe("PingUploader", function() {
...
-    assert.strictEqual(uploader["queue"].length, 10);
-  });
-
   it("whenever the pings dabase records a new ping, upload is triggered", async function() {
     const spy = sandbox.spy(Glean["pingUploader"], "triggerUpload");
     await fillUpPingsDatabase(10);
     assert.strictEqual(spy.callCount, 10);
   });
-
+/*
```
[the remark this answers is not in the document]
Dexterp37: Because this is still in draft mode and I wanted to make a few tests pass first :-P

glean/src/core/upload/index.ts
Dexterp37: Oh neat, this is much better than what I did!

brizental (approved):

PR #180, opened 2021-04-09
Bug 1701578 - Break the circular dependency between pings/index.ts and pings/maker.ts

Dexterp37:
With this, we're down to ~18~ 17! 🥳 

I also fixed a bug in our linting configuration about semicolon!


glean/src/core/pings/maker.ts
```diff
@@ -225,14 +235,19 @@ export function makePath(identifier: string, ping: PingType): string {
...
  * @param identifier The pings UUID identifier.
  * @param ping The ping to submit.
  * @param reason An optional reason code to include in the ping.
+ * @param debugOptions The debug options.
  *
  * @returns A promise that is resolved once collection and storing is done.
  */
-export async function collectAndStorePing(identifier: string, ping: PingType, reason?: string): Promise<void> {
-  const collectedPayload = await collectPing(ping, reason);
+export async function collectAndStorePing(metricsDatabase: MetricsDatabase, eventsDatabase: EventsDatabase, pingsDatabase: PingsDatabase, applicationId: string, identifier: string, ping: PingType, reason?: string, debugOptions?: DebugOptions): Promise<void> {
```
brizental, line 249: AAAAA

glean/src/core/glean.ts
```diff
@@ -41,9 +42,9 @@ class Glean {
   // The ping uploader. Note that we need to use the definite assignment assertion
   // because initialization will not happen in the constructor, but in the `initialize`
   // method.
-  private _pingUploader!: PingUploader
+  private _pingUploader!: PingUploader;
   // The application ID (will be sanitized during initialization).
-  private _applicationId?: string;
+  private _applicationId!: string;
```
brizental, line 47: Would you mind just changing this in all the "properties set on init"? Follow up is fine too.

glean/src/core/pings/ping_data_interface.ts
```diff
@@ -0,0 +1,17 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
brizental, line 1: nit: Rename this file to `common_ping_data`.

glean/src/core/config.ts
```diff
@@ -6,18 +6,7 @@ import { DEFAULT_TELEMETRY_ENDPOINT, GLEAN_MAX_SOURCE_TAGS } from "./constants.j
...
- */
-interface DebugOptions {
-  // Whether or not lot log pings when they are collected.
-  logPings?: boolean,
-  // The value of the X-Debug-ID header to be included in every ping.
-  debugViewTag?: string,
-  // The value of the X-Source-Tags header to be included in every ping.
-  sourceTags?: string[],
-}
+import { DebugOptions } from "./debug_options.js";
```
brizental, line 9: Why was it necessary to move this into another file?

glean/src/core/pings/maker.ts
```diff
@@ -8,21 +8,25 @@ import DatetimeMetricType, { DatetimeMetric } from "../metrics/types/datetime.js
...
 import TimeUnit from "../metrics/time_unit.js";
 import { ClientInfo, PingInfo, PingPayload } from "../pings/ping_payload.js";
-import PingType from "../pings/index.js";
-import Glean from "../glean.js";
+import CommonPingData from "../pings/ping_data_interface.js";
 import CoreEvents from "../events/index.js";
+import MetricsDatabase from "../metrics/database.js";
+import EventsDatabase from "../metrics/events_database.js";
+import { DebugOptions } from "../debug_options.js";
+import PingsDatabase from "./database.js";
```
brizental, line 16:
```suggestion
import MetricsDatabase from "../metrics/database.js";
import EventsDatabase from "../metrics/events_database.js";
import PingsDatabase from "./database.js";
import { DebugOptions } from "../debug_options.js";
```

uber nit: can we keep the databases together? 😄

glean/src/core/pings/maker.ts
```diff
@@ -225,14 +235,19 @@ export function makePath(identifier: string, ping: PingType): string {
...
  * @param identifier The pings UUID identifier.
  * @param ping The ping to submit.
  * @param reason An optional reason code to include in the ping.
+ * @param debugOptions The debug options.
  *
  * @returns A promise that is resolved once collection and storing is done.
  */
-export async function collectAndStorePing(identifier: string, ping: PingType, reason?: string): Promise<void> {
-  const collectedPayload = await collectPing(ping, reason);
+export async function collectAndStorePing(metricsDatabase: MetricsDatabase, eventsDatabase: EventsDatabase, pingsDatabase: PingsDatabase, applicationId: string, identifier: string, ping: CommonPingData, reason?: string, debugOptions?: DebugOptions): Promise<void> {
```
brizental, line 249: nit: Give a default value to the debug options and we don't need to put in the end of the arguments list.

brizental (approved):
Only nits. Today I am nitpicky. 

Otherwise, LGTM.

glean/src/core/glean.ts
Dexterp37: https://bugzilla.mozilla.org/show_bug.cgi?id=1704083

glean/src/core/config.ts
Dexterp37: Because due to the [other imports in the config.ts file](https://github.com/mozilla/glean.js/blob/3eefee94d66b78b076e722f53c12632aecbd5a61/glean/src/core/config.ts#L5-L8), if we wanted to use `DebugOptions` in the pings module without splitting it, we'd create a new circular dependency :)

This change is why this PR allows us to drop 2 circular dependencies :)

glean/src/core/pings/maker.ts
Dexterp37: https://bugzilla.mozilla.org/show_bug.cgi?id=1704097

PR #184, opened 2021-04-12
Bug 1701578 - Remove more circular dependencies

Dexterp37:
This is a big PR, sorry. However, it's fairly well split across commits and takes down the detected circular dependencies from 17 to ....

6.

We're almost there!


glean/src/core/internal_metrics.ts
```diff
@@ -98,10 +98,10 @@ export class CoreMetrics {
...
     await this.initializeClientId();
     await this.initializeFirstRunDate();
-    await this.initializeOs();
-    await this.initializeOsVersion();
-    await this.initializeArchitecture();
-    await this.initializeLocale();
+    await StringMetricType._private_setUndispatched(this.os, await Glean.platform.info.os());
+    await StringMetricType._private_setUndispatched(this.osVersion, await Glean.platform.info.osVersion());
+    await StringMetricType._private_setUndispatched(this.architecture, await Glean.platform.info.arch());
+    await StringMetricType._private_setUndispatched(this.locale, await Glean.platform.info.locale());
```
brizental: Very nice.

glean/src/core/metrics/events_database.ts
```diff
@@ -4,7 +4,7 @@
 
 import Store from "../storage/index.js";
 import { isUndefined, JSONArray, JSONObject, JSONValue } from "../utils.js";
-import EventMetricType from "./types/event.js";
+import type EventMetricType from "./types/event.js";
```
brizental, line 7: Hm, should we be doing this everywhere? There are multiple moments when we import things that are actually just types. There is a [linter rule](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/consistent-type-imports.md) for this.

brizental (commented):
Hey, great work here. Glad we are almost done with the circular kittens :D

I do not understand the purpose of dcd9ef207ac617618acb5d637c897bb8bd55c162. How does this change make sure `Glean.ts` isn't mistakenly imported?

glean/src/core/metrics/events_database.ts
Dexterp37: Ah! TIL! I think enabling this rule would be great. I haven't taken a big look around to see what else we could fix, we should probably do it. Let me file a bug.

Dexterp37: https://bugzilla.mozilla.org/show_bug.cgi?id=1704505

Dexterp37:
> 
> 
> Hey, great work here. Glad we are almost done with the circular kittens :D
> 
> I do not understand the purpose of [dcd9ef2](https://github.com/mozilla/glean.js/commit/dcd9ef207ac617618acb5d637c897bb8bd55c162). How does this change make sure `Glean.ts` isn't mistakenly imported?

For example, in places like [this one](https://github.com/mozilla/glean.js/blob/de3b09158c1a60ecbb272c3955b539c56271dbb0/glean/src/core/metrics/utils.ts#L9-L13), we import the file but are exclusively interested to the internal metric, not the type. This file is further included in many chains that end up in the Glean.ts include line.

brizental:
> For example, in places like this one, we import the file but are exclusively interested to the internal metric, not the type. This file is further included in many chains that end up in the Glean.ts include line.

So, basically, put as little exports as possible in any files that include a `import Glean from "glean"` statement?

Dexterp37:
> 
> 
> > For example, in places like this one, we import the file but are exclusively interested to the internal metric, not the type. This file is further included in many chains that end up in the Glean.ts include line.
> 
> So, basically, put as little exports as possible in any files that include a `import Glean from "glean"` statement?

By splitting we are actually guaranteeing that the e.g. 'datetime_metric.ts' file, included in many places, does not include Glean.ts at all: so utils.ts won't require Glean.ts at all!

brizental (approved):

glean/src/core/internal_metrics.ts
```diff
@@ -95,24 +97,26 @@ export class CoreMetrics {
...
-    await StringMetricType._private_setUndispatched(this.appDisplayVersion, appDisplayVersion || "Unknown");
+  async initialize(config: ConfigurationInterface, platform: Platform, metricsDatabase: MetricsDatabase): Promise<void> {
+    await this.initializeClientId(metricsDatabase);
+    await this.initializeFirstRunDate(metricsDatabase);
+    await StringMetricType._private_setUndispatched(this.os, await platform.info.os());
+    await StringMetricType._private_setUndispatched(this.osVersion, await platform.info.osVersion());
+    await StringMetricType._private_setUndispatched(this.architecture, await platform.info.arch());
+    await StringMetricType._private_setUndispatched(this.locale, await platform.info.locale());
+    await StringMetricType._private_setUndispatched(this.appBuild, config.appBuild || "Unknown");
+    await StringMetricType._private_setUndispatched(this.appDisplayVersion, config.appDisplayVersion || "Unknown");
```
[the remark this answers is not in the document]
Dexterp37: Done!

PR #198, opened 2021-04-13
Bug 1701578 - Remove the final circular dependencies!

Dexterp37:
As discussed with @brizental , this introduces a new shared state object called `Context` and uses that in the metric types, instead of referring to the Glean object (creating circular dependencies).

It additionally makes the `Dispatcher` a singleton.

Finally, it enabled the circular dependencies checker on CI.

### DIFF
--- a/glean/tests/core/pings/index.spec.ts
+++ b/glean/tests/core/pings/index.spec.ts
@@ -9,6 +9,7 @@ import PingType from "../../../src/core/pings";
 import CounterMetricType from "../../../src/core/metrics/types/counter";
 import { Lifetime } from "../../../src/core/metrics/lifetime";
 import Glean from "../../../src/core/glean";
+import Dispatcher from "../../../src/core/dispatcher";
 
 const sandbox = sinon.createSandbox();
 
@@ -20,7 +21,7 @@ const sandbox = sinon.createSandbox();
 async function submitSync(ping: PingType): Promise<void> {
   ping.submit();
   // TODO: Drop this whole approach once Bug 1691033 is resolved.
-  await Glean.dispatcher.testBlockOnQueue();
+  await Dispatcher.instance.testBlockOnQueue();
 }
 
 describe("PingType", function() {
--- a/glean/tests/core/pings/index.spec.ts
+++ b/glean/tests/core/pings/index.spec.ts
@@ -10,6 +10,7 @@ import CounterMetricType from "../../../src/core/metrics/types/counter";
 import { Lifetime } from "../../../src/core/metrics/lifetime";
 import Glean from "../../../src/core/glean";
 import Dispatcher from "../../../src/core/dispatcher";
+import { Context } from "../../../src/core/context";
 
 const sandbox = sinon.createSandbox();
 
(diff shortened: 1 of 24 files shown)


glean/tests/core/pings/maker.spec.ts
```diff
@@ -135,7 +136,7 @@ describe("PingMaker", function() {
 
     Glean.unsetDebugViewTag();
     Glean.unsetSourceTags();
-    await Glean.dispatcher.testBlockOnQueue();
+    await Dispatcher.instance.testBlockOnQueue();
```
brizental, line 140: Why not put the dispatcher in the Context object too?

glean/tests/core/pings/index.spec.ts
```diff
@@ -74,11 +75,11 @@ describe("PingType", function() {
     });
 
     await submitSync(ping1);
-    let storedPings = await Glean.pingsDatabase["store"]._getWholeStore();
+    let storedPings = await Context.instance.pingsDatabase["store"]._getWholeStore();
     assert.strictEqual(Object.keys(storedPings).length, 0);
 
     await submitSync(ping2);
-    storedPings = await Glean.pingsDatabase["store"]._getWholeStore();
+    storedPings = await Context.instance.pingsDatabase["store"]._getWholeStore();
```
brizental, line 82: Can we have getters that allow for skipping the `.instance` part? Like we do on the Glean singleton?

glean/tests/core/upload/index.spec.ts
```diff
@@ -32,7 +32,7 @@ async function fillUpPingsDatabase(numPings: number): Promise<string[]> {
 
   const identifiers = Array.from({ length: numPings }, () => UUIDv4());
   for (const identifier of identifiers) {
-    await collectAndStorePing(Context.instance.metricsDatabase, Context.instance.eventsDatabase, Context.instance.pingsDatabase, Glean.applicationId, identifier, ping);
+    await collectAndStorePing(Context.instance.metricsDatabase, Context.instance.eventsDatabase, Context.instance.pingsDatabase, Context.instance.applicationId, identifier, ping);
```
brizental, line 35: Should we just pass `Context` where we need to do any dependency injection?

brizental (approved):
r+wc. 

All my comments can be dealt with as follow ups, the only one I'd prefer to deal with before merging is the one about the dispatcher, I think it should not be a singleton by itself, but be in the Context. Unless I am missing something.

We should also file a bug for documenting this architecture and why the Context is important, and if we haven't yet a bug about removing the default export from Glean.ts and exporting it as something like `export const DO_NOT_IMPORT_ME_UNLESS_YOU_KNOW_WHAT_YOU_ARE_DOING = Glean` :D

Finally, did you check that this fixes the original bug about import order of Glean singleton and Metrics?

glean/tests/plugins/encryption.spec.ts
```diff
@@ -59,11 +59,11 @@ describe("PingEncryptionPlugin", function() {
     const pingId = "ident";
 
     const postSpy = sandbox.spy(TestPlatform.uploader, "post").withArgs(
-      sinon.match(makePath(Context.instance.applicationId, pingId, ping)),
+      sinon.match(makePath(Context.applicationId, pingId, ping)),
       sinon.match.string
     );
 
-    await collectAndStorePing(Context.instance.metricsDatabase, Context.instance.eventsDatabase, Context.instance.pingsDatabase, Context.instance.applicationId, pingId, ping, undefined, Context.instance.debugOptions);
+    await collectAndStorePing(pingId, ping);
```
brizental, line 66: ✨ self reminder to close [this](https://bugzilla.mozilla.org/show_bug.cgi?id=1704097) once we merge this PR ✨

brizental (approved):

PR #201, opened 2021-04-13
Bug 1701652 - Don't let the platform be changed after Glean was initialized

brizental:


CHANGELOG.md
```diff
@@ -2,15 +2,15 @@
 
 [Full changelog](https://github.com/mozilla/glean.js/compare/v0.8.0...main)
 
+* [#201](https://github.com/mozilla/glean.js/pull/201): BUGFIX: Do not let the platform be changed after Glean is initialized.
```
Dexterp37, line 5:
```suggestion
* [#201](https://github.com/mozilla/glean.js/pull/201): BUGFIX: Do not let the platform be changed after Glean is initialized.

```

glean/src/core/glean.ts
```diff
@@ -443,6 +443,14 @@ class Glean {
    *        Please check out the available environments in the platform/ module.
    */
   static setPlatform(platform: Platform): void {
+    if (Context.initialized) {
```
Dexterp37, line 446: This seems very verbose and log-cluttering. Can we make it debug or only print this if it's already initialized and the new platform is different than the existing one?

Dexterp37 (changes requested):
I have one additional concern: by doing this, if Glean is initialized in the code being tested (e.g. Rally inits Glean), then the tests might send pings because the uploader is no longer the one from the test platform. To cover these cases, and avoid the potential of submitting pings from tests, can we make `testResetGlean` use a mock default uploader is none is passed via the configuration?

brizental:
> I have one additional concern: by doing this, if Glean is initialized in the code being tested (e.g. Rally inits Glean), then the tests might send pings because the uploader is no longer the one from the test platform. 

Note that `testResetGlean` always [uninitializes before initializing](https://github.com/mozilla/glean.js/blob/main/glean/src/core/glean.ts#L518). Therefore even if Glean is initialized in the code being tested before `testResetGlean` is called, the platform and config is still overwritten.

This is why I decided to rely on `initialized` to set the platform or not, instead of just checking if there is a platform there already. Writing this out gives me the vibe that it is fragile, but such fragility would be mitigated by work on [Bug 1682771](https://bugzilla.mozilla.org/show_bug.cgi?id=1682771) and a clear testing state for Glean.

glean/src/core/glean.ts
brizental: That would still make it be logged everytime Glean.initialize is called in tests code. The platform will be the TestPlatform and the code will be trying to change to WebExt or something else. Making it a debug log will still make it be logged... I can just remove it :) I agree it will be an annyoing log, I put it because I thought it was causing your issue over in Rally and it would help us debug that. Since it isn't I think we can live without it.

brizental: Actually, maybe I can just add a log in case we are switching from a platform to another. Wait for my next commit.

Dexterp37: > 
> 
>  ...case we are switching from a platform to another. Wait for my next commit.

Yes, that's what I was suggesting :)

glean/src/core/glean.ts
```diff
@@ -443,6 +443,27 @@ class Glean {
...
+    // uninitializes Glean before setting the testing platform.
+    if (Context.initialized) {
+      return;
+    }
+
+    // In case the platform is being swapped in the test scenario described above,
+    // we log a debug message about the change.
+    if (Glean.instance._platform && Glean.instance._platform.name !== platform.name) {
+      console.debug(
+        `Changing Glean platform from "${Glean.platform.name}" to "${platform.name}".`
```
Dexterp37, line 463: nit: should we mention that this should only happen in tests in the message?

Dexterp37 (approved):

glean/src/core/glean.ts
brizental: Hm, I tried it but it looks weird. What if I instead add a comment:

```js
// TODO: Only show this message outside of test mode, and rephrase it as an error (depends on Bug 1682771).
```

PR #202, opened 2021-04-14
Bug 1701889 - Implement the ping testing api

Dexterp37:
This implements the pings testing API, as discussed and designed in the related bug.

### DIFF
--- a/glean/src/core/pings/ping_type.ts
+++ b/glean/src/core/pings/ping_type.ts
@@ -27,7 +27,7 @@ class PingType implements CommonPingData {
   // execute and synchronize with the submission API.
   private resolveTestPromiseFunction?: PromiseCallback;
   private rejectTestPromiseFunction?: PromiseCallback;
-  private testValidator?: ValidatorFunction;
+  private testCallback?: ValidatorFunction;
 
   constructor (meta: CommonPingData) {
     this.name = meta.name;
@@ -59,69 +59,51 @@ class PingType implements CommonPingData {
     // other dispatched tasks. Unfortunately, this causes a deadlock.
     // In order to work around that problem, we kick off validation
     // right before the actual submission takes place, through another
-    // async function (and not through the dispatcher). We then await
-    // in the dispatched submission task on the related promise, which
-    // will always resolve due to the use of `finally`.
-    if (this.testValidator) {
-      const cleanup = () => {
-        this.resolveTestPromiseFunction = undefined;
-        this.rejectTestPromiseFunction = undefined;
-        this.testValidator = undefined;
-      };
-
-      this.testValidator(reason)
+    // async function (and not through the dispatcher). After validation
+    // is complete, regardless of the outcome, the ping submission is
+    // finally triggered.
+    if (this.testCallback) {
+      this.testCallback(reason)
         .then(() => {
-          // Temporarily store the function and then clean up, so that fast consecutive
-          // calls to `testBeforeNextSubmit` don't fail because things are still set
-          // after the calling promise is resolved.
-          const resolver = this.resolveTestPromiseFunction;
-          cleanup();
-
-          this.internalSubmit(reason, resolver);
+          PingType._private_internalSubmit(this, reason, this.resolveTestPromiseFunction);
         })
         .catch(e => {
           console.error(`There was an error validating "${this.name}" (${reason ?? "no reason"}):`, e);
-
-          // Temporarily store the function and then clean up, so that fast consecutive
-          // calls to `testBeforeNextSubmit` don't fail because things are still set
-          // after the calling promise is resolved.
-          const rejecter = this.rejectTestPromiseFunction;
-          cleanup();
-
-          this.internalSubmit(reason, rejecter);
+          PingType._private_internalSubmit(this, reason, this.rejectTestPromiseFunction);
         });
     } else {
-      this.internalSubmit(reason);
+      PingType._private_internalSubmit(this, reason);
     }
   }
 
-  private internalSubmit(reason?: string, testResolver?: PromiseCallback): void {
+  private static _private_internalSubmit(instance: PingType, reason?: string, testResolver?: PromiseCallback): void {
     Context.dispatcher.launch(async () => {
       if (!Context.initialized) {
         console.info("Glean must be initialized before submitting pings.");
         return;
       }
 
-      if (!Context.uploadEnabled && !this.isDeletionRequest()) {
+      if (!Context.uploadEnabled && !instance.isDeletionRequest()) {
         console.info("Glean disabled: not submitting pings. Glean may still submit the deletion-request ping.");
         return;
       }
 
       let correctedReason = reason;
-      if (reason && !this.reasonCodes.includes(reason)) {
+      if (reason && !instance.reasonCodes.includes(reason)) {
         console.error(`Invalid reason code ${reason} from ${this.name}. Ignoring.`);
         correctedReason = undefined;
       }
 
       const identifier = generateUUIDv4();
-      await collectAndStorePing(identifier, this, correctedReason);
+      await collectAndStorePing(identifier, instance, correctedReason);
 
-      // This guarantees that, when running tests, the promise returned by
-      // `testBeforeNextSubmit` is resolved after the ping is collected: this is
-      // needed to make sure calling the testing APIs on metrics behave consistently
-      // if tests run fast.
       if (testResolver) {
         testResolver();
+
+        // Finally clean up!
+        instance.resolveTestPromiseFunction = undefined;
+        instance.rejectTestPromiseFunction = undefined;
+        instance.testCallback = undefined;
       }
     });
   }
@@ -133,14 +115,14 @@ class PingType implements CommonPingData {
    *
    * TODO: Only allow this function to be called on test mode (depends on Bug 1682771).
    *
-   * @param validatorFn The asynchronous validation function to run in order to validate
-   *                    the ping content.
+   * @param callbackFn The asynchronous validation function to run in order to validate
+   *        the ping content.
    *
    * @returns A `Promise` resolved when the ping is collected and the validation function
    *          is executed.
    */
-  async testBeforeNextSubmit(validatorFn: ValidatorFunction): Promise<void> {
-    if (this.testValidator) {
+  async testBeforeNextSubmit(callbackFn: ValidatorFunction): Promise<void> {
+    if (this.testCallback) {
       console.error(`There is an existing test call for ping "${this.name}". Ignoring.`);
       return;
     }
@@ -148,7 +130,7 @@ class PingType implements CommonPingData {
     return new Promise((resolve, reject) => {
       this.resolveTestPromiseFunction = resolve;
       this.rejectTestPromiseFunction = reject;
-      this.testValidator = validatorFn;
+      this.testCallback = callbackFn;
     });
   }
 }
--- a/glean/src/core/pings/ping_type.ts
+++ b/glean/src/core/pings/ping_type.ts
@@ -8,6 +8,9 @@ import collectAndStorePing from "../pings/maker.js";
 import type CommonPingData from "./common_ping_data.js";
 import { Context } from "../context.js";
 
+type ValidatorFunction = (reason?: string) => Promise<void>;
+type PromiseCallback = (value: void | PromiseLike<void>) => void;
+
 /**
  * Stores information about a ping.
  *
@@ -19,6 +22,12 @@ class PingType implements CommonPingData {
   readonly includeClientId: boolean;
   readonly sendIfEmpty: boolean;
   readonly reasonCodes: string[];
+
+  // The functions and promises required for the test API to
+  // execute and synchronize with the submission API.
+  private resolveTestPromiseFunction?: PromiseCallback;
+  private rejectTestPromiseFunction?: PromiseCallback;
+  private testValidator?: ValidatorFunction;
 
   constructor (meta: CommonPingData) {
     this.name = meta.name;
@@ -44,6 +53,49 @@ class PingType implements CommonPingData {
    *               `ping_info.reason` part of the payload.
    */
   submit(reason?: string): void {
+    // **** Read this before changing the following code! ****
+    //
+    // The Dispatcher does not allow dispatched tasks to await on
+    // other dispatched tasks. Unfortunately, this causes a deadlock.
+    // In order to work around that problem, we kick off validation
+    // right before the actual submission takes place, through another
+    // async function (and not through the dispatcher). We then await
+    // in the dispatched submission task on the related promise, which
+    // will always resolve due to the use of `finally`.
+    if (this.testValidator) {
+      const cleanup = () => {
+        this.resolveTestPromiseFunction = undefined;
+        this.rejectTestPromiseFunction = undefined;
+        this.testValidator = undefined;
+      };
+
+      this.testValidator(reason)
+        .then(() => {
+          // Temporarily store the function and then clean up, so that fast consecutive
+          // calls to `testBeforeNextSubmit` don't fail because things are still set
+          // after the calling promise is resolved.
+          const resolver = this.resolveTestPromiseFunction;
+          cleanup();
+
+          this.internalSubmit(reason, resolver);
+        })
+        .catch(e => {
+          console.error(`There was an error validating "${this.name}" (${reason ?? "no reason"}):`, e);
+
+          // Temporarily store the function and then clean up, so that fast consecutive
+          // calls to `testBeforeNextSubmit` don't fail because things are still set
+          // after the calling promise is resolved.
+          const rejecter = this.rejectTestPromiseFunction;
+          cleanup();
+
+          this.internalSubmit(reason, rejecter);
+        });
+    } else {
+      this.internalSubmit(reason);
+    }
+  }
+
+  private internalSubmit(reason?: string, testResolver?: PromiseCallback): void {
     Context.dispatcher.launch(async () => {
       if (!Context.initialized) {
         console.info("Glean must be initialized before submitting pings.");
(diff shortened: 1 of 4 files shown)


glean/src/core/pings/index.ts
```diff
@@ -60,12 +98,34 @@ class PingType implements CommonPingData {
...
+   *
+   * TODO: Only allow this function to be called on test mode (depends on Bug 1682771).
+   *
+   * @param validatorFn The asynchronous validation function to run in order to validate
+   *                    the ping content.
+   *
+   * @returns A `Promise` resolved when the ping is collected and the validation function
+   *          is executed.
+   */
+  async testBeforeNextSubmit(validatorFn: ValidatorFunction): Promise<void> {
```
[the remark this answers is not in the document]
brizental: Shouldn't it instead return `testValidatorPromise`?

brizental (changes requested):
I am ok with the overall approach. I don't see any other way to implement this feature either. I'll move Bug 1682771 to P2 though. Disallowing test APIs outside of tests seems even more important with this feature.

glean/src/core/pings/index.ts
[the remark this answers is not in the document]
brizental: Argh, I get why it returns this promise now.

glean/src/core/pings/index.ts
```diff
@@ -20,11 +22,22 @@ class PingType implements CommonPingData {
   readonly sendIfEmpty: boolean;
   readonly reasonCodes: string[];
 
+  // The functions and promises required for the test API to
+  // execute and synchronize with the submission API.
+  private testResolutionFunction?: (value: void | PromiseLike<void>) => void;
```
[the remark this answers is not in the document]
Dexterp37: This is used to resolve the promise returned by `testBeforeNextSubmit`. Without this, we'd have no way to resolve it.

glean/src/core/pings/index.ts
```diff
@@ -20,11 +22,22 @@ class PingType implements CommonPingData {
...
+  private testValidator?: ValidatorFunction;
+  private testValidatorPromise: Promise<void>;
```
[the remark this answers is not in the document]
Dexterp37: Mh, what's wrong with the use of validator here? I wanted to make sure the intent was clear. Callback, IMHO, might be a bit misleading: it may give the impression that callback could be used to wait on the submission, while it's really useful exclusively for validating.

glean/src/core/pings/index.ts
```diff
@@ -44,6 +53,49 @@ class PingType implements CommonPingData {
...
   submit(reason?: string): void {
+    // **** Read this before changing the following code! ****
+    //
+    // The Dispatcher does not allow dispatched tasks to await on
+    // other dispatched tasks. Unfortunately, this causes a deadlock.
+    // In order to work around that problem, we kick off validation
+    // right before the actual submission takes place, through another
+    // async function (and not through the dispatcher). We then await
+    // in the dispatched submission task on the related promise, which
+    // will always resolve due to the use of `finally`.
```
Dexterp37, line 64: Note to self: update the comment to remove the mention to finally.

glean/src/core/pings/index.ts
```diff
@@ -44,6 +53,49 @@ class PingType implements CommonPingData {
...
+        .catch(e => {
+          console.error(`There was an error validating "${this.name}" (${reason ?? "no reason"}):`, e);
+
+          // Temporarily store the function and then clean up, so that fast consecutive
+          // calls to `testBeforeNextSubmit` don't fail because things are still set
+          // after the calling promise is resolved.
+          const rejecter = this.rejectTestPromiseFunction;
+          cleanup();
+
+          this.internalSubmit(reason, rejecter);
```
brizental, line 91: Do we want to submit even if the validator got an error? This is testing code, it is better to raise an exception if a test error happened, otherwise output is misleading i.e. it looks like the tests passed when there are actually error logs around (encountered this confusing scenario many times yesterday when testing this PR).

glean/src/core/pings/index.ts
```diff
@@ -60,10 +112,43 @@ class PingType implements CommonPingData {
...
+   *
+   * @param validatorFn The asynchronous validation function to run in order to validate
+   *                    the ping content.
+   *
+   * @returns A `Promise` resolved when the ping is collected and the validation function
+   *          is executed.
+   */
+  async testBeforeNextSubmit(validatorFn: ValidatorFunction): Promise<void> {
+    if (this.testValidator) {
+      console.error(`There is an existing test call for ping "${this.name}": ignoring.`);
```
brizental, line 144:
```suggestion
      console.error(`There is an existing test call for ping "${this.name}". Ignoring.`);
```

Just to keep consistency with other messages around the code.

glean/tests/core/pings/index.spec.ts
```diff
@@ -107,4 +107,175 @@ describe("PingType", function() {
...
+    const p = ping.testBeforeNextSubmit(async () => {
+      throw new Error("This should reject!");
+    });
+    
+    ping.submit();
+
+    await assert.rejects(p);
+  });
+
+  it("runs a validator: sorry", async function() {
```
brizental, line 239: LOL

glean/tests/core/pings/index.spec.ts
```diff
@@ -107,4 +107,175 @@ describe("PingType", function() {
...
+
+    let validatorRun = false;
+
+    const p = ping.testBeforeNextSubmit(async () => {
+      await new Promise<void>(resolve => {
+        setTimeout(() => resolve(), 100);
+      });
+      const value = await counter.testGetValue();
+      console.log("!!!", value);
+      assert.strictEqual(value, 100);
```
brizental, line 263: Note: the fact that this assertion passes is the issue. Not sure it did pass, because we currently raise an error when Validators reject, but if it did, it shouldn't.

We record to the counter _after_ the next submit, so `value` should be `undefined` when this assertion is executed. From my tests yesterday I did not reach a conclusion as to whether that was indeed the case, I'll do more testing after this review.

From how I understand this code works, there is no way to guarantee this will be executed before the `counter.add` calls below, unless we `await p` before going into the `counter.add` loop.

brizental (changes requested):
Will continue on the tests I was doing yesterday, my concerns still remain. I explained what are these concerns on the comment I left on the tests code.

glean/src/core/pings/index.ts
Dexterp37: Yes, I think we should submit the ping regardless. I'd rather not change the code too much, we should keep this as close as production as possible, with the only exception that some test code runs before the submission.

The code already rejects/throws (exceptions will cause rejections anyway): see the related test.

glean/tests/core/pings/index.spec.ts
Dexterp37: > 
> 
> Note: the fact that this assertion passes is the issue. Not sure it did pass, because we currently don't raise an error when Validators reject, but if it did, it shouldn't.

Then the test is probably wrong, because if the validator throws, the promise is rejected. See `"runs a validator that rejects"`.

> From how I understand this code works, there is no way to guarantee this will be executed before the `counter.add` calls below, unless we `await p` before going into the `counter.add` loop.

Yes, exactly.

Dexterp37:
Ad discussed with @brizental , sadly we have no way to guarantee that metric calls executed after `testBeforeNextSubmit` get executed before the ping collection. This is due to the fact that the Glean.js dispatcher is asynchronous even in test mode (it's not in the other bindings). One way to make this happen would be to change the dispatcher to add "task dependencies", e.g. if task A has a dependency B, run B _right before_ running A.

Sadly, that's a bigger change. I'm closing this PR.

Dexterp37:
Let's give this another opportunity.

glean/src/core/dispatcher.ts
```diff
@@ -116,12 +118,51 @@ class Dispatcher {
         await this.executeTask(nextCommand.task);
         nextCommand.resolver();
         nextCommand = this.getNextCommand();
+        continue;
+      case(Commands.TestUnnestTask):
+        let nestedCommands: {task: Task, resolver: () => void}[] = [];
+
+        console.log("**** DEBUG - Running test task");
+        const originalFunc = this.testLaunch;
```
brizental, line 126: There is an assumption here that only `testLaunch` will be called inside a `TestUnnestTask`, but there are no guarantees around it. I don't think it is unlikely that `launch` calls will also be inside `testBeforeNextSubmit`, so they should be monkeypatched too?

Did you consider monkeypatching Glean's whole dispatcher instead and then waiting for the temp dispatcher's queue to be emptied before leaving this block?

brizental: Relying on global state is what bothers me here though. If we have this scenario, would this break?

```js
const p = ping.testBeforeNextSubmit(async () => {
     assert.strictEqual(await counter.testGetValue(), 10);
});
ping.submit();
// Here testGetValue could be the monkey patched value, no?
assert.ok(await someOtherMetric.testGetValue());
await p;
```

glean/src/core/pings/ping_type.ts
```diff
@@ -44,6 +53,49 @@ class PingType implements CommonPingData {
...
+          cleanup();
+
+          this.internalSubmit(reason, rejecter);
+        });
+    } else {
+      this.internalSubmit(reason);
+    }
+  }
+
+  private internalSubmit(reason?: string, testResolver?: PromiseCallback): void {
```
brizental, line 98: The `private` keywork is ignored by Javascript and even in Typescript it's very easy to bypass. Let's make this a static method l[ike on the MetricTypes that have an undispatched method](https://github.com/mozilla/glean.js/blob/main/glean/src/core/metrics/types/counter.ts#L35). So it doesn't show up on code completion for folks.

glean/src/core/pings/ping_type.ts
```diff
@@ -44,6 +53,49 @@ class PingType implements CommonPingData {
...
   submit(reason?: string): void {
+    // **** Read this before changing the following code! ****
+    //
+    // The Dispatcher does not allow dispatched tasks to await on
+    // other dispatched tasks. Unfortunately, this causes a deadlock.
+    // In order to work around that problem, we kick off validation
+    // right before the actual submission takes place, through another
+    // async function (and not through the dispatcher). We then await
+    // in the dispatched submission task on the related promise, which
+    // will always resolve due to the use of `finally`.
```
brizental, line 64: This comment is outdated.

glean/src/core/pings/ping_type.ts
```diff
@@ -44,6 +53,49 @@ class PingType implements CommonPingData {
...
+        this.resolveTestPromiseFunction = undefined;
+        this.rejectTestPromiseFunction = undefined;
+        this.testValidator = undefined;
+      };
+
+      this.testValidator(reason)
+        .then(() => {
+          // Temporarily store the function and then clean up, so that fast consecutive
+          // calls to `testBeforeNextSubmit` don't fail because things are still set
+          // after the calling promise is resolved.
```
brizental, line 76: This comment is a bit confusing, IMHO. The code itself is clear enough in this specific part I would say we can drop the comment.

glean/src/core/pings/ping_type.ts
```diff
@@ -63,7 +115,40 @@ class PingType implements CommonPingData {
 
       const identifier = generateUUIDv4();
       await collectAndStorePing(identifier, this, correctedReason);
+
+      // This guarantees that, when running tests, the promise returned by
+      // `testBeforeNextSubmit` is resolved after the ping is collected: this is
+      // needed to make sure calling the testing APIs on metrics behave consistently
+      // if tests run fast.
```
brizental, line 122: Again, I think this comment is a bit confusing and the code speaks for itself.

glean/tests/core/pings/index.spec.ts
```diff
@@ -107,4 +107,175 @@ describe("PingType", function() {
...
+    let validatorRun = false;
+    const p = ping.testBeforeNextSubmit(r => {
+      assert.strictEqual(r, "test");
+      validatorRun = true;
+      return Promise.resolve();
+    });
+
+    // Internals should be defined after the API was called.
+    assert.notStrictEqual(ping["resolveTestPromiseFunction"], undefined);
+    assert.notStrictEqual(ping["testValidator"], undefined);
```
brizental, line 132: There is no guarantee that these values were not cleaned up before we got here. This _could_ create intermittents. No strong opinions on removing it or not, because most of the time it _should_ pass. If we leave it we should also check for `rejectTestPromiseFunction`, for consistency. Same for `undefined` checks above.

glean/src/core/pings/ping_type.ts
brizental: Okay, looking at the tests code, if the API is used correctly i.e. aways `await` on the promise returned by `testBeforeNextSubmit` before doing anything else, people should never get the sort of issue this behaviour is trying to solve... We could clean up inside `internalSubmit` and remove some extra lines from this file.

glean/tests/core/pings/index.spec.ts
```diff
@@ -107,4 +107,175 @@ describe("PingType", function() {
...
+    });
+
+    assert.strictEqual(ping["testValidator"], undefined);
+
+    const testFunction = async () => Promise.resolve();
+    void ping.testBeforeNextSubmit(testFunction);
+    assert.strictEqual(ping["testValidator"], testFunction);
+
+    void ping.testBeforeNextSubmit(() => Promise.resolve());
+    assert.strictEqual(ping["testValidator"], testFunction);
```
brizental, line 219: The behaviour here is not guaranteed. If somehow everything happens super fast, this could cause intermittents. If you do the `cleanup` inside `internalSubmit`, you could stop the dispatcher for this test. With the dispatcher stopped `cleanup` is guaranteed not to happen and this test is more predictable.

glean/tests/core/pings/index.spec.ts
```diff
@@ -253,29 +256,79 @@ describe("PingType", function() {
     });
 
     let validatorRun = false;
+    const TEST_NUM_ADDITIONS = 100;
 
     const p = ping.testBeforeNextSubmit(async () => {
-      await new Promise<void>(resolve => {
-        setTimeout(() => resolve(), 100);
-      });
+      await new Promise(r => setTimeout(r, 100));
```
brizental, line 262: Maybe leave a little comment stating why the timeout is necessary here. Without the timeout what happens is even more unpredictable.

glean/tests/core/pings/index.spec.ts
```diff
@@ -253,29 +256,79 @@ describe("PingType", function() {
...
+
       const value = await counter.testGetValue();
-      console.log("!!!", value);
-      assert.strictEqual(value, 100);
+      assert.strictEqual(value, undefined);
```
brizental, line 265: Also, a small comment stating why the expected value is `undefined`.

glean/tests/core/pings/index.spec.ts
```diff
@@ -253,29 +256,79 @@ describe("PingType", function() {
...
+      disabled: false
+    });
+
+    let validatorRun = false;
+    const TEST_NUM_ADDITIONS = 100;
+
+    const p = ping.testBeforeNextSubmit(async () => {
+      await new Promise(r => setTimeout(r, 100));
+
+      assert.strictEqual(await counter.testGetValue(), 37, "Canary must match");
```
brizental, line 313: `"Canary must match"` but the test is for `counter`.

glean/tests/core/pings/index.spec.ts
```diff
@@ -253,29 +256,79 @@ describe("PingType", function() {
...
+
+    const counter = new CounterMetricType({
+      category: "aCategory",
+      name: "aCounterMetric",
+      sendInPings: ["custom"],
+      lifetime: Lifetime.Ping,
+      disabled: false
+    });
+
+    const canary = new CounterMetricType({
```
brizental, line 299: Can we give these two counter more distinctive names? I was bit by the similarities when I was meddling with this test.

glean/src/core/pings/ping_type.ts
Dexterp37: I retained the keyword and additionally used static as you suggested.

glean/src/core/pings/ping_type.ts
Dexterp37: Removing this caused intermittents :(

glean/tests/core/pings/index.spec.ts
Dexterp37: Good catch for the missing asserts, I added them.

With respect to the interkittenity of this, I'm not entirely sure. Our [docs report](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#description):

> The body of an async function can be thought of as being split by zero or more await expressions. Top-level code, up to and including the first await expression (if there is one), is run synchronously. In this way, an async function without an await expression will run synchronously. If there is an await expression inside the function body, however, the async function will always complete asynchronously.

Since `testBeforeNextSubmit` has no `await`, it should be executed synchronously. The same reasoning should apply to the executor part of the Promise, since it does not contain any await or asynchronous operation.

Does it make sense? Did you see this intermittently failing locally?

glean/src/core/pings/ping_type.ts
brizental: Hm, that means I understand _something_ wrong from this code. Would you mind sharing what failed?

glean/tests/core/pings/index.spec.ts
brizental: Interesting, TIL! I assumed because the whole body of the `async` function is wrapped inside a promise it was not guaranteed to be run sync unless `await`ed on.

Either way my concerns have no ground, I was wrong :) You make these checks before calling `ping.submit` so this is guaranteed to be true. Sorry for the false interkitten alarm.

glean/tests/core/pings/index.spec.ts
Dexterp37: I think the same rationale of my previous comment should apply here, too.

brizental: And once again I am wrong here. `testBeforeNextSubmit` does not trigger submission!

glean/src/core/pings/ping_type.ts
brizental: Still think this comment is not necessary... But also wrong now? Specifically

```
// [...] this is
// needed to make sure calling the testing APIs on metrics behave consistently
// if tests run fast.
```

glean/src/core/pings/ping_type.ts
```diff
@@ -44,26 +53,88 @@ class PingType implements CommonPingData {
...
+
+  /**
+   * **Test-only API**
+   *
+   * Runs a validation function before the ping is collected.
+   *
+   * TODO: Only allow this function to be called on test mode (depends on Bug 1682771).
+   *
+   * @param callbackFn The asynchronous validation function to run in order to validate
+   *                   the ping content.
```
brizental, line 123: nit: Align the second line with `callbackFn`.

glean/tests/core/pings/index.spec.ts
```diff
@@ -107,4 +107,233 @@ describe("PingType", function() {
     const storedPings = await Context.pingsDatabase["store"]._getWholeStore();
```
brizental, line 107: nit: can you rename this file to `ping_type.spec.ts` since we are here?

brizental (approved):
r+wc

PR #209, opened 2021-04-14
Enable the no trailing spaces eslint rule

Dexterp37:
I'm not sure why we didn't have this :)


brizental (approved):
Oversight probably, thanks for adding this.

PR #215, opened 2021-04-17
Update glean-parser to version 3.1.0

Dexterp37:


glean/src/cli.ts
```diff
@@ -18,7 +18,7 @@ const execAsync = promisify(exec.exec);
 const VIRTUAL_ENVIRONMENT_DIR = ".venv";
 
 // The version of glean_parser to install from PyPI.
-const GLEAN_PARSER_VERSION = "2.5.0";
+const GLEAN_PARSER_VERSION = "3.1.0";
```
brizental, line 21: We should include this as a step in https://mozilla.github.io/glean_parser/contributing.html#deploying

brizental (approved):

glean/src/cli.ts
Dexterp37: Ah good point. I'll take that as a follow up

Dexterp37: https://bugzilla.mozilla.org/show_bug.cgi?id=1706043

PR #226, opened 2021-04-20
Bug 1706255 - Fix Qt sample and Qt CI check

brizental:
The CI check did not account for errors other than Glean errors i.e. import errors.

Opening this as a draft to check that CI now does fail (as the sample is broke, it cannot import Glean at the moment).


brizental:
Ok this is ready for review, it fixes with the Qt sample and ~part of the issues with~ the CI check.

"Oh, but the check is red!" -- Indeed, but the check is red because there is _another issue_ on the CI check that is failing to install Qt properly. You will need to take my word that the sample is working, but to give a bit more confidence, here is a video of it working:

https://user-images.githubusercontent.com/25176023/115385285-41919580-a1d8-11eb-8a40-e6d1c98a0d4e.mov

I would like to get this merged before I work on the dependencies issue in CI check, because I would not like to block the other fixes (which are more important) on it.

glean/package.json
```diff
@@ -90,7 +90,7 @@
     "ts-node": "^9.1.1",
     "typescript": "^4.1.5",
     "web-ext-types": "^3.2.1",
-    "webpack": "^5.21.2",
+    "webpack": "^5.34.0",
```
brizental, line 93: This happened when I ran `npm install` locally. I just saw it, but since it didn't break anything, would you mind leaving it in the PR?

bin/qt-js-check.sh
```diff
@@ -13,14 +13,15 @@ source venv/bin/activate
 pip install -r requirements.txt
 glean_parser translate metrics.yaml pings.yaml -f javascript -o generated
 
+export QT_DEBUG_PLUGINS=1
```
Dexterp37, line 16: Would you kindly add a doc line to explain why this is important? (e.g. more info about plugin loads)

bin/qt-js-check.sh
```diff
@@ -13,14 +13,15 @@ source venv/bin/activate
...
 sudo apt-get install xvfb
 sudo apt-get install libxcb-icccm4 libxcb-image0 libdbus-1-3 libxcb-xfixes0 libxkbcommon-x11-0 libxcb-xkb1 libxcb-xinerama0 libxcb-keysyms1 libxcb-randr0 libxcb-shape0 libxrender1 libxcb-render-util0
 xvfb-run python main.py &> qml.log &
 
 sleep 10
 
-if grep -q "Error executing task:" "qml.log"; then
-  echo "Failed to initialize Glean in QML! See more logs below."
+if ! grep -q "Initialized Glean succesfully." "qml.log"; then
+  echo "\n\n\033[1;91m** Failed to initialize Glean in Qt! See more logs below. **\033[0m\n\n"
```
Dexterp37, line 24: Woha. Shell coloring.

glean/package.json
Dexterp37: Just make sure Rally and Bergamot still work before merging, please.

samples/qt-qml-app/pings.js
```diff
@@ -0,0 +1,21 @@
...
+"use strict";
+
+/**
+ * A custom ping to help showcase Glean.js.
+ * It is sent everytime the user clicks the "Submit ping" button
+ * on this sample app.
+ *
+ * Generated from `custom`.
+ */
+const custom = new Glean.Glean.default._private.PingType({
```
Dexterp37, line 16: Mh, why are we committing this?

samples/qt-qml-app/sample.js
```diff
@@ -0,0 +1,32 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+"use strict";
+
+/**
+ * The time the sample app was started.
+ *
+ * Generated from `sample.app_started`.
+ */
```
Dexterp37, line 13: Mh, why are we committing this?

glean/package.json
brizental: This should not make any difference for either of them as it only touches the Qt build part.

samples/qt-qml-app/pings.js
brizental: Because this is not generated, glean_parse templates don't support Qt anymore.

brizental: Maybe change the comment to make that more obivous.

glean/package.json
brizental: Either way, I tested in rally and `npm test` turned out green while using the linked glean package.

samples/qt-qml-app/pings.js
Dexterp37: Yeah, let's add something at the top (and maybe a reference to the bug number for adding the QT template to the glean_parser) explaining why we need this in-tree.

samples/qt-qml-app/sample.js
Dexterp37: As above :)

Dexterp37 (approved):
r+wc

PR #228, opened 2021-04-20
Bug 1706103 - Add step to relase Qt build in Publish CI script

brizental:
I copied the way the Glean SDK does it and also how it names the final file.  I tested this locally and it worked, I added a Qt asset to the 0.9.2 with the loca test, see https://github.com/mozilla/glean.js/releases/tag/v0.9.2. 

(self reminder: For this to work I need to add a `GITHUB_TOKEN` env var to CircleCI).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -109,6 +109,23 @@ jobs:
       - run:
           name: Publish to npm
           command: export PATH=.:$PATH && (cd glean && npm publish --access public)
+      - run:
+          name: Get ghr release tool
+          command: |
+            GHR=ghr_v0.13.0_linux_amd64
+            GHR_SHA256=c428627270ae26e206cb526cb8c7bdfba475dd278f6691ddaf863355adadfa13
+            curl -sfSL --retry 5 -O "https://github.com/tcnksm/ghr/releases/download/v0.13.0/${GHR}.tar.gz"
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -109,6 +109,23 @@ jobs:
...
+            tar -xf "${GHR}.tar.gz"
+            cp ./${GHR}/ghr ghr
+      - run:
+          name: Publish to Github
+          command: |
+            # Attach Qt build to release
+            npm --prefix ./glean install
+            npm --prefix ./glean run build:qt
+            cp glean/dist/glean.js glean/dist/glean_js-${CIRCLE_TAG}-qt.js
+            ./ghr -u mozilla -replace ${CIRCLE_TAG} glean/dist/glean_js-${CIRCLE_TAG}-qt.js
```
Dexterp37, line 128: Why do we need to explicitly pass in `-u mozilla` ? [We don't seem to be doing that in the SDK repo](https://github.com/mozilla/glean/blob/ba5f1c8c49471bba22f4cc8e441425c806fc8c19/.circleci/config.yml#L1029).

brizental: From local testing, it seems this is used to build the repo URL. Initially in my local tests it was trying to send it to my fork. This should not change any behaviour in case it is unnecessary.

Dexterp37 (approved):

PR #246, opened 2021-04-26
Remove the observer mechanism to wait for test tasks

Dexterp37:
This PR removes the mechanism based on timeouts and observers to wait on test tasks in the dispatcher. It instead defines a new task type, TestTask, and passes a resolver function when its created.

### DIFF
--- a/glean/src/core/dispatcher.ts
+++ b/glean/src/core/dispatcher.ts
@@ -29,6 +29,8 @@ const enum Commands {
   Stop,
   // The dispatcher should stop executing the queued tasks and clear the queue.
   Clear,
+  // Exactly like a normal Task, but spawned for tests.
+  TestTask,
 }
 
 // A task the dispatcher knows how to execute.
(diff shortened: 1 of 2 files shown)


glean/src/core/dispatcher.ts
```diff
@@ -29,36 +29,26 @@ const enum Commands {
...
 type Task = () => Promise<void>;
 
 // An executable command.
 type Command = {
-  testId?: string,
   task: Task,
   command: Commands.Task,
 } | {
-  command: Exclude<Commands, Commands.Task>,
+  testId?: string,
```
brizental, line 44: Can we also remove the `testId` since we are here. That has proved to be unnecessary.

glean/src/core/dispatcher.ts
```diff
@@ -29,36 +29,26 @@ const enum Commands {
...
+  resolver: (value: void | PromiseLike<void>) => void,
```
brizental, line 45: Are the arguments necessary in this type? If they are never going to be used _I think_ they can be ommited e.g. `() => void`.

brizental (changes requested):
I think we should keep the timeout part. Otherwise tests will hang forever in case `Glean.initialize` is not called before using the testing APIs.

Dexterp37:
> 
> 
> I think we should keep the timeout part. Otherwise tests will hang forever in case `Glean.initialize` is not called before using the testing APIs.
> 
> A clearer error message can be logged in that case, to make the error more actionable for users:
> 
> ```js
> console.error(
>     "Test task took to long to execute.",
>     "Please check if Glean was initialized.",
>     "Bailing out."
> );
> ```

I'm not sure it's worth the complexity to be honest. The whole observer mechanism added quite a few problems in the dispatcher only to account for the fact that Glean was not initialized. I think we should probably just have a boolean somewhere in the dispatcher that's set to true by Glean.js and, if test APIs are called and that's still false, we throw (this is roughly how other bindings do it).

glean/src/core/dispatcher.ts
Dexterp37: Unfortunately they can't be omitted unless we disable TS checks. Typescript was quite unhappy.

brizental:
> I'm not sure it's worth the complexity to be honest. The whole observer mechanism added quite a few problems in the dispatcher only to account for the fact that Glean was not initialized. I think we should probably just have a boolean somewhere in the dispatcher that's set to true by Glean.js and, if test APIs are called and that's still false, we throw (this is roughly how other bindings do it).

Yeah, now that you mention it you don't even need to set a boolean somewhere, you can just get it from the Context.

That was there in case people called testing APIs before Glean was initialized, but initialized Glean right after of something of the sort. Doesn't matter, if this becomes a problem we can file a bug.

brizental (approved):

PR #254, opened 2021-04-26
Bug 1705720 - Ensure the dispatcher is always available

Dexterp37:
This guarantees the dispatcher is always available even before the Glean initialization. We need this because our APIs can be used before initialization, to perform fast recording of metrics.

### DIFF
--- a/glean/src/core/context.ts
+++ b/glean/src/core/context.ts
@@ -58,6 +58,9 @@ export class Context {
       await Context.instance._dispatcher.testUninitialize();
     }
 
+    // Due to the test requirement of keeping storage in place,
+    // we can't simply wipe out the full `Context` instance.
+    // The closest thing we can do is making the dispatcher `null`.
     Context.instance._dispatcher = null;
     Context.initialized = false;
   }
--- a/glean/src/core/context.ts
+++ b/glean/src/core/context.ts
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import type { DebugOptions } from "./debug_options";
-import type Dispatcher from "./dispatcher";
+import Dispatcher from "./dispatcher";
 import type MetricsDatabase from "./metrics/database";
 import type EventsDatabase from "./metrics/events_database";
 import type PingsDatabase from "./pings/database";
@@ -23,7 +23,7 @@ import type PingsDatabase from "./pings/database";
 export class Context {
   private static _instance: Context;
 
-  private _dispatcher!: Dispatcher;
+  private _dispatcher!: Dispatcher | null;
 
   private _uploadEnabled!: boolean;
   private _metricsDatabase!: MetricsDatabase;
@@ -31,13 +31,12 @@ export class Context {
   private _pingsDatabase!: PingsDatabase;
 
   private _applicationId!: string;
-  private _initialized!: boolean;
+  private _initialized: boolean;
 
   private _debugOptions!: DebugOptions;
 
   private constructor() {
-    // Intentionally empty, exclusively defined to mark the
-    // constructor as private.
+    this._initialized = false;
   }
 
   static get instance(): Context {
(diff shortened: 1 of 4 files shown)


glean/src/core/context.ts
```diff
@@ -23,21 +23,20 @@ import type PingsDatabase from "./pings/database";
 export class Context {
   private static _instance: Context;
 
-  private _dispatcher!: Dispatcher;
+  private _dispatcher!: Dispatcher | null;
```
brizental, line 26: Why does this need to be null? Can't we initialize the dispatcher on the Context's constructor?

glean/src/core/context.ts
```diff
@@ -48,7 +47,25 @@ export class Context {
     return Context._instance;
   }
 
+  /**
+   * **Test-only API**
+   *
+   * Resets the Context to an uninitialized state.
+   */
+  static testReset(): void {
+    Context.instance._dispatcher = null;
```
brizental, line 56: This makes [this line](https://github.com/mozilla/glean.js/pull/254/files#diff-e8fb6737d2cc01074b542c0ab05c2221b45f8b178c14d64e353ff4439c17fad4L511) unnecessary. We should choose between one of them.

glean/src/core/context.ts
```diff
@@ -48,7 +47,25 @@ export class Context {
     return Context._instance;
   }
 
+  /**
+   * **Test-only API**
+   *
+   * Resets the Context to an uninitialized state.
+   */
+  static testReset(): void {
```
brizental, line 55: nit: Rename to `testUninitialize` to keep consistency.

brizental (changes requested):
Would you mind also adding a simple test to ensure this behaviour?

glean/src/core/context.ts
Dexterp37: This needs to be null because we can't really tear down the whole `Context` when we reset Glean, because we need to retain the storage. So we do a "soft reset" by nulling out the dispatcher in `testReset`

glean/src/core/context.ts
Dexterp37: Mh, I don't think it does: if we just null out the dispatcher without tearing down the old one we might end up with tasks still being executed because they still are there.

glean/src/core/context.ts
brizental: nit: I would prefer `private _dispatcher?: Dispatcher;`, but no strong requirement for this change.

glean/src/core/context.ts
brizental: Hm, indeed. If the Context is responsible for the Dispatcher maybe it should be one to uninitialize it as well though. This is a nit though, feel free to ignore.

Dexterp37:
> 
> 
> Would you mind also adding a simple test to ensure this behaviour?

 Now that we're properly resetting, all the tests fail without the fix (instantiating the dispatcher if not available). Any test I might add would basically be the same, but let me see what I can pull off

glean/src/core/context.ts
Dexterp37: Good point, done!

glean/src/core/context.ts
Dexterp37: That makes it `undefined` though,

brizental: Yes, but this a very philosophical thing (`null` vs `undefined`) and makes no real difference here. Feel free to ignore the request ;)

glean/src/core/context.ts
```diff
@@ -48,7 +47,30 @@ export class Context {
...
+   *
+   * Resets the Context to an uninitialized state.
+   */
+  static async testUninitialize(): Promise<void> {
+    // Clear the dispatcher queue and return the dispatcher back to an uninitialized state.
+    if (Context.instance._dispatcher) {
+      await Context.instance._dispatcher.testUninitialize();
+    }
+
+    Context.instance._dispatcher = null;
```
brizental, line 61: Can you add a comment here why we make it `null` even after uninitializing?

brizental (approved):

PR #261, opened 2021-04-28
Add a spinner to the console output while setting up python virtual env

ChinYing-Li:
The PR aims to tackle bug 1699391, by running a spinner in the console during the creation of the python virtual environment.
Any suggestion is appreciated!


glean/src/cli.ts
```diff
@@ -214,3 +218,26 @@ async function run(args: string[]) {
 run(argv).catch(e => {
   console.error("There was an error running Glean", e);
 });
+
+/**
+ * Returns a spinner
+ *
+ * @returns an Interval ID that logs certain characters
+ */
+function GetSpinner() {
```
brizental, line 227: nit: What do you think about renaming this function to `startSpinner`?

glean/src/cli.ts
```diff
@@ -184,6 +187,7 @@ async function runGlean(projectRoot: string, parserArgs: string[]) {
     exec.exec(cmd, (err, stdout, stderr) => resolve({err, stdout, stderr}));
   });
 
+  StopSpinner(spinner);
```
brizental, line 190: I think the spinner is not necessary for this function. For me it doesn't even show up because this one is super fast :)

glean/src/cli.ts
```diff
@@ -214,3 +218,26 @@ async function run(args: string[]) {
...
+    i %= 4;
+  }, 250);
+}
+
+/**
+ * Stops the spinner
+ *
+ * @param spinner is created by GetSpinner
+ */
+function StopSpinner(spinner: NodeJS.Timeout) {
```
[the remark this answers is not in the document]
brizental: Is it possible in this function to also clean up the last character left by the spinner that was just stopped?

If that would be too hard or require hacks, don't worry, it's not a big deal. Just wondering.

brizental (changes requested):
Big fan of this change. Great job @ChinYing-Li !

Left a few comments but nothing major :D

glean/src/cli.ts
ChinYing-Li: I initially want to use `startSpinner`, but the function does more than that. Since it returns something, I changed the function name to `getStartedSpinner`. What do you think?

brizental: Sounds good!

glean/src/cli.ts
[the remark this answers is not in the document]
ChinYing-Li: Can be done with one line of code.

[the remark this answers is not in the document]
brizental: Amazing!

glean/src/cli.ts
brizental: That sounds totally reasonable, Let's keep it.

ChinYing-Li:
Once this PR is good to go, I will rebase it to resolve the conflicts..

PR #279, opened 2021-05-03
Bug 1708366: Ensure only empty pings triggers logging of "empty ping" message

ChinYing-Li:
I think the original logic is wrong, in that the message `Storage for ${ping.name} empty. Ping will still be sent.` will be logged if `eventsData` is defined.
https://github.com/mozilla/glean.js/blob/0557f318a4637b471850313a8006f46a32b6a1fa/glean/src/core/pings/maker.ts#L198-L203

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 [Full changelog](https://github.com/mozilla/glean.js/compare/v0.11.0...main)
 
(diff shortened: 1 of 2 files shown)


Dexterp37 (approved):
Good catch! Yes, the logic was wrong before. Great PR!

PR #288, opened 2021-05-06
Bug 1706102 - Collect Qt platform information

Dexterp37:
This uses the global `Qt` object to get the OS name and the locale. Unfortunately, os version and architecture are not available.

I manually tested that this works using the Glean.js Qt sample app. I get the following output:

```
D:\Mozilla\glean.js\samples\qt-qml-app>python main.py
qml: Initialized Glean succesfully.
qml: Adding to the `button_clicked` metric.
qml: Attempted to delete an entry from an invalid index. Ignoring.
qml: {
  "metrics": {
    "datetime": {
      "sample.app_started": "2021-05-06T16:17:26.529+02:00"
    },
    "counter": {
      "sample.button_clicked": 1
    }
  },
  "ping_info": {
    "seq": 0,
    "start_time": "2021-05-06T16:17+02:00",
    "end_time": "2021-05-06T16:17+02:00"
  },
  "client_info": {
    "telemetry_sdk_build": "0.11.0",
    "client_id": "06cbc5b3-2c1b-4cc5-bdaf-85ba62a4afca",
    "first_run_date": "2021-05-06+02:00",
    "os": "Windows",
    "os_version": "Unknown",
    "architecture": "Unknown",
    "locale": "it_IT",
    "app_build": "Unknown",
    "app_display_version": "Unknown"
  }
}
qml: Ping e23108de-1a60-4091-9b7f-b70215d3115d succesfully sent 200.
```

### DIFF
--- a/glean/src/platform/qt/platform_info.ts
+++ b/glean/src/platform/qt/platform_info.ts
@@ -1,0 +1,55 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import type PlatformInfo from "../../core/platform_info.js";
+import { KnownOperatingSystems } from "../../core/platform_info.js";
+
+// The `Qt` symbol can be used in this file without TS compiler errors
+// because that symbol is described as a constant in `./types/Qt/index.d.ts`.
+
+const QtPlatformInfo: PlatformInfo = {
+  // eslint-disable-next-line @typescript-eslint/require-await
+  async os(): Promise<KnownOperatingSystems> {
+    // Possible values are listed in https://doc.qt.io/qt-5/qml-qtqml-qt.html#platform-prop
+    const osName = Qt.platform.os;
+    switch(osName) {
+    case "android":
+      return KnownOperatingSystems.Android;
+    case "ios":
+      return KnownOperatingSystems.iOS;
+    case "tvos":
+      return KnownOperatingSystems.TvOS;
+    case "linux":
+      return KnownOperatingSystems.Linux;
+    case "osx":
+      return KnownOperatingSystems.MacOS;
+    case "qnx":
+      return KnownOperatingSystems.Qnx;
+    case "windows":
+    case "winrt":
+      return KnownOperatingSystems.Windows;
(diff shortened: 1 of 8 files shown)


glean/src/platform/qt/platform_info.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  async osVersion(): Promise<string> {
+    // This data point is not available in Qt QML.
+    return Promise.resolve("Unknown");
+  },
+
+  async arch(): Promise<string> {
+    // This data point is not available in Qt QML.
+    return Promise.resolve("Unknown");
+  },
```
Dexterp37, line 47: @bakulf do you know any way to get this in Qt-QML?

glean/src/platform/qt/platform_info.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  async locale(): Promise<string> {
+    const locale = Qt.locale();
```
bakulf, line 50: Reading this: https://doc.qt.io/qt-5/qml-qtqml-qt.html#locale-method it seems that if `name` is not valid, QT uses the 'C' locale. Have you tested if this works with non-'C' locales?

bakulf (commented):
My main concern is related to the use of Qt.locale(). The rest of the code looks good!

glean/src/platform/qt/platform_info.ts
Dexterp37: Yup, this seems to behave as expected, thanks!

badboy: I'm not sure I understand the docs, so for understanding: this will give us the dash-separated locale like `es-ES` or `en-US`?

badboy (approved):
Seems simple enough for what's there. Just the one question about the locale.
If Qt has functionality to get os_version and arch, then yeah, we should include that, but it's fine to report unknown for now too

glean/src/platform/qt/platform_info.ts
Dexterp37: > 
> 
> I'm not sure I understand the docs, so for understanding: this will give us the dash-separated locale like `es-ES` or `en-US`?

Good thing you mentioned this, I double checked and I missed that we were sending `es_ES` and not `es-ES`. I twaked the code to fix this :)

PR #292, opened 2021-05-10
Bug 1710377 - Only check string format when validating UUID

brizental:
For legacy reasons we need to accept non-RFC'd UUID strings. 

See also: https://github.com/mozilla-extensions/bergamot-browser-extension/issues/106#issuecomment-831838830

### DIFF
--- a/glean/src/core/metrics/types/uuid.ts
+++ b/glean/src/core/metrics/types/uuid.ts
@@ -6,9 +6,11 @@ import type { CommonMetricData } from "../index.js";
 import { MetricType } from "../index.js";
 import { generateUUIDv4, isString } from "../../utils.js";
 import { Context } from "../../context.js";
-import { validate as UUIDvalidate } from "uuid";
-import { KNOWN_CLIENT_ID } from "../../constants.js";
 import { Metric } from "../metric.js";
+
+// Loose UUID regex for checking if a string has a UUID shape.
+// Note: This does not contain version checks.
+const UUID_REGEX = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
 
 export class UUIDMetric extends Metric<string, string> {
   constructor(v: unknown) {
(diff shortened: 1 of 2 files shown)


glean/src/core/metrics/types/uuid.ts
```diff
@@ -6,10 +6,12 @@ import type { CommonMetricData } from "../index.js";
 import { MetricType } from "../index.js";
 import { generateUUIDv4, isString } from "../../utils.js";
 import { Context } from "../../context.js";
-import { validate as UUIDvalidate } from "uuid";
-import { KNOWN_CLIENT_ID } from "../../constants.js";
 import { Metric } from "../metric.js";
 
+// Loose UUID regex for checking if a string has a UUID shape.
+// Note: This does not contain version checks.
```
Dexterp37, line 12: Can you expand on why we need this? (I have the context given the chat in the team channel, but it may be worth mentioning it here too for future references)

glean/tests/core/metrics/timespan.spec.ts
```diff
@@ -76,7 +76,7 @@ describe("TimespanMetric", function() {
   });
 
   it("ping payload is correct", async function() {
-    const fakeDateNow = sandbox.stub(Date, "now");
+    const fakeDateNow = performance ? sandbox.stub(performance, "now") : sandbox.stub(Date, "now");
```
Dexterp37, line 79: Can we do this once in `beforeEach`?

Dexterp37 (changes requested):

glean/tests/core/metrics/timespan.spec.ts
```diff
@@ -274,13 +272,12 @@ describe("TimespanMetric", function() {
     // Nothing should have been recorded.
     assert.strictEqual(await metric.testGetValue("aPing"), undefined);
 
-    // TODO: Make sure also incalid state error was recorded.
+    // TODO: Make sure also invalid state error was recorded.
```
Dexterp37, line 275: Why did you change this? incalid is much clearer!

Dexterp37 (approved):

.circleci/config.yml
```diff
@@ -9,7 +9,7 @@ version: 2.1
 jobs:
   lint:
     docker:
-      - image: cimg/node:14.13.1
```
brizental, line 12: Somehow this started failiing because the npm version that comes with this image relies on package-lock v1.

Dexterp37: Mh, that's odd.

Dexterp37 (approved):
Looks good to me!

PR #303, opened 2021-05-11
Bug 1709787 - Implement setRawNanos API for the TimespanMetricType

brizental:


glean/src/core/metrics/types/timespan.ts
```diff
@@ -172,6 +198,30 @@ class TimespanMetricType extends MetricType {
...
+   * than one timespan measurement. To be safe, this function should generally
+   * be followed by sending a custom ping containing the timespan.
+   *
+   * @param elapsed The elapsed time to record, in nanoseconds.
+   */
+  setRawNanos(elapsed: number): void {
+    Context.dispatcher.launch(async () => {
+      // `elapsed` is in nanoseconds in order to match the glean-core API.
+      // Javascript actually does not provide nanosecond precision,
+      // the highest resolution it provides is milliseconds
```
Dexterp37, line 218: That is not the reason why we convert: we convert because the API expects nanoseconds. I think we can safely get these two lines out of this comment.

Dexterp37 (changes requested):

Dexterp37 (approved):

PR #313, opened 2021-05-12
Bug 1682574 - Implement error recording

brizental:
Fun stuff: 

- The errors recorded on the datetime metric type over on glean-core do not make sense for the Datetime metric in Glean.js.
- It is not possible to implement testGetNumRecordedErrors for labeled metrics.

The latter maybe we can work around. The first is a documentation challenge.


Dexterp37 (changes requested):
This looks great, thanks for doing this.

There's a couple of instances of missing tests in `labeled.spec.ts` (search for "bug 1682574"). Would you kindly add them to this PR?

glean/src/core/context.ts
```diff
@@ -29,6 +30,7 @@ export class Context {
   private _metricsDatabase!: MetricsDatabase;
   private _eventsDatabase!: EventsDatabase;
   private _pingsDatabase!: PingsDatabase;
+  private _errorManager!: typeof ErrorManager;
```
Dexterp37, line 33: Why do we need this `typeof``?

glean/src/core/error/index.ts
```diff
@@ -0,0 +1,86 @@
...
+    name: combineIdentifierAndLabel(error, name),
+    category: "glean.error",
+    lifetime: "ping",
+    // TODO: Also add the metric ping to the list. Depends on Bug 1710838.
+    sendInPings: metric.sendInPings,
+    disabled: false,
+  });
+}
+
+export default {
```
Dexterp37, line 38: Do we need the typeof above because of the way we are exporting here, i.e. not using a class?

glean/src/core/glean.ts
```diff
@@ -197,6 +197,7 @@ class Glean {
     Context.metricsDatabase = new MetricsDatabase(Glean.platform.Storage);
     Context.eventsDatabase = new EventsDatabase(Glean.platform.Storage);
     Context.pingsDatabase = new PingsDatabase(Glean.platform.Storage);
+    Context.errorManager = ErrorManager;
```
Dexterp37, line 200: Can you add a comment here about why we're setting this on the `Context`?

Dexterp37 (approved):
r+wc

glean/src/core/error/index.ts
brizental: Yes, would you prefer that I change that?

brizental: To clarify I just didn't see the point of making this class, but I can if you hate the `typeof` :D

brizental: Eh, I changed it to a class.

Dexterp37: Thanks :)

PR #319, opened 2021-05-17
Bug 1710827 - Fail validation of counter and quantity if value is not int

brizental:


glean/src/core/utils.ts
```diff
@@ -102,6 +102,21 @@ export function isNumber(v: unknown): v is number {
...
+ *
+ * @param v The value to verify.
+ *
+ * @returns A special Typescript value (which compiles down to a boolean)
+ *          stating whether `v` is a number.
+ */
+export function isInteger(v: unknown): v is number {
+  // Decided against using `Number.isInteger` here,
+  // because that fails if you pass it a number created through `new Number(...)`.
+  return isNumber(v) && v % 1 === 0;
```
Dexterp37, line 116: If we don't want to use `Number.isInteger`, can we at least use the [recommended polyfill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger#polyfill)?

glean/src/core/utils.ts
```diff
@@ -102,6 +102,21 @@ export function isNumber(v: unknown): v is number {
...
+}
+
+
```
Dexterp37, line 119: extra blank?

glean/tests/core/utils.spec.ts
```diff
@@ -79,6 +79,20 @@ describe("utils", function() {
     assert.strictEqual(utils.isNumber(new Number(10)), true);
   });
 
+  it("isInteger validates correctly", function() {
+    // Invalid values
```
Dexterp37, line 83: There's some other cases in [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger#using_isinteger) :)

Dexterp37 (changes requested):

glean/src/core/utils.ts
brizental: Sure :) Didn't know there was one... FWIW, all the listed test cases still pass with the current implementation the polyfill. I'll change to be extra sure.

glean/tests/core/utils.spec.ts
brizental:
```js
assert.strictEqual(utils.isInteger(5.0000000000000001), true);
assert.strictEqual(utils.isInteger(5.000000000000001), false);
```

Now this is a fun fact.

glean/src/core/utils.ts
brizental: Where is that linter rule when we need it?

glean/src/core/utils.ts
brizental: Hm, actually we can't really use the polyfill because `Number.isFinite` has the same issue as `Number.isInteger`. This is interesting behaviour though. Why don't these work? I though `new Number()` was just another way to create a number in Javascript.

brizental: Ok, I was wrong about it! https://stackoverflow.com/questions/2381399/what-is-the-difference-between-new-number-and-number-in-javascript

brizental: Ok so, at some point I stumbled into this before because I do account for it in our checks. https://github.com/mozilla/glean.js/blob/main/glean/src/core/utils.ts#L78

To be consistent, I will remove `new Type()` from all our checks. These should not be the way most users if not all create primitives anyways. If ever this becomes a problem we can change it.

Dexterp37 (approved):

PR #324, opened 2021-05-18
Bug 1707629 - Add a apply import extensions lint rule

brizental:
The rule will actually warn about missing ".ts" extension. When you add the ".js" extension it will stop complaining though, so that works well enough I guess :)

The plugin I added also added some neat new rules about duplicates and what not.


glean/tests/platform/utils/webext/sample/background.js
```diff
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-"use strict";
```
Dexterp37, line 5: Why is this removing "use strict"?

brizental: It is superfluous. Modules are always run in strict mode.

brizental: There is a new rule in the added plugin that will check if it is a module and remove "use strict" if so.

Dexterp37 (approved):

PR #342, opened 2021-05-21
Bug 1682282 - Add tests to validate generated pings against the schema

brizental:
Found a nasty bug: timespan payload was incorrect (see: 8b44189). Guess it's good that we have the schema test now...


glean/package.json
```diff
@@ -35,24 +35,28 @@
...
-    "build:webext": "rm -rf dist/webext && npm run build:webext:lib && npm run build:webext:types",
+    "build:webext": "rm -rf dist/webext && run-s build:webext:lib build:webext:types",
     "build:qt": "webpack --config webpack.config.qt.js --mode production",
     "build:docs": "rm -rf dist/docs && typedoc --out dist/docs src/ --theme minimal",
     "publish:docs": "NODE_DEBUG=gh-pages gh-pages --dotfiles --message \"[skip ci] Updates\" --dist dist/docs",
-    "prepublishOnly": "cp ../README.md ./README.md && npm run build:cli && npm run build:webext",
-    "postpublish": "rm ./README.md"
+    "prepublishOnly": "cp ../README.md ./README.md && run-s build:cli build:webext",
+    "postpublish": "rm ./README.md",
+    "glean-internal": "node --loader=ts-node/esm src/cli.ts"
```
Dexterp37, line 59: For the sake of clarify, what about calling this `glean-cli-internal`?

glean/tests/integration/schema/metrics.yaml
```diff
@@ -0,0 +1,158 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+---
+$schema: moz://mozilla.org/schemas/glean/metrics/2-0-0
+
+for_testing:
```
Dexterp37, line 8: Should we change [this document](https://github.com/mozilla/glean.js/blob/main/docs/adding_a_new_metric_type.md) to reference this file and add updating the related test as a requirement?

glean/tests/integration/schema/schema.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+    }).on("error", (err) => {
+      reject(err);
+    });
+  });
+}
+
+/**
+ * A Glean mock HTTP which allows one to wait for a specific ping submission.
+ */
+class WaitableHttpClient implements Uploader {
```
Dexterp37, line 46:
```suggestion
class WaitableHttpUploader implements Uploader {
```

?

glean/tests/unit/core/metrics/timespan.spec.ts
```diff
@@ -171,7 +174,7 @@ describe("TimespanMetric", function() {
       assert.strictEqual(await metric.testGetValue("aPing"), testCases[i].expected);
 
       sandbox.restore();
-      fakeNow = performance ? sandbox.stub(performance, "now") : sandbox.stub(Date, "now");
+      fakeNow = typeof performance === "undefined" ? sandbox.stub(Date, "now") : sandbox.stub(performance, "now");
```
Dexterp37, line 177:
```suggestion
      fakeNow = performance === undefined ? sandbox.stub(Date, "now") : sandbox.stub(performance, "now");
```

?

Dexterp37 (changes requested):

glean/tests/unit/core/metrics/timespan.spec.ts
brizental: Type of is mandatory here otherwise it will fail if performance is not in the environment. This was wrong before and failed after I changed the Docker image used in the test CI job.

glean/tests/integration/schema/schema.spec.ts
brizental: Sure, The name I chose was because the property in the config is `httpClient`, should we rename that also?

glean/package.json
brizental: Yeah, hm... Also not a fan od that command name. How about just `cli`?

glean/tests/integration/schema/schema.spec.ts
Dexterp37: Mh no, fine: if we want to change the name of that property then we should probably do it for all the clients. Let's ignore this comment.

glean/package.json
Dexterp37: Sounds good!

glean/tests/integration/schema/metrics.yaml
brizental: I did update that file in this PR with instructions to add tests to new metric types here. Is that sort of change what you are referring to?

Dexterp37: Oh, right.

Dexterp37 (approved):

PR #375, opened 2021-05-28
Bug 1707896 - Make the Qt package a QML module

brizental:
Opening as a draft as this depends on https://github.com/mozilla/glean_parser/pull/342

I have tested manually of course, but we will be sure once CI turns out green after updating glean_parser 👀 

Just remembered that I still need to:

- [x] Change the way we release the Qt package in order to release it as a QML module (i.e. with the long path + qmldir file).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -209,8 +209,8 @@ jobs:
             # Attach Qt build to release
             npm --prefix ./glean install
             npm --prefix ./glean run build:qt
-            cp glean/dist/glean.js glean/dist/glean_js-${CIRCLE_TAG}-qt.js
-            ./ghr -u mozilla -replace ${CIRCLE_TAG} glean/dist/glean_js-${CIRCLE_TAG}-qt.js
+            tar -zcvf "glean/dist/glean_js-${CIRCLE_TAG}-qt.tar.gz" glean/dist/qt
+            ./ghr -u mozilla -replace ${CIRCLE_TAG} glean/dist/glean_js-${CIRCLE_TAG}-qt.tar.gz
 
 workflows:
   version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -135,7 +135,8 @@ jobs:
             python3 -m venv venv
             source venv/bin/activate
             pip install -r requirements.txt
-            glean_parser translate metrics.yaml pings.yaml -f javascript -o generated --option platform=qt
+            glean_parser translate metrics.yaml pings.yaml -f javascript -o generated \
+            --option platform=qt --option version=0.14
 
             sudo apt-get install xvfb
             xvfb-run python main.py &> qml.log &
--- a/bin/prepare-release.sh
+++ b/bin/prepare-release.sh
@@ -114,13 +114,16 @@ run $SED -i.bak -E \
 run rm "${WORKSPACE_ROOT}/${FILE}.bak"
 
 ### Qt sample app ###
+# Qt changes are necessary because QML requires that you add
+# version number along with the import statements.
 
-# This gets the version without the path version.
-GLEAN_VERSION_FOR_QML=$(node -p -e "require('${WORKSPACE_ROOT}/glean/package.json').version.split('.').reverse().slice(1).reverse().join('.')")
+# This gets the version string without the patch version.
+GLEAN_VERSION_FOR_QML=$(node -p -e "'${NEW_VERSION}'.split('.').reverse().slice(1).reverse().join('.')")
 
 FILE=samples/qt-qml-app/main.qml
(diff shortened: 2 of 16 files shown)


glean/qt.entry.js
```diff
@@ -0,0 +1,86 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
brizental, line 1: This file will be copied to the final `dist/qt/org/mozilla/` folder and it will be the actual entry point from QML, not the file under `src/index/`

samples/qt-qml-app/main.qml
```diff
@@ -6,10 +6,8 @@ import QtQuick 2.0
 import QtQuick.Controls 2.0
 import QtGraphicalEffects 1.15
 
-import "../../glean/dist/glean.js" as Glean;
-// These must be imported after Glean because they rely on Glean being in the environment.
-import "./generated/pings.js" as Pings;
-import "./generated/sample.js" as Metrics;
+import org.mozilla.Glean 0.14
+import "./generated"
```
brizental, line 10: Because the module created by glean_parser will [depend on `org.mozilla.Glean`](https://github.com/mozilla/glean_parser/pull/342/files#diff-4fc68f3830dc22e9638cb0c79eecf5185858b67c5fbf693b4744c74921c00ba2R5) it does not matter anymore whether the generated stuff is imported before of after the Glean import. As a matter of fact, it's not even necessary to have a Glean import at all for them to work, because [they import Glean themselves](https://github.com/mozilla/glean_parser/pull/342/files#diff-df168f83dce8d90814da12deb8e661a62d1ab187c912e5c9b3667716ecc21412R29).

bin/prepare-release.sh
```diff
@@ -113,6 +113,29 @@ run $SED -i.bak -E \
     "${WORKSPACE_ROOT}/${FILE}"
 run rm "${WORKSPACE_ROOT}/${FILE}.bak"
 
+### Qt sample app ###
```
brizental, line 116: This is necessary because QML required that you add version number along with the import statements.

Dexterp37: Can you add this as a comment in this file?

bin/prepare-release.sh
```diff
@@ -113,6 +113,29 @@ run $SED -i.bak -E \
...
+
+# This gets the version without the path version.
```
Dexterp37, line 118:
```suggestion
# This gets the version without the patch version.
```

bin/prepare-release.sh
```diff
@@ -113,6 +113,29 @@ run $SED -i.bak -E \
...
+GLEAN_VERSION_FOR_QML=$(node -p -e "require('${WORKSPACE_ROOT}/glean/package.json').version.split('.').reverse().slice(1).reverse().join('.')")
```
Dexterp37, line 119: Mh, instead of doing this magic incantation, can't we take it from `${NEW_VERSION}` (dropping the patch version)? This will also make sure this still works if the order of the updates within this file changes..

glean/qt.entry.js
Dexterp37: Can we copy this under `src/index` then? Can we also add a comment to the `src/index/qt.ts` file stating this? (and also to this file..)

Dexterp37: Maybe also expand on why we need this file and the index file is not alone (I understood why, but we should track this knowledge in the code)

Dexterp37 (changes requested):

bin/prepare-release.sh
brizental: Yeah... sure...  I was avoiding doing that lol

glean/qt.entry.js
brizental: Sure,k I thought about putting it there but it looked weird. Then again, it looks weird where it is right now.

Dexterp37 (approved):
Neat. Would you kindly file a bug about adding docs about "adding Glean.js to a QML project"?

.circleci/config.yml
```diff
@@ -135,7 +135,8 @@ jobs:
             python3 -m venv venv
             source venv/bin/activate
             pip install -r requirements.txt
-            glean_parser translate metrics.yaml pings.yaml -f javascript -o generated --option platform=qt
+            glean_parser translate metrics.yaml pings.yaml -f javascript -o generated \
```
Dexterp37, line 138: Please let's also add a changelog entry.

brizental:
> Would you kindly file a bug about adding docs about "adding Glean.js to a QML project"?

https://bugzilla.mozilla.org/show_bug.cgi?id=1714361

PR #376, opened 2021-05-29
Bug 1712912 - Remove unset API for debugging features

AmrDhaliwal:


AmrDhaliwal:
Removed the unset API from the debugging feature. 
Had a few errors with circleci but, fixed them by removing
leftover parts of the unset code blocks.

glean/tests/unit/core/pings/maker.spec.ts
```diff
@@ -133,11 +133,6 @@ describe("PingMaker", function() {
       "X-Debug-ID": "test",
       "X-Source-Tags": "tag1,tag2,tag3"
     }, PingMaker.getPingHeaders(Context.debugOptions));
-
-    Glean.unsetDebugViewTag();
-    Glean.unsetSourceTags();
-    await Context.dispatcher.testBlockOnQueue();
-    assert.strictEqual(PingMaker.getPingHeaders(Context.debugOptions), undefined);
```
brizental, line 140: I am not confortable removing this assertion. The way to unset these tags after removal of the unset APIs is to reset Glean itself. Thus, running `testResetGlean` should work to unset them.

brizental (changes requested):
Thanks for taking on this work, @AmrDhaliwal ! Let's also remove the `unserSourceTags` API here.

glean/tests/unit/core/pings/maker.spec.ts
AmrDhaliwal: Would I just want to place this assertion back into the maker.spec.ts file?

brizental: Yes.

AmrDhaliwal:
@brizental I put the assertion back in the maker.spec.ts file and removed the unsetSourceTags
but, I am unable to find why the circleci test is failing.

brizental:
> @brizental I put the assertion back in the maker.spec.ts file and removed the unsetSourceTags but, I am unable to find why the circleci test is failing.

Try running the failing test locally, to investigate the error message. Instructions on how to run the tests can be found [here](https://github.com/mozilla/glean.js/blob/main/docs/getting_started.md#running-tests). 

Note, that simply adding the assertion back will not be enough. That assertion verifies that the tags were cleared succesfully, but the code to API to clear the tags was removed, so we need to clear them in some other way. That way is by resetting Glean, which can be done in tests by calling the `testResetGlean` API.

AmrDhaliwal:
@brizental I added "Glean.testResetGlean();" to the maker.spec.ts file and am no longer receiving the error message
and all tests on circleci are passing as well.

brizental (approved):
Looks great! Thanks for the thorough work here. 

Les us know if you would like to take a next bug :)

AmrDhaliwal:
@brizental If you have another bug in mind I would be happy to take it on.

brizental:
> @brizental If you have another bug in mind I would be happy to take it on.

For sure, check out  https://bugzilla.mozilla.org/show_bug.cgi?id=1712914. It is very similar to this one in the sense that it is the _removal_ of an API. If you think it looks interesting just leave a comment there and I will assign it to you :) Otherwise I am open to discuss other bugs that you think are more interesting. Thanks a lot for the excitement in contributing to Glean.js!

PR #399, opened 2021-06-07
Bug 1707581 - Check if there are ping data before attempting to delete it

brizental:
Thought this was an actual bug. Glad it wasn't really :)


Dexterp37 (approved):
r+wc: please add a changelog entry.

PR #411, opened 2021-06-08
Bug 1712285 - Tag all log messages

brizental:
This allows to easily add functionality to turn debug logs off. Filed a bug for it: https://bugzilla.mozilla.org/show_bug.cgi?id=1715312

We should also audit the levels of these messages, some messages are internal error but are in the Error level. They should really be debug, or trace... Anyways, that can be done as part of the but I created above.


Dexterp37 (commented):
r+ with a changelog entry added.

Dexterp37 (approved):

PR #415, opened 2021-06-14
Bug 1712080 - Gzip ping paylod before upload

brizental:
Should I file a bug to document that pings are gzipped? I looked and that is not mentioned on the Glean book. That could be important for folks using custom uploaders for testing, but that could also be behaviour we want to disencourage... Even if we don't disencourage it, we might want to expose the [WaitableHttpClient](https://github.com/mozilla/glean.js/blob/main/glean/tests/integration/schema/schema.spec.ts#L46-L81) in the future. Then people still don't need to worry about gzipping.

### DIFF
--- a/glean/src/core/upload/index.ts
+++ b/glean/src/core/upload/index.ts
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+import { gzipSync } from "fflate";
+
 import type Platform from "../../platform/index.js";
 import type { Configuration } from "../config.js";
 import { GLEAN_VERSION } from "../constants.js";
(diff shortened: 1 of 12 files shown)


glean/src/core/upload/index.ts
```diff
@@ -124,25 +126,35 @@ class PingUploader implements PingsDatabaseObserver {
    */
   private async preparePingForUpload(ping: QueuedPing): Promise<{
     headers: Record<string, string>,
-    payload: string
+    payload: string | Uint8Array
```
Dexterp37, line 129: Would you kindly document why accepting both is needed? (QML not supporting encoding easily?)

glean/src/platform/qt/fflate.stub.js
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * This file is a drop in replacement for `fflate` in QML.
+ * This library raises errors in QML. https://github.com/101arrowz/fflate/issues/71
+ * could be the the culprit, but it is unclear at this point.
+ */
+
+// eslint-disable-next-line jsdoc/require-jsdoc
+export function gzipSync() {
+  throw new Error("Attempted to use `gzipSync` from QML, but that is not supported.");
```
Dexterp37, line 13: Can you add a comment explaining why we throw and a link to the bug for implementing this, if we're planning to

glean/src/core/upload/index.ts
brizental: Well, WML support is not the only reason. We do the same in glean-core, gzipping action may fail and it it does then we send the unzipped ping. I will add a line about this though :)

Dexterp37 (approved):
Would you kindly do, if you didn't already, a test with the debug view (before merging)?

brizental:
> Would you kindly do, if you didn't already, a test with the debug view (before merging)?

https://debug-ping-preview.firebaseapp.com/pings/gzipped-webext Did that while developing, ignore the failing one that was just _development_.

PR #422, opened 2021-06-15
Bug 1681899 - Update and re-enable size check CI job

brizental:
The catch here is that it is not possible to run the CI check in this PR, because the script expects the code in the main branch to have some scripts which it currently doesn't.

I'll file a fake PR to validate.


brizental:
UPDATE: Size report is working, but the issue on getting environment variables from forks is not. I have contacted Data SRE since I have no permissions to create restricted contexts. Until that is setup, this PR is blocked.

Sample report can be checked out [here])(https://github.com/mozilla/glean.js/pull/423#issuecomment-862318040)

brizental:
Ok, simply enabling the restricted context was not enough for giving forks access to the environment variables. It is indeed necessary to turn on "Sharing secrets with forks". Still, running in restricted contexts is not allowed for PRs coming from forks, [apparently](https://circleci.com/docs/2.0/contexts/#running-workflows-with-a-restricted-context). Now, I tried to test that in the validator PR, but maybe because it was my fork and I am part of the group that has access to the restricted context, it did not fail for me.

For now, what I did was move all our environment variables to the restricted context, so none of them are in the project anymore (only people from the `glean` github organization are allowed to access that context). For the other job that requires a secret, I added the context as required. Finally, I changed our configuration to allow forks to access secrets. This should be fine: they will not be able to access the secret unless a memeber from the `glean` org manually approves it, at least that is how I understand the CircleCI documentation I read above. 

One think to keep in mind (and write documentation about, possibly): we should not add sensitive information to the project environment variables from now on, only to the restricted context!

brizental:
Ok, this is finally ready for review. Thanks to @scholtzan for checking that forks indeed get an unauthorized error if they try to access restricted contexts (see: https://github.com/mozilla/glean.js/pull/449) and also for flagging that SSH keys needed to be added to the restricted context as well.

Dexterp37 (approved):
💯

PR #448, opened 2021-06-21
Bug 1716724 - Events treshold for custom pings

Dexterp37:
This work in progress PR is a first attempt at implementing [this proposal](https://docs.google.com/document/d/1R-tIZYz4Izx8rDB_YUiJXU9Xtk8VcJoF8Pmhggps_GU/edit?ts=60992051#).

TODO

- [x] Compute the timestamps relative to the reference data.
- [x] Add unit tests.
- [ ] Add integration tests.
- [ ] Update documentation.

### DIFF
--- a/glean/src/core/metrics/events_database.ts
+++ b/glean/src/core/metrics/events_database.ts
@@ -58,6 +58,24 @@ export class RecordedEvent {
       e["extra"] as ExtraMap | undefined
     );
   }
+
+  public withoutReservedExtras(): RecordedEvent {
+    // If this event does not have any extras, return the same instance.
+    if (!this.extra) {
+      return this;
+    }
+
+    // The following line applies destructuring to return an object without
+    // the `gleanStartupDate` and `gleanExecutionCounter` properties.
+    const {gleanStartupDate, gleanExecutionCounter, ...filteredExtras } = this.extra;
+
+    return new RecordedEvent(
+      this.category,
+      this.name,
+      this.timestamp,
+      (filteredExtras && Object.keys(filteredExtras).length > 0) ? filteredExtras : undefined
+    );
+  }
 }
 
 /**
@@ -197,7 +215,8 @@ class EventsDatabase {
       // Only report events for the requested metric.
       .filter((e) => {
         return (e.category === metric.category) && (e.name === metric.name);
-      });
+      })
+      .map((e) => e.withoutReservedExtras());
   }
 
   /**
@@ -300,8 +319,6 @@ class EventsDatabase {
         const dateOffsetInMs = Date.parse(event.extra!["gleanStartupDate"]) - Date.parse(previousRestartDate);
         previousRestartDate = event.extra!["gleanStartupDate"];
 
-        // TODO Drop the 'extras' from this event.
-
         // Update the current offset and move to the next event.
         currentOffsetInMs += dateOffsetInMs;
       }
@@ -311,7 +328,7 @@ class EventsDatabase {
       sortedData[index] = new RecordedEvent(event.category, event.name, event.timestamp + currentOffsetInMs, event.extra);
     });
 
-    return sortedData.map((e) => RecordedEvent.toJSONObject(e));
+    return sortedData.map((e) => RecordedEvent.toJSONObject(e.withoutReservedExtras()));
   }
 
   /**
--- a/glean/src/core/metrics/events_database.ts
+++ b/glean/src/core/metrics/events_database.ts
@@ -85,15 +85,16 @@ class EventsDatabase {
   /**
    * Records a given event.
    *
-   * @param metric The metric to record to.
+   * @param isDisabled Whether or not the metric is disabled.
+   * @param sendInPings The list of pings to record the event to.
    * @param value The value we want to record to the given metric.
    */
-  async record(metric: EventMetricType, value: RecordedEvent): Promise<void> {
-    if (metric.disabled) {
+  async record(isDisabled: boolean, sendInPings: string[], value: RecordedEvent): Promise<void> {
+    if (isDisabled) {
       return;
     }
 
-    for (const ping of metric.sendInPings) {
+    for (const ping of sendInPings) {
       const transformFn = (v?: JSONValue): JSONArray => {
         const existing: JSONArray = (v as JSONArray) ?? [];
         existing.push(RecordedEvent.toJSONObject(value));
--- a/glean/src/core/metrics/events_database.ts
+++ b/glean/src/core/metrics/events_database.ts
@@ -252,7 +252,7 @@ class EventsDatabase {
    *          `undefined` in case the ping doesn't contain any recorded metrics.
    */
   private getCustomPingEvents(pingData: RecordedEvent[], referenceTime: Date): JSONArray | undefined {
-    // Sort the events by their timestamp.
+    // Sort events by `gleanExecutionCounter`.
     const sortedData = pingData.sort((a, b) => {
       if (a.extra && b.extra) {
         // TODO bug 1693487: remove the stringification once the new events API
@@ -267,10 +267,10 @@ class EventsDatabase {
 
         // Intentionally fall-through in case the two execution counter values are equal.
         // It will perform a timestamp based comparision in that case.
+        // The timestamp check below takes care of any missing `gleanExecutionCounter`(s).
+        // While theoretically possible, it should never happen to have events without a
+        // corresponding `gleanExecutionCounter`.
       }
-
-      // While theoretically possible, it should never happen to have events without a
-      // corresponding gleanExecutionCounter.
       return a.timestamp - b.timestamp;
     });
 
--- a/glean/src/core/metrics/events_database.ts
+++ b/glean/src/core/metrics/events_database.ts
@@ -8,6 +8,12 @@ import { isUndefined } from "../utils.js";
 import type EventMetricType from "./types/event.js";
 import type { StorageBuilder } from "../../platform/index.js";
 import log, { LoggingLevel } from "../log.js";
+import CounterMetricType from "./types/counter.js";
+import { Context } from "../context.js";
+import { Lifetime } from "./lifetime.js";
+import { PING_INFO_STORAGE } from "../constants.js";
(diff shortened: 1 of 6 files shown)


glean/src/core/metrics/events_database.ts
```diff
@@ -135,32 +136,36 @@ class EventsDatabase {
     }
 
     for (const ping of sendInPings) {
```
brizental, line 133: Not possible to just filter like you do above?

glean/src/core/metrics/events_database.ts
```diff
@@ -247,15 +242,89 @@ class EventsDatabase {
...
+   *        timestamps will be computed relative to this time, for the given ping.
+   * @returns An object containing all the metrics recorded to the given ping,
+   *          `undefined` in case the ping doesn't contain any recorded metrics.
+   */
+  private getCustomPingEvents(pingData: RecordedEvent[], referenceTime: Date): JSONArray | undefined {
+    // Sort the events by their timestamp.
+    const sortedData = pingData.sort((a, b) => {
+      if (a.extra && b.extra) {
+        // TODO: Fixme, remove the stringification once the new events API
+        // is implemented.
```
brizental, line 259: nit: Bug number.

glean/src/core/metrics/events_database.ts
```diff
@@ -247,15 +242,89 @@ class EventsDatabase {
...
+   * Gets all of the persisted metrics related to a given custom ping.
+   *
+   * @param pingData A list of unordered, unprocessed recorded events.
+   * @param referenceTime The date/time to be considered as the "zero" time. All the events
+   *        timestamps will be computed relative to this time, for the given ping.
+   * @returns An object containing all the metrics recorded to the given ping,
+   *          `undefined` in case the ping doesn't contain any recorded metrics.
+   */
+  private getCustomPingEvents(pingData: RecordedEvent[], referenceTime: Date): JSONArray | undefined {
+    // Sort the events by their timestamp.
```
brizental, line 255: nit: `Sort events by gleanExecutionCounter.`

glean/src/core/metrics/events_database.ts
```diff
@@ -247,15 +242,89 @@ class EventsDatabase {
...
+        if(Number(a.extra["gleanExecutionCounter"]) > Number(b.extra["gleanExecutionCounter"])) {
+          return 1;
+        }
+
+        if(Number(a.extra["gleanExecutionCounter"]) < Number(b.extra["gleanExecutionCounter"])) {
+          return -1;
+        }
+
+        // Intentionally fall-through in case the two execution counter values are equal.
+        // It will perform a timestamp based comparision in that case.
```
brizental, line 269: nit: Combine this comment with the one below. The way they are separate right now is a bit confusing.

glean/src/core/metrics/events_database.ts
```diff
@@ -247,15 +242,89 @@ class EventsDatabase {
...
+      }
+
+      // While theoretically possible, it should never happen to have events without a
+      // corresponding gleanExecutionCounter.
+      return a.timestamp - b.timestamp;
+    });
+
+    // If the first event is not a 'glean.restarted' one, insert it and use the
+    // provided reference time as the glean restart time.
```
brizental, line 278: Not sure I understand why we need to inject `glean.restarted` as a first event as well. Wouldn't it only be necessary for subsequent runs?

brizental: Even more confusing now that I see it is remove from the events array again in line 293.

glean/src/core/metrics/events_database.ts
```diff
@@ -135,32 +136,36 @@ class EventsDatabase {
...
+        if (rawEventObject["extra"]) {
           // TODO: Fixme, remove the stringification once the new events API
           // is implemented.
-          "gleanExecutionCounter": currentExecutionCount.toString()
+          (rawEventObject["extra"] as JSONObject)["gleanExecutionCounter"] = currentExecutionCount.toString();
+        } else {
+          rawEventObject["extra"] =
+          {
+            // TODO: Fixme, remove the stringification once the new events API
+            // is implemented.
```
brizental, line 161: nit: Bug number.

glean/src/core/metrics/events_database.ts
Dexterp37: Note that we're also defining `rawEventObject` in this scope, which gets used down below, so we can't filter unless we refactor this a bit, but I'm now sure we'd gain much in readability

brizental: Ok, nevermind then.

glean/src/core/metrics/events_database.ts
Dexterp37: Depending on how initialization goes, a `glean.restarted` event may or may not be there already. In order for us to not have two different branches applying different logic, I thought it was easier to just inject a "fake" restart event and remove it later: we would always need the removal part anyway (see the proposal).

brizental: Ok, understood.

glean/src/core/metrics/events_database.ts
```diff
@@ -82,21 +108,85 @@ class EventsDatabase {
...
   }
 
+  async initialize(): Promise<void> {
+    const stores = await this.getAvailableStoreNames();
+    // Inject the glean.restarted event only in custom pings.
+    for (const storeName of stores.filter(n => n != "events")) {
+      const executionCounter = new CounterMetricType({
+        category: "",
+        name: `${storeName}#execution_counter`,
+        sendInPings: [PING_INFO_STORAGE],
```
Dexterp37, line 118: The problem with this approach is that, unlike what described in the proposal, the counter doesn't reset every time the ping is cleared. This is because we're using `PING_INFO_STORAGE`. The only other option I could think of is somehow add the concept of "metrics that would not be collected but are related to a ping. For example, for a given ping `test`, Glean would own a `glean.test` ping (the "glean" prefix would make it reserved). Both pings will be cleared when `test` is submitted. Any internal metric would be stored in `glean.test`. WDYT? @brizental @badboy ?

brizental: Hm, did you consider a simpler approach of adding a new property to a metric type, something line `doNotSend?: boolean`? With such an approach you could change the storage from `PING_INFO_STORAGE` to `storeName` here, which would ensure the metric to be cleared at the correct time. Then you could check for that property when [collecting metrics from the database](https://github.com/mozilla/glean.js/blob/main/glean/src/core/metrics/database.ts#L283) and not collect metrics that should not be added to the ping payload.

This would have the same result of "metrics that would not be collected but are related to a ping", bit might be simpler.

brizental: Ack, I see the issue already that when getting the metric from the database the `doNotSend` property would be lost.

brizental: Do we see this pattern being necessary in other places? We could have a blocklist based on metric identifier... That would be simpler. When gathering metrics we then check if the identifier of the metric matches against a pattern in the blocklist and if so we don't add it to the final metrics bundle.

Just trying to figure out a simpler way around this.

brizental: As discussed in https://github.com/mozilla/glean.js/pull/462#discussion_r657797861 the `#` character is not allowed by glean_parser, so if we matched agains a pattern like `(.*)#execution_counter$` that would never clash with a user defined metric.

brizental:
I wonder if c392fc517897cf93a3f9abdce66e32a09e755078 should be dropped based on the implemention note left by jan-erik on the proposal. WDYT @badboy ? 

Currently Glean.js does not send events pings, so that would not really have any effect.

badboy:
> I wonder if [c392fc5](https://github.com/mozilla/glean.js/commit/c392fc517897cf93a3f9abdce66e32a09e755078) should be dropped based on [the implemention note](https://docs.google.com/document/d/1R-tIZYz4Izx8rDB_YUiJXU9Xtk8VcJoF8Pmhggps_GU/edit?usp=sharing) left by jan-erik on the proposal. WDYT @badboy ?
> 
> Currently Glean.js does not send events pings, so that would not really have any effect.

Can you quote whatever implementation note you mean? That link points me ... not to anything specific.

brizental:
> Can you quote whatever implementation note you mean? That link points me ... not to anything specific.

Heh, sure. It's your comment down at the comments section:

> Implementation note: The logic will need to slightly differ between the builtin pings and custom pings.
However I think this might be implemented in an _okayish_ way: No special-casing of the glean.restarted stuff; if we handle the builtin ping first it will never hit the glean.restarted logic.

badboy:
> > Can you quote whatever implementation note you mean? That link points me ... not to anything specific.
> 
> Heh, sure. It's your comment down at the comments section:
> 
> > Implementation note: The logic will need to slightly differ between the builtin pings and custom pings.
> > However I think this might be implemented in an _okayish_ way: No special-casing of the glean.restarted stuff; if we handle the builtin ping first it will never hit the glean.restarted logic.

What's the current plan on sending the builtin events ping?
I'd say it's ok to leave out that commit if it's not applicable right now, but make an explicit note to bring it back if the events ping is added.

brizental:
> What's the current plan on sending the builtin events ping?

We discussed sending it when the threshold is hit and when Glean is started, but all of that is not final. A real proposal is still to come over on [Bug 1690253](https://bugzilla.mozilla.org/show_bug.cgi?id=1690253).

badboy (commented):
(Still sending some initial comments here for @brizental)

glean/src/core/metrics/events_database.ts
```diff
@@ -82,21 +108,85 @@ class EventsDatabase {
...
+        lifetime: Lifetime.Ping,
+        disabled: false
+      });
+
+      await CounterMetricType._private_addUndispatched(executionCounter, 1);
+
+      // Get the start date with a the minute resolution.
+      const startTime = DatetimeMetric.fromDate(Context.startTime, TimeUnit.Minute);
```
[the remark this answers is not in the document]
brizental: I do not see any reason why a higher precision would be problematic here. I will change this is in the current implementation and see if prbolems arise.

PR #462, opened 2021-06-23
Bug 1681483 - Implement persistent storage for QML

brizental:
See: https://glean-debug-view-dev-237806.firebaseapp.com/pings/brzntl-from-qml

The two pings with that tag were sent from different application runs 🎉 


glean/src/platform/qt/storage.ts
```diff
@@ -0,0 +1,198 @@
...
+
+    this.initialized = Promise.all(startQueries);
+  }
+
+  private _createKeyFromIndex(index: StorageIndex) {
+    return index.join(SEPARATOR);
+  }
+
+  private _getDbHandle(): LocalStorage.DatabaseHandle {
+    return LocalStorage.LocalStorage.openDatabaseSync(
```
brizental, line 127: At the danger of showing my database ignorance, question: would it be better if I called this function only once instead of everytime I execute the query?

brizental:
Converted to draft because somehow running the QML tests make it so that the WebExt storage tests are not run (and also don't error, thus the green CI). 

This PR is ready for review on the QML side though.

brizental:
Ok, the Storage/WebExt error is not related, and actually might be happening for a while /o\ It is also happening on the events threshold PR.

Filed a bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1717845

moz-glean:
# Build size report

Merging https://github.com/mozilla/glean.js/pull/462 into [main](https://github.com/mozilla/glean.js) will:
* **Increase** the size of full **Qt/QML** bundle build by `4%`.

---

|--:|:---:|:---:|:---:|
| **Web Extension** |
| core only | 45 KB | 45 KB | 📈 75 bytes |
| full bundle | 67 KB | 67 KB | 📈 75 bytes |
| **Qt/QML** |
| core only | 104 KB | 109 KB | 📈 4.4 KB |
| full bundle | 104 KB | 109 KB | 📈 4.4 KB |

glean/src/platform/qt/storage.ts
```diff
@@ -0,0 +1,198 @@
...
+    if (clear) {
+      startQueries.push(this._executeQuery(`DROP TABLE ${tableName};`));
+    }
+
+    startQueries.push(
+      // Initialize the database.
+      this._executeQuery(
+        `CREATE TABLE IF NOT EXISTS ${tableName}(key VARCHAR(255), value VARCHAR(255));`
+      ),
+      // This allows for `REPLACE ITEM` to work.
```
Dexterp37, line 114: Let's expand a bit: the "this" here is the unique index :)

glean/src/platform/qt/storage.ts
Dexterp37: From [Qt's sample](https://code.qt.io/cgit/qt/qtdeclarative.git/tree/examples/quick/localstorage/localstorage/Database.js?h=5.15#n63) it seems that it should be fine. The biggest I/O hit should be the first time this gets called, because the files get created.

Note that this [can throw](https://doc.qt.io/qt-5/qtquick-localstorage-qmlmodule.html#opendatabasesync): please handle accordingly.

glean/src/platform/qt/storage.ts
```diff
@@ -0,0 +1,198 @@
...
+    this.initialized = Promise.all(startQueries);
+  }
+
+  private _createKeyFromIndex(index: StorageIndex) {
+    return index.join(SEPARATOR);
+  }
+
+  private _getDbHandle(): LocalStorage.DatabaseHandle {
+    return LocalStorage.LocalStorage.openDatabaseSync(
+      this.name, "1.0", `${this.name} Storage`, 1000000
```
Dexterp37, line 128: Why did you pick this value as the estimated DB size?

glean/src/platform/qt/storage.ts
```diff
@@ -0,0 +1,198 @@
...
+    index: StorageIndex,
+    transformFn: (v?: JSONValue | undefined) => JSONValue
+  ): Promise<void> {
+    const result = (await this._getFullResultObject(index)) || {};
+    const transformedResult = updateNestedObject(result, index, transformFn);
+    const updates = getKeyValueArrayFromNestedObject(transformedResult);
+    for (const update of updates) {
+      const [ key, value ] = update;
+      await this._executeOnceInitialized(
+        `REPLACE INTO ${this.tableName}(key, value) VALUES('${key}', '${value.replace("'", "''")}');`
```
Dexterp37, line 185: Why can't we use `UPDATE` here? Any specific reason?

glean/src/platform/qt/storage.ts
```diff
@@ -0,0 +1,198 @@
...
+const LOG_TAG = "platform.qt.Storage";
+// The name of the file that will hold the SQLite database.
+const DATABASE_NAME = "Glean";
+// Since we cannot have nesting in SQL databases,
+// we will have a database with only two columns: `key` and `value`.
+// The `key` column will contain the StorageIndex as a string, joined by SEPARATOR.
+//
+// !IMPORTANT! This separator cannot be "." or "#" because these values
+// are already used as separators for label and category in other places of the code.
+const SEPARATOR = "+";
```
Dexterp37, line 21: Are you sure `#` can't be used? For labels we explicitly look at the last `#` because we know we could be using it :)

Dexterp37 (changes requested):

glean/src/platform/qt/storage.ts
brizental: Because I copy pasted: https://doc.qt.io/qt-5/qtquick-localstorage-qmlmodule.html

glean/src/platform/qt/storage.ts
brizental: Hm, no reason really. We can use update too.

glean/src/platform/qt/storage.ts
brizental: I did not test it, no. But if we did use `#`, then that would mean labeled stuff would be saved as:

```json
{
    "type": {
        "category.name": {
            "label": "value"
        }
    }
}
```

Instead of:

```json
{
    "type": {
        "category.name#label": "value"
    }
}
```

And that would cause problems.

brizental: (Wouldn't it? /me goes check)

glean/src/platform/qt/storage.ts
Dexterp37: I think we can inspect the average size of the Glean DB for the mobile and pick the double of the 90 percentile or the 99p as a good starting value :-)

glean/src/platform/qt/storage.ts
Dexterp37: I think using UPDATE would allow us to remove the unique value index

glean/src/platform/qt/storage.ts
Dexterp37: Wouldn't a more realistic index be something like `{ping}#{type}#{id}#{label}`?

Anyway, note that we already use `#` for names too. If we'll ever _split_ based on `#`, then using this as a separator would cause problems. Otherwise, I believe it should work. If we end up using `+` we need to make sure we won't allow it in metric names (I don't think we do, but we should make sure)

brizental: Ah! I remembered wrong and [we actually use `/` for labels](https://github.com/mozilla/glean.js/blob/caa10fe041666306f2fdae8dc3e5006f80159c47/glean/src/core/metrics/types/labeled.ts#L76). So this character cannot be used as a separator here either. You are right, we use `#` for names e.g. [on the sequence number metric](https://github.com/mozilla/glean.js/blob/caa10fe041666306f2fdae8dc3e5006f80159c47/glean/src/core/pings/maker.ts#L33). That can't be a separator either. Finally,  [we use `.` for joining category and name](https://github.com/mozilla/glean.js/blob/caa10fe041666306f2fdae8dc3e5006f80159c47/glean/src/core/metrics/index.ts#L76). That can't be a separator.

I checked, for categories glean_parser allows [dotted_snake_case](https://github.com/mozilla/glean_parser/blob/main/glean_parser/schemas/metrics.2-0-0.schema.yaml#L28), and metric names [snake_case](https://github.com/mozilla/glean_parser/blob/main/glean_parser/schemas/metrics.2-0-0.schema.yaml#L24). Neither of them allow the `+` character, so we should be fine unless we decide to go adding more special characters to metric names we create internally. If we do something that breaks this storage implementation test should fail though (once we have Qt integration tests, at least /o\).

Dexterp37: Since you did this nice and thorough investigation, would you mind adding it to the glean dev docs?

glean/src/platform/qt/storage.ts
```diff
@@ -0,0 +1,227 @@
...
+import type { JSONObject, JSONValue } from "../../core/utils.js";
+import { getValueFromNestedObject, updateNestedObject } from "../../core/storage/utils.js";
+import { isObject, isUndefined } from "../../core/utils.js";
+import log, { LoggingLevel } from "../../core/log.js";
+
+const LOG_TAG = "platform.qt.Storage";
+// The name of the file that will hold the SQLite database.
+const DATABASE_NAME = "Glean";
+// Estimated size of database file.
+const ESTIMATED_DATABASE_SIZE = 150 * 2 * 10**3; // 300Kb
```
Dexterp37, line 16: Can you write a comment documenting how you got to this value?

glean/src/platform/qt/storage.ts
```diff
@@ -0,0 +1,227 @@
...
+   *
+   * @returns The database handle or `undefined`.
+   */
+  private get _dbHandle(): LocalStorage.DatabaseHandle | undefined {
+    try {
+      const handle = LocalStorage.LocalStorage.openDatabaseSync(
+        this.name, "1.0", `${this.name} Storage`, ESTIMATED_DATABASE_SIZE
+      );
+
+      this.dbHandle = handle;
```
Dexterp37, line 138: Why are we storing this internally? Can't we only rely on the returned state?

brizental: This is just in case `openDatabaseSync` throws and then we can use the handle retrived on the previous call `dbHandle`. Not sure how common it is for this to throw, so just trying to do a best effort here.

glean/src/platform/qt/storage.ts
brizental: https://bugzilla.mozilla.org/show_bug.cgi?id=1718040

glean/src/platform/qt/storage.ts
Dexterp37: But as the docs state, the handle might be invalid (the handle gets closed when garbage collection kicks in). Moreover, the reason why opening fails might invalidate that handle, too. I'd really just rely on the currently active handle.

brizental: > I'd really just rely on the currently active handle.

That is what the code is doing right now. We always _attempt_ to get the handle when we call `this._dbHandle`. It's only in case that call fails that we fallback and use the handle we got from a previous call.

We can not do that, but then what should we do if we fail to get the handle?

Dexterp37: > We can not do that, but then what should we do if we fail to get the handle?

If we fail, I'm not sure getting a previous handle (which might be broken as well) is better. I would just do the same thing we do in glean-core when we fail with storage :)

Dexterp37 (approved):
r+, the handle thing can be done in a follow-up and reviewed by @badboy if needed.

glean/src/platform/qt/storage.ts
brizental: So, in glean-core, IIRC, [we bail initialization if we can't open the database](https://github.com/mozilla/glean/blob/1a8ad8003c220281e2f1cddc3b6aeabe7875cd35/glean-core/src/lib.rs#L271xf). If we are attempting to open it on every query execution that is not really possible. 

How about we return `undefined` if opening the database fails, instead of saving the handle and file a bug to record some metric on database errors, so we can understand better how many times does this error and then handle accordingly once we have that information?

brizental: Ok, just saw the approval, neverming the above comment. Filed a bug for further investigation. https://bugzilla.mozilla.org/show_bug.cgi?id=1718213

PR #480, opened 2021-07-01
RFC: Bug 1702208: Reuse WaitableHttpClient in schema tests for other test …

ChinYing-Li:
…files
Coming back a bit late, and here's the draft PR.
I didn't add much new stuff to `WaitableUploader`, of which the code is largely borrowed from the `waitableHttpClient`.
After looking into the places where `testBeforeNextSubmit`, I am not sure where's the pitfall of this - `testBeforeNextSubmit` allows the ping to assert the state or value of other variables in the test scope, and it looks fine to me.

Also, it seems like this [test case](https://github.com/mozilla/glean.js/blob/88a349c651e3c2a76dba00ebff458270fc7400f0/glean/tests/unit/core/pings/ping_type.spec.ts#L244), which is expected to fail, actually succeeded if I comment out [this line](https://github.com/mozilla/glean.js/blob/88a349c651e3c2a76dba00ebff458270fc7400f0/glean/tests/unit/core/pings/ping_type.spec.ts#L266).

It would be great if people can provide some feedback and what they would like to see.
Thank you!

* Changelog not modified, as this bug only concerns tests.

This PR does not include test(s) 

### DIFF
--- a/glean/src/platform/test/waitable_uploader.ts
+++ b/glean/src/platform/test/waitable_uploader.ts
@@ -1,0 +1,54 @@
+import Uploader, { UploadResult, UploadResultStatus } from "../../core/upload/uploader";
+import { JSONObject } from "../../core/utils";
+import { unzipPingPayload } from "../../../tests/utils";
+
+/**
+ * A Glean mock HTTP which allows one to wait for a specific ping submission.
+ */
+class WaitableUploader implements Uploader {
+  private waitingForName?: string;
+  private waitingForPath? : string;
+  private waitResolver?: (pingBody: JSONObject) => void;
+
+  /**
+   * Returns a promise that resolves once a ping is submitted or times out after a 2s wait.
+   *
+   * @param name The name of the ping to wait for.
+   * @param path
+   * @returns A promise that resolves once a ping is submitted or times out after a 2s wait.
+   */
+  waitForPingSubmission(name: string, path?: string): Promise<JSONObject> {
+    this.waitingForName = name;
+    this.waitingForPath = path;
+    return new Promise<JSONObject>((resolve, reject) => {
+      this.waitResolver = (pingBody: JSONObject) => {
+        this.waitingForName = undefined;
+        this.waitingForPath = undefined;
+        // Uncomment for debugging the ping payload.
+        // console.log(JSON.stringify(pingBody, null, 2));
+        resolve(pingBody);
+      };
+
+      setTimeout(() => reject(), 2000);
+    });
+  }
+
+  post(url: string, body: string): Promise<UploadResult> {
+    if (this.waitingForPath) {
+      if (url.includes(this.waitingForPath)) {
+        this.waitResolver?.(unzipPingPayload(body));
+      } else {
+        return Promise.reject(new Error('The submitted ping is not from the url we are waiting for.'));
+      }
+    } else if (this.waitingForName && url.includes(this.waitingForName)) {
+      this.waitResolver?.(unzipPingPayload(body));
+    }
+
+    return Promise.resolve({
+      result: UploadResultStatus.Success,
+      status: 200
+    });
+  }
+}
+
+export default WaitableUploader
--- a/glean/tests/unit/plugins/encryption.spec.ts
+++ b/glean/tests/unit/plugins/encryption.spec.ts
@@ -12,6 +12,7 @@ import Glean from "../../../src/core/glean";
 import PingType from "../../../src/core/pings/ping_type";
 import type { JSONObject } from "../../../src/core/utils";
 import TestPlatform from "../../../src/platform/test";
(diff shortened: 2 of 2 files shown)


brizental:
> Also, it seems like this test case, which is expected to fail, actually succeeded if I comment out this line.

Yeah, that is one of the points of that test. To showcase the issues with the `testBeforeNextSubmit` API. Notice that test is skipped.

glean/src/platform/test/waitable_uploader.ts
```diff
@@ -0,0 +1,54 @@
+import Uploader, { UploadResult, UploadResultStatus } from "../../core/upload/uploader";
```
brizental, line 1: nit: Can we add the WaitableUploader class in `test/utils.js` instead?

glean/tests/unit/plugins/encryption.spec.ts
```diff
@@ -36,6 +37,20 @@ describe("PingEncryptionPlugin", function() {
...
+    const pingId = "ident";
+    const ping = new PingType({
+      name: "test",
+      includeClientId: false,
+      sendIfEmpty: true,
+    });
+
+    const path = makePath(Context.applicationId, pingId, ping);
+    const mockUploader = new WaitableUploader();
+    const postSpy = sandbox.spy(mockUploader, "post");
```
brizental, line 49: There is no need to spy on the post function when using the `WaitableUploader`, the ping body is returned by `waitForPingSubmission`.

brizental (commented):
The approach here is correct. It is importat to remove the `spy` part, that is one of the points from this bug, to stop spying and relying on internal details of functions for assertions. That makes for fragile assertions :)

We should change to WaitableUploader as well on:

-  [`flipping upload enabled respects order of events`](https://github.com/mozilla/glean.js/blob/b08f0dc04a523c1776814ed4b8c3600069e12458/glean/tests/unit/core/glean.spec.ts#L237): Here we can use the waitable client to wait for the deletion-request ping instead of the whole arguments and spying approach;
- [`deletion request ping is sent when toggling upload status between runs`](https://github.com/mozilla/glean.js/blob/b08f0dc04a523c1776814ed4b8c3600069e12458/glean/tests/unit/core/glean.spec.ts#L278): Here we can use the waitable client to wait on the deletion-request ping as well, and avoid reaching for the `.currentJob` inside the uploader.

 I think that is it. There are some other places that mention using a custom uploader, e.g. [here](https://github.com/mozilla/glean.js/blob/02d44d9f1e0ac152cab80ba82e4910a7a9cf6784/glean/tests/unit/core/metrics/labeled.spec.ts#L58-L61) and [here](https://github.com/mozilla/glean.js/blob/02d44d9f1e0ac152cab80ba82e4910a7a9cf6784/glean/tests/unit/core/pings/ping_type.spec.ts#L24), but those places might be better served by just using the [internal sync submit function](https://github.com/mozilla/glean.js/blob/main/glean/src/core/pings/ping_type.ts#L101) that we added a while back. I'll send a related PR with that fix.

glean/src/platform/test/waitable_uploader.ts
ChinYing-Li: Just to confirm, do you mean `glean/tests/util.ts`?

brizental: Yes.

ChinYing-Li:
I wonder if we should also substitute `MockUploader` in "decrypting encrypted ping returns expected payload and protected headers" with `WaitableUploader`? Thanks!

brizental:
> I wonder if we should also substitute MockUploader in "decrypting encrypted ping returns expected payload and protected headers" with WaitableUploader? Thanks!

Yes, that does look like a great candidate for refactoring here :)

ChinYing-Li:
I couldn't immediately come up with a solution for refactoring "decrypting encrypted ping returns expected payload and protected headers", so I will just leave it alone for now. PR ready for review, thanks!

glean/tests/unit/core/glean.spec.ts
```diff
@@ -242,22 +243,23 @@ describe("Glean", function() {
...
+      true,
+      {
+        httpClient: mockUploader,
+      });
     // Immediatelly disable upload.
     Glean.setUploadEnabled(false);
     ping.submit();
 
     await Context.dispatcher.testBlockOnQueue();
     // TODO: Make this nicer once we resolve Bug 1691033 is resolved.
```
brizental, line 261: nit: This comment can be removed now :)

brizental (approved):
This look great. Thanks so much for the hard work investigating this. 

> I couldn't immediately come up with a solution for refactoring "decrypting encrypted ping returns expected payload and protected headers", so I will just leave it alone for now. PR ready for review, thanks!

That is no problem. Would you mind filing a bug so we remember to revisit this task in the future?

Let me know if you like to find a next bug to work on :)

PR #526, opened 2021-07-12
Bug 1716724 - Implement new logic to sort events from different application runs

brizental:

### DIFF
--- a/glean/src/core/metrics/events_database.ts
+++ b/glean/src/core/metrics/events_database.ts
@@ -4,9 +4,9 @@
 
 import type Store from "../storage/index.js";
 import type { JSONArray, JSONObject, JSONValue } from "../utils.js";
-import { isUndefined } from "../utils.js";
-import type EventMetricType from "./types/event.js";
 import type { StorageBuilder } from "../../platform/index.js";
+import { isUndefined } from "../utils.js";
+import EventMetricType from "./types/event.js";
 import log, { LoggingLevel } from "../log.js";
 import CounterMetricType from "./types/counter.js";
 import { Lifetime } from "./lifetime.js";
@@ -16,8 +16,10 @@ import { generateReservedMetricIdentifiers } from "./database.js";
 const LOG_TAG = "core.Metric.EventsDatabase";
 
 export const GLEAN_EXECUTION_COUNTER_EXTRA_KEY = "glean_execution_counter";
+const GLEAN_STARTUP_DATE_EXTRA_KEY = "glean_startup_date";
 const GLEAN_RESERVED_EXTRA_KEYS = [
   GLEAN_EXECUTION_COUNTER_EXTRA_KEY,
+  GLEAN_STARTUP_DATE_EXTRA_KEY
 ];
 
 /**
@@ -33,6 +35,22 @@ function getExecutionCounterMetric(sendInPings: string[]): CounterMetricType {
     lifetime: Lifetime.Ping,
     disabled: false
   });
+}
+
+/**
+ * Creates an `glean.restarted` event metric.
+ *
+ * @param sendInPings The list of pings this metric is sent in.
+ * @returns A metric type instance.
+ */
+function getGleanRestartedEvent(sendInPings: string[]): EventMetricType {
+  return new EventMetricType({
+    category: "glean",
+    name: "restarted",
+    sendInPings: sendInPings,
+    lifetime: Lifetime.Ping,
+    disabled: false
+  }, [ GLEAN_STARTUP_DATE_EXTRA_KEY ]);
 }
 
 // An helper type for the 'extra' map.
@@ -160,7 +178,17 @@ class EventsDatabase {
 
     const storeNames = await this.getAvailableStoreNames();
     // Increment the execution counter for known stores.
+    // !IMPORTANT! This must happen before any event is recorded for this run.
     await CounterMetricType._private_addUndispatched(getExecutionCounterMetric(storeNames), 1);
+
+    // Record the `glean.restarted` event.
+    await EventMetricType._private_recordUndispatched(
+      getGleanRestartedEvent(storeNames),
+      {
+        // TODO (bug 1693487): Remove the stringification once the new events API is implemented.
+        [GLEAN_STARTUP_DATE_EXTRA_KEY]: Context.startTime.toISOString()
+      }
+    );
 
     this.initialized = true;
   }
--- a/glean/src/core/metrics/events_database.ts
+++ b/glean/src/core/metrics/events_database.ts
@@ -383,19 +383,41 @@ class EventsDatabase {
         const dateOffset = nextRestartDate.getTime() - lastRestartDate.getTime();
         lastRestartDate = nextRestartDate;
 
-        // Update the current offset and move to the next event.
-        restartedOffset += dateOffset;
+        // Calculate the new offset since new restart.
+        const newRestartedOffset = restartedOffset + dateOffset;
+        if (newRestartedOffset < 0) {
+          // In case the new offset is negative, just increase the previous timestamp by one
+          // to make sure timestamps keep increasing.
+          //
+          // TODO: Record an error when this happens. !SELF_REMINDER BEFORE MERGING! File a bug.
+          restartedOffset = sortedEvents[index - 1].timestamp + 1;
+        } else {
+          restartedOffset = newRestartedOffset;
+        }
       } catch {
         // Do nothing,
         // this is expected to fail in case the current event is not a `glean.restarted` event.
       }
 
-      // Update the timestamp for the current event,
-      // to account for the computed offset.
+      // Apply necessary offsets to timestamps:
+      // 1. If it is the first execution, subtract the firstEventOffset;
+      // 2. Otherwise add restartedOffset.
+
+      // The execution counter is a counter metric, the smallest value it can have is `1`.
+      // At this stage all metrics should have an execution counter, but out of caution we
+      // will fallback to `1` in case it is not present.
+      const executionCount = Number(event.extra?.[GLEAN_EXECUTION_COUNTER_EXTRA_KEY] || 1);
+      let adjustedTimestamp: number;
+      if (executionCount === 1) {
+        adjustedTimestamp = event.timestamp - firstEventOffset;
+      } else {
+        adjustedTimestamp = event.timestamp + restartedOffset;
+      }
+
       sortedEvents[index] = new RecordedEvent(
         event.category,
         event.name,
-        event.timestamp + restartedOffset - firstEventOffset,
+        adjustedTimestamp,
         event.extra
       );
     });
--- a/glean/src/core/metrics/events_database.ts
+++ b/glean/src/core/metrics/events_database.ts
@@ -23,6 +23,23 @@ const GLEAN_RESERVED_EXTRA_KEYS = [
 ];
 
 /**
+ * Attempts to create a date object from a string.
+ *
+ * Throws if unsuccesfull.
+ *
+ * @param str The string to generate a date from.
+ * @returns The Date object created.
+ */
+function createDateObject(str = ""): Date {
+  const date = new Date(str);
+  // Date object will not throw errors when we attempt to create an invalid date.
+  if (isNaN(date.getTime())) {
+    throw new Error(`Error attempting to generate Date object from string: ${str}`);
+  }
+  return date;
+}
+
+/**
  * Creates an execution counter metric.
  *
  * @param sendInPings The list of pings this metric is sent in.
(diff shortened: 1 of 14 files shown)


glean/src/metrics.yaml
```diff
@@ -373,3 +375,40 @@ glean.error:
...
+
+          **This extra key is recorded by Glean internaly, but is never sent in pings.**
+        type: string
+      glean_execution_counter:
+        description: |
+          This is a special key holding the number of times Glean was
+          initialized up to the moment this event gets injected.
+
+          **This extra key is recorded by Glean internaly, but is never sent in pings.**
+        type: quantity
```
brizental, line 414: I left these here, but I kinda think we should not add them. These are internal Glean.js details, don't need to be called out here. WDYT?

badboy (changes requested):
The code looks good, I left a couple of minor nits throughout the files (I hope they show up correctly).

I do have one request though before greenlighting that:
Can we add some more explicit tests for the timings of events?
I know in the proposal we had several scenarios sketched. We could reproduce those in the tests with explicitly chosen timestamps and then validate that the ping payload has the right times.
Right now we check that the timestamps are monotonically increasing, but don't check that the offsets are what we expect.

glean/src/core/metrics/events_database.ts
```diff
@@ -22,6 +22,23 @@ const GLEAN_RESERVED_EXTRA_KEYS = [
...
 
+/**
+ * Attempts to create a date object from a string.
+ *
+ * Throws if unsuccesfull.
+ *
+ * @param str The string to generate a date from.
+ * @returns The Date object created.
+ */
+function createDateObject(str = ""): Date {
```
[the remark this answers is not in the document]
brizental: Because in the places where we call this function, the value we pass to it might be undefined. In order not to have to add a default everytime I call it, I decided to just add a default here.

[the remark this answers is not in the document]
badboy: Wait, if you pass undefined _explicitly_ it will fall back to the default value?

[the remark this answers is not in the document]
brizental: Heh, yeap.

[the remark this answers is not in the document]
badboy: Iiiiinteresting

glean/src/core/metrics/events_database/index.ts
```diff
@@ -0,0 +1,356 @@
...
+      //
+      // 1. The ping was already sent during this session and the events storare was cleared;
+      // 2. No event has ever been recorded for this ping.
+      if (!currentExecutionCount) {
+        await CounterMetricType._private_addUndispatched(executionCounter, 1);
+        currentExecutionCount = 1;
+
+        // Record the `glean.restarted` event,
+        // this must **always** be the first event of any events list.
+        await recordGleanRestartedEvent([ping]);
```
brizental, line 158: I debated not using `Context.startTime` when injecting the `restarted` event on the first recording for a given ping and instead doing `new Date()`. This way timestamp offsets on different sessions would be calculated based on the time of the recording of the first event, which is what we do for events before restarts.

I thought that would make it ~even more~ confusing in the code though, so I left it like this.

glean/src/core/metrics/events_database/index.ts
```diff
@@ -0,0 +1,356 @@
...
+
+      // Sort by timestamp if events come from same execution.
+      return a.timestamp - b.timestamp;
+    });
+
+    let lastRestartDate: Date;
+    try {
+      lastRestartDate = createDateObject(sortedEvents[0].extra?.[GLEAN_STARTUP_DATE_EXTRA_KEY]);
+      // Drop the first `restarted` event.
+      sortedEvents.shift();
```
brizental, line 286: Note: We only drop the first `glean.restarted` event, so it is possible to have stray `glean.restarted` events at the end of the `events` array. I wonder if it would be best to drop stray `restarts`. That can be done as follow-up though.

glean/src/core/metrics/events_database/index.ts
```diff
@@ -146,12 +145,17 @@ class EventsDatabase {
       const executionCounter = getExecutionCounterMetric([ping]);
 
       let currentExecutionCount = await Context.metricsDatabase.getMetric(ping, executionCounter);
-      // There might not be an execution counter stored
-      // in case the ping was already sent during this session,
-      // because in this case the ping storage will have been cleared.
+      // There might not be an execution counter stored in case:
+      //
+      // 1. The ping was already sent during this session and the events storare was cleared;
```
badboy, line 150:
```suggestion
      // 1. The ping was already sent during this session and the events storage was cleared;
```

glean/tests/unit/core/metrics/events_database.spec.ts
```diff
@@ -470,4 +470,50 @@ describe("EventsDatabase", function() {
       assert.strictEqual("restarted", (snapshot[(i * 2) + 1] as JSONObject)["name"]);
     }
   });
+
+  it("events are correctly sorted if time decides to stand still throughout restarts", async function() {
```
badboy, line 475: Remind me: is the `glean.restarted` timer trimmed to minutes or is it to second precision?
If it's minute, this might occur easily because of using the app again within a minute, right?
So regardless second precision sounds more reasonable to me, unless there was some other reason

glean/tests/unit/core/metrics/events_database.spec.ts
```diff
@@ -516,4 +517,180 @@ describe("EventsDatabase", function() {
...
+    );
+
+    assert.ok(payload);
+    const [
+      firstEvent,
+      secondEvent,
+      restartedEvent,
+      thirdEvent,
+      fourthEvent
+    ] = payload.events as JSONObject[];
```
badboy, line 570: Would this fail if there are more or less events in `payload.events`?
If not please add an explicit assert for the length.

badboy (approved):
This looks pretty good now. Thanks for adding the additional tests.

Let's land it and take it to production!

glean/tests/unit/core/metrics/events_database.spec.ts
brizental: It's in milliseconds, actually. We just `toISOString` a `Date` object. This way it's easy to convert it back to a `Date` object from the string.

glean/tests/unit/core/metrics/events_database.spec.ts
brizental: If there are less, yes, this would fail. If there are more it depends, if the events are just appended to the env of the array this would pass, otherwise no.

glean/src/core/metrics/events_database/index.ts
badboy: Hm, code-wise it could be as easy as passing `new Date()` here as a second argument and make `recordGleanRestartedEvent` take that as optional and use `Context.startTime` as the default. Not too complicated.

I _think_ this would make more sense.

Consider this:

1. App started. Idle for 10 minutes.
2. 1 Event (`a`) recorded. Initial `glean.restarted` added.
3. App stopped. Not touched for 10 minutes.
4. App started. `glean.restarted` inserted. Idle for 1 minute.
5. 1 Event (`b`) recorded
6. Ping submitted.

If using `Contex.startTime` then we have these events with timestamps (in minutes):

1. `a, ts=0`
2. `glean.restarted, ts=20` (10 minutes from initial start + 10 minutes stopped)
3. `b, ts=21`

With `new Date` we get:

1. `a, ts=0`
2. `glean.restarted, ts=10` (10 minutes stopped)
3. `b, ts=11`

First: is this example right? Please check if my thoughts make sense.
Second: 11 minutes difference between event `a` and event `b` is the "correct" thing here.

glean/src/core/metrics/events_database/index.ts
badboy: yeah, do a follow-up.

glean/src/core/metrics/events_database/index.ts
brizental: Yes, the code would be straight forward. What I think would make it confusing is having the `GLEAN_STARTUP_DATE_EXTRA_KEY`, but then record an arbitrary time in it. That makes it so that it is not really the `GLEAN_STARTUP_DATE_EXTRA_KEY` anymore, its more of a `GLEAN_REFERENCE_POINT_EXTRA_KEY`.

The first `glean.restarted` event is always trimmed from the payload, so that is more of an internal confusion. That would be less of a big deal if this implementation was not already super confusing.

Also, yes. Example is exactly what I had in mind :)

brizental:
# DATA REVIEW REQUEST

1. What questions will you answer with this data?

Was there an application restart between recorded events in a ping?

2. Why does Mozilla need to answer these questions? Are there benefits for users?
   Do we need this information to address product or business requirements?

Knowing when a restart happens in between events in a ping is useful
for sessionization of events at analysis time.

3. What alternative methods did you consider to answer these questions?
   Why were they not sufficient?

The current behaviour of the Glean SDKs is to send any ping that contains
events at startup. That is fine for the Glean-internal `events` ping, but is
unwanted behaviour for user defined pings, aka custom pings.

That is the current alternative. The `glean.restarted` event is necessary
in order to drop it, but retain the knowledge regarding whether there was a
restart between events. Additionally this alternative makes it more explicit
that a restart happened, while the current approach leaves it for the analyzer
to guess from shape of pings / submission_timestamps / other metrics
that a restart did happen.

See also: https://docs.google.com/document/d/1R-tIZYz4Izx8rDB_YUiJXU9Xtk8VcJoF8Pmhggps_GU/

4. Can current instrumentation answer these questions?

Sort of[^1] and at the cost of submitting custom pings at startup time.

[^1]: The current instrumentation does not _reliably_ answer the question about
whether a restart happened or not. For `events` pings, where the `restarted` ping reason
can be consulted to make sure the ping was sent due to a restart, yes it does. However,
for custom pings guess work is necessary to figure out that there was a restart,
since the `restarted` reason is dropped before ping submission, unless it is defined
as a known reason for the ping.

5. List all proposed measurements and indicate the category of data collection for each
   measurement, using the Firefox data collection categories found on the Mozilla wiki.

Measurement Name | Measurement Description | Data Collection Category | Tracking Bug
---------------- | ----------------------- | ------------------------ | ------------
glean.restarted | Event that signals an application restart. | 2 | [1716724](https://bugzilla.mozilla.org/show_bug.cgi?id=1716724)

6. Please provide a link to the documentation for this data collection which
   describes the ultimate data set in a public, complete, and accurate way.

This collection is Glean so is documented
[in the Glean Dictionary](https://dictionary.telemetry.mozilla.org).

7. How long will this data be collected?

This collection never expires.

8. What populations will you measure?

All channels, countries, and locales. No filters.

9. If this data collection is default on, what is the opt-out mechanism for users?

These collections are Glean. The opt-out can be found in the product's preferences.

10. Please provide a general description of how you will analyze this data.

Users analyzing events in custom pings will analyze this metric in a case-by-case basis.

11. Where do you intend to share the results of your analysis?

N/A

12. Is there a third-party tool (i.e. not Telemetry) that you
    are proposing to use for this data collection?

No.

brizental:
Hey @chutten pinging you here specifically for the data review :) Thanks!

glean/src/core/metrics/events_database/index.ts
brizental: https://bugzilla.mozilla.org/show_bug.cgi?id=1723090

chutten:
@brizental Are you the individual responsible for the data collection over its non-expiring lifetime?

brizental:
> @brizental Are you the individual responsible for the data collection over its non-expiring lifetime?

Yes. Me, falling back to the Glean team if I am not available.

chutten:
DATA COLLECTION REVIEW RESPONSE:

    Is there or will there be documentation that describes the schema for the ultimate data set available publicly, complete and accurate?

Yes.

    Is there a control mechanism that allows the user to turn the data collection on and off?

Yes. This collection is Telemetry so can be controlled through Firefox's Preferences.

    If the request is for permanent data collection, is there someone who will monitor the data over time?

Yes, Bea Rizental is responsible.

    Using the category system of data types on the Mozilla wiki, what collection type of data do the requested measurements fall under?

Category 1, Technical.

    Is the data collection request for default-on or default-off?

Default on for all channels.

    Does the instrumentation include the addition of any new identifiers?

No.

    Is the data collection covered by the existing Firefox privacy notice?

Yes.

    Does the data collection use a third-party collection tool?

No.

---
Result: datareview+

chutten (approved):
(data-r+ specifically)

glean/src/core/metrics/events_database/index.ts
brizental: 11caecc

moz-glean:
# Build size report

Merging https://github.com/mozilla/glean.js/pull/526 into [main](https://github.com/mozilla/glean.js) will:
* **Increase** the size of full **Web Extension** bundle build by `3%`.
* **Increase** the size of full **Qt/QML** bundle build by `3%`.

---

|--:|:---:|:---:|:---:|
| **Web Extension** |
| core only | 46 KB | 50 KB | 📈 3.1 KB |
| full bundle | 71 KB | 73 KB | 📈 2.4 KB |
| **Qt/QML** |
| core only | 115 KB | 119 KB | 📈 3.6 KB |
| full bundle | 115 KB | 119 KB | 📈 3.6 KB |

PR #614, opened 2021-08-07
Bug 1716956: Initial attempt to implement metric type "StringList"

ChinYing-Li:

~~Currently one unit test case and schema test fails. Opening a draft and is not ready for review.~~
EDIT: PR now ready for review; For `StringMetricType`, I have to declare `add` as async.

### DIFF
--- a/glean/src/core/metrics/types/string_list.ts
+++ b/glean/src/core/metrics/types/string_list.ts
@@ -26,6 +26,7 @@ export class StringListMetric extends Metric<string[], string[]> {
     if (v.length > MAX_LIST_LENGTH) {
       return false;
     }
+
     for (const s of v) {
       if (!isString(s) || s.length > MAX_STRING_LENGTH) {
         return false;
@@ -57,18 +58,19 @@ class StringListMetricType extends MetricType {
    *
    * # Note
    *
-   * Truncates the list if it is longer than `MAX_LIST_LENGTH` and logs an error.
+   * Truncates the list if it is longer than `MAX_LIST_LENGTH` and records an error.
    *
-   * Truncates the value if it is longer than `MAX_STRING_LENGTH` bytes
-   * and logs an error.
+   * Truncates the value if it is longer than `MAX_STRING_LENGTH` characters
+   * and records an error.
    *
-   * @param value the list of string to set the metric to.
+   * @param value The list of strings to set the metric to.
    */
   set(value: string[]): void {
     Context.dispatcher.launch(async () => {
       if (!this.shouldRecord(Context.uploadEnabled)) {
         return;
       }
+
       const truncatedList: string[] = [];
       if (value.length > MAX_LIST_LENGTH) {
         await Context.errorManager.record(
--- a/glean/src/core/metrics/types/string_list.ts
+++ b/glean/src/core/metrics/types/string_list.ts
@@ -46,7 +46,7 @@ export class StringListMetric extends Metric<string[], string[]> {
  *
  * This allows appending a string value with arbitrary content to a list.
  * The list is length-limited to `MAX_LIST_LENGTH`.
- * Strings are length-limited to `MAX_STRING_LENGTH` bytes.
+ * Strings are length-limited to `MAX_STRING_LENGTH` characters.
  */
 class StringListMetricType extends MetricType {
   constructor(meta: CommonMetricData) {
@@ -94,9 +94,9 @@ class StringListMetricType extends MetricType {
    *
    * # Note
    *
-   * - If the list is already of length `MAX_LIST_LENGTH`, log an error.
-   * - Truncates the value if it is longer than `MAX_STRING_LENGTH` bytes
-   * and logs an error.
+   * - If the list is already of length `MAX_LIST_LENGTH`, record an error.
+   * - Truncates the value if it is longer than `MAX_STRING_LENGTH` characters
+   * and records an error.
    *
    * @param value The string to add.
    */
@@ -107,7 +107,7 @@ class StringListMetricType extends MetricType {
       }
 
       const truncatedValue = await truncateStringAtBoundaryWithError(this, value, MAX_STRING_LENGTH);
-      let currentLen = -1;
+      let currentLen = 0;
 
       const transformFn = ((value) => {
         return (v?: JSONValue): StringListMetric => {
@@ -131,7 +131,7 @@ class StringListMetricType extends MetricType {
 
       await Context.metricsDatabase.transform(this, transformFn);
 
-      if (currentLen == MAX_LIST_LENGTH) {
+      if (currentLen >= MAX_LIST_LENGTH) {
         await Context.errorManager.record(
           this,
           ErrorType.InvalidValue,
--- a/glean/src/core/metrics/types/string_list.ts
+++ b/glean/src/core/metrics/types/string_list.ts
@@ -6,7 +6,8 @@ import type { CommonMetricData } from "../index.js";
 import { MetricType } from "../index.js";
 import { Context } from "../../context.js";
 import { Metric } from "../metric.js";
-import { isString, JSONValue, truncateStringAtBoundaryWithError } from "../../utils.js";
+import { isString, truncateStringAtBoundaryWithError } from "../../utils.js";
+import type { JSONValue } from "../../utils.js";
 import { ErrorType } from "../../error/error_type.js";
 
 export const MAX_LIST_LENGTH = 20;
@@ -79,7 +80,7 @@ class StringListMetricType extends MetricType {
 
       for (let i = 0; i < Math.min(value.length, MAX_LIST_LENGTH); ++i) {
         const truncatedString = await truncateStringAtBoundaryWithError(this, value[i], MAX_STRING_LENGTH);
-        truncatedList.push(truncatedString)
+        truncatedList.push(truncatedString);
       }
       const metric = new StringListMetric(truncatedList);
       await Context.metricsDatabase.record(this, metric);
@@ -104,6 +105,7 @@ class StringListMetricType extends MetricType {
       }
 
       const truncatedValue = await truncateStringAtBoundaryWithError(this, value, MAX_STRING_LENGTH);
+      let currentLen = -1;
 
       const transformFn = ((value) => {
         return (v?: JSONValue): StringListMetric => {
@@ -112,26 +114,28 @@ class StringListMetricType extends MetricType {
           try {
             metric = new StringListMetric(v);
             result = metric.get();
+            currentLen = result.length;
             if (result.length < MAX_LIST_LENGTH) {
               result.push(value);
-            } else {
-              Context.errorManager.record(
-                this,
-                ErrorType.InvalidValue,
-                `String list length of ${result.length}+1 exceeds maximum of ${MAX_LIST_LENGTH}.`
-              );
             }
           } catch {
             metric = new StringListMetric([value]);
             result = [value];
           }
-
           metric.set(result);
-          return metric
-        }
+          return metric;
+        };
       })(truncatedValue);
 
       await Context.metricsDatabase.transform(this, transformFn);
+
+      if (currentLen == MAX_LIST_LENGTH) {
+        await Context.errorManager.record(
+          this,
+          ErrorType.InvalidValue,
+          `String list length of ${currentLen}+1 exceeds maximum of ${MAX_LIST_LENGTH}.`
+        );
+      }
     });
   }
 
--- a/glean/src/core/metrics/types/string_list.ts
+++ b/glean/src/core/metrics/types/string_list.ts
@@ -135,7 +135,7 @@ class StringListMetricType extends MetricType {
         await Context.errorManager.record(
           this,
           ErrorType.InvalidValue,
-          `String list length of ${currentLen}+1 exceeds maximum of ${MAX_LIST_LENGTH}.`
+          `String list length of ${currentLen+1} exceeds maximum of ${MAX_LIST_LENGTH}.`
         );
       }
     });
--- a/glean/src/core/metrics/types/string_list.ts
+++ b/glean/src/core/metrics/types/string_list.ts
@@ -1,0 +1,160 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import type { CommonMetricData } from "../index.js";
+import { MetricType } from "../index.js";
+import { Context } from "../../context.js";
+import { Metric } from "../metric.js";
+import { isString, JSONValue, truncateStringAtBoundaryWithError } from "../../utils.js";
+import { ErrorType } from "../../error/error_type.js";
+
+export const MAX_LIST_LENGTH = 20;
+export const MAX_STRING_LENGTH = 50;
+
+export class StringListMetric extends Metric<string[], string[]> {
+  constructor(v: unknown) {
+    super(v);
+  }
+
+  validate(v: unknown): v is string[] {
+    if (!Array.isArray(v)) {
+      return false;
+    }
+
+    if (v.length > MAX_LIST_LENGTH) {
+      return false;
+    }
+    for (const s of v) {
+      if (!isString(s) || s.length > MAX_STRING_LENGTH) {
+        return false;
+      }
+    }
+
+    return true;
+  }
+
+  payload(): string[] {
+    return this._inner;
+  }
+}
+
+/**
+ * A string list metric.
+ *
+ * This allows appending a string value with arbitrary content to a list.
+ * The list is length-limited to `MAX_LIST_LENGTH`.
+ * Strings are length-limited to `MAX_STRING_LENGTH` bytes.
+ */
+class StringListMetricType extends MetricType {
+  constructor(meta: CommonMetricData) {
+    super("string_list", meta);
+  }
+
+  /**
+   * Sets to the specified string list value.
+   *
+   * # Note
+   *
+   * Truncates the list if it is longer than `MAX_LIST_LENGTH` and logs an error.
+   *
+   * Truncates the value if it is longer than `MAX_STRING_LENGTH` bytes
+   * and logs an error.
+   *
+   * @param value the list of string to set the metric to.
+   */
+  set(value: string[]): void {
+    Context.dispatcher.launch(async () => {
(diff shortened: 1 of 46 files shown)


brizental:
Hey @ChinYing-Li , just FYI I am on PTO this week and will come back to review this PR on Monday! Thank you for seding this in and sorry for the delayed response 👋

glean/src/core/metrics/types/string_list.ts
```diff
@@ -0,0 +1,207 @@
...
+
+  validate(v: unknown): v is string[] {
+    if (!Array.isArray(v)) {
+      return false;
+    }
+
+    if (v.length > MAX_LIST_LENGTH) {
+      return false;
+    }
+    for (const s in v) {
```
brizental, line 29: This should be a  [for...of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) loop, otherwise `s` is the indexes in `v`, not the values.

glean/src/core/metrics/types/string_list.ts
```diff
@@ -0,0 +1,207 @@
...
+   *
+   * This is absolutely not meant to be used outside of Glean itself.
+   * It may cause multiple issues because it cannot guarantee
+   * that the recording of the metric will happen in order with other Glean API calls.
+   *
+   * @param instance The metric instance to record to.
+   * @param value The string we want to add
+   *
+   */
+  static async _private_addUndispatched(instance: StringListMetricType, value?: string): Promise<void> {
```
brizental, line 127: The undispatched methods are only necessary for metric types that will be recorded internally by Glean. That is not the case for this metric type. In this case, please merge this method with the `add`. Same with `set`.

glean/src/core/metrics/types/string_list.ts
```diff
@@ -0,0 +1,207 @@
...
+ *
+ * This allows appending a string value with arbitrary content to a list.
+ * The list is length-limited to `MAX_LIST_LENGTH`.
+ * Strings are length-limited to `MAX_STRING_LENGTH` bytes.
+ */
+class StringListMetricType extends MetricType {
+  length: number;
+
+  constructor(meta: CommonMetricData, length: number) {
+    super("stringlist", meta);
```
brizental, line 58:
```suggestion
    super("string_list", meta);
```

glean/src/core/metrics/types/string_list.ts
```diff
@@ -0,0 +1,207 @@
...
+
+/**
+ * A string list metric.
+ *
+ * This allows appending a string value with arbitrary content to a list.
+ * The list is length-limited to `MAX_LIST_LENGTH`.
+ * Strings are length-limited to `MAX_STRING_LENGTH` bytes.
+ */
+class StringListMetricType extends MetricType {
+  length: number;
```
brizental, line 55: Why is this a property on the metric type itself?

glean/src/core/metrics/types/string_list.ts
```diff
@@ -0,0 +1,207 @@
...
+    const truncatedList: string[] = [];
+    if (value.length > MAX_LIST_LENGTH) {
+      await Context.errorManager.record(
+        instance,
+        ErrorType.InvalidValue,
+        `String list length of ${value.length} exceeds maximum of ${MAX_LIST_LENGTH}.`
+      );
+    }
+
+    let i = 0;
```
brizental, line 87: nit: This variable can be declared inside the loop.

glean/src/core/metrics/types/string_list.ts
```diff
@@ -0,0 +1,207 @@
...
+   * that the recording of the metric will happen in order with other Glean API calls.
+   *
+   * @param instance The metric instance to record to.
+   * @param value The string we want to add.
+   */
+  static async _private_setUndispatched(instance: StringListMetricType, value: string[]): Promise<void> {
+    if (!instance.shouldRecord(Context.uploadEnabled)) {
+      return;
+    }
+    const truncatedList: string[] = [];
```
brizental, line 78: nit: Let's declare this variable right before the loop below, so it's closer to the place where it is used.

glean/src/core/metrics/types/string_list.ts
```diff
@@ -0,0 +1,207 @@
...
+    if (!instance.shouldRecord(Context.uploadEnabled)) {
+      return;
+    }
+
+    if (isUndefined(value)) {
+      return;
+    }
+
+    const truncatedValue = await truncateStringAtBoundaryWithError(instance, value, MAX_LENGTH_VALUE);
+    console.log(`Length of string list: ${this.length}`)
```
brizental, line 137: Letf over `console.log`?

glean/src/core/metrics/types/string_list.ts
```diff
@@ -0,0 +1,207 @@
...
+   * @param instance The metric instance to record to.
+   * @param value The string we want to add
+   *
+   */
+  static async _private_addUndispatched(instance: StringListMetricType, value?: string): Promise<void> {
+    if (!instance.shouldRecord(Context.uploadEnabled)) {
+      return;
+    }
+
+    if (isUndefined(value)) {
```
brizental, line 132: Let's make `value` a required argument so that it is not necessary to have this check here.

glean/src/core/metrics/utils.ts
```diff
@@ -30,6 +31,7 @@ const METRIC_MAP: {
   "labeled_string": LabeledMetric,
   "quantity": QuantityMetric,
   "string": StringMetric,
+  "stringlist": StringListMetric,
```
brizental, line 34:
```suggestion
  "string_list": StringListMetric,
```

glean/tests/unit/core/metrics/stringlist.spec.ts
```diff
@@ -0,0 +1,181 @@
...
+      await metric.testGetValue("aPing"),
+      expectedList
+    );
+
+    assert.strictEqual(
+      await metric.testGetNumRecordedErrors(ErrorType.InvalidOverflow), 1
+    );
+  });
+
+  it("attempt to add string to string list of maximum length logs an error", async function() {
```
brizental, line 142:
```suggestion
  it("attempt to add string to string list of maximum length records an error", async function() {
```

glean/tests/unit/core/metrics/stringlist.spec.ts
```diff
@@ -0,0 +1,181 @@
...
+    assert.deepStrictEqual(
+      await metric.testGetValue("aPing"),
+      testStringList
+    );
+
+    assert.strictEqual(
+      await metric.testGetNumRecordedErrors(ErrorType.InvalidValue), 1
+    );
+  });
+})
```
brizental, line 181: Make sure all the test cases covered on glean-core https://github.com/mozilla/glean/blob/3d0b2cc2ea0ce696e92cc8b30cb6f69d9f9a1056/glean-core/tests/string_list.rs are also covered here. If I understand correct, the only one missing here is `set_does_not_record_error_when_receiving_empty_list`.

glean/src/core/metrics/types/string_list.ts
```diff
@@ -0,0 +1,207 @@
...
+   *
+   * # Important
+   *
+   * This is absolutely not meant to be used outside of Glean itself.
+   * It may cause multiple issues because it cannot guarantee
+   * that the recording of the metric will happen in order with other Glean API calls.
+   *
+   * @param instance The metric instance to record to.
+   * @param value The string we want to add.
+   */
```
brizental, line 73: Reminder to update all the function doc strings before taking this PR out of draft mode :)

brizental (commented):
Great work here @ChinYing-Li ! This is definitely on the right path, nothing major is wrong and the apporach is correct. Left a few comments, let me know if you have any questions.

brizental:
> Currently one unit test case and schema test fails. Opening a draft and is not ready for review.

Heh, just saw this comment. Left a review anyways, hope it is helpful.

Regarding the schema test, it's possible it is failing because of the `stringlist` -> `string_list` change I requested in the review. [The Glean schema](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/18ae30533ca6b7235725c99fb272362b0dcee307/schemas/glean/glean/glean.1.schema.json#L360-L373) is expecting the payload to contain `string_list` with the underline and anything different should cause a schema error.

glean/src/core/metrics/types/string_list.ts
ChinYing-Li: Since `add` calls the async function `Context.metricsDatabase.transform`, should `add` and `set` be made async? I saw that other metric types does not have async `set`.

ChinYing-Li:
Without `async` means I have to write something like
```javascript
truncatedValule = "";
 const truncatedValuePromise = truncateStringAtBoundaryWithError(this, value, MAX_STRING_LENGTH);
truncatedValuePromise.then((v) => {
  truncatedValue = v;
});
```
However the code above led to failure of unit test case `attempt to add string to string list of maximum length records an error`: the empty string "" was inserted instead.

glean/src/core/metrics/types/string_list.ts
ChinYing-Li: IMO, it's better to keep the `static asyn` internal methods, so that we can maintain the coherence of the interface (let `add` and `set` be normal functions) and avoid lint error `Promises must be handled appropriately or explicitly marked as ignored `.

brizental: > Since add calls the async function Context.metricsDatabase.transform, should add and set be made async? I saw that other metric types does not have async set.

The body of public functions in MetricTypes need to be wrapped in a `dispatcher.launch` call, which returns `void`. They must never return a `Promise`, this is very important as we don't want users to have to `await` on any of Glean's public APIs.

> IMO, it's better to keep the static asyn internal methods, so that we can maintain the coherence of the interface (let add and set be normal functions)

Sure, this makes sense. You are right :)

> and avoid lint error Promises must be handled appropriately or explicitly marked as ignored .

As I mentioned above this functions must never return promises, so this should not happen.

glean/src/core/metrics/types/string_list.ts
```diff
@@ -0,0 +1,162 @@
...
+    const truncatedList: string[] = [];
+    const newLength = Math.min(value.length, MAX_LIST_LENGTH);
+    for (let i = 0; i < newLength; ++i) {
+      const truncatedString = truncateStringAtBoundaryWithError(this, value[i], MAX_STRING_LENGTH);
+      truncatedString.then((v) => {
+        truncatedList.push(v);
+      });
+    }
+    const metric = new StringListMetric(truncatedList);
+    Context.metricsDatabase.record(this, metric);
```
brizental, line 90: The whole body needs to be wrapped in a `dispatcher.launch` call, like so:

```suggestion
Context.dispatcher.launch(async () => {
    if (!this.shouldRecord(Context.uploadEnabled)) {
      return;
    }

    if (value.length > MAX_LIST_LENGTH) {
      Context.errorManager.record(
        this,
        ErrorType.InvalidValue,
        `String list length of ${value.length} exceeds maximum of ${MAX_LIST_LENGTH}.`
      );
    }

    const truncatedList: string[] = [];
    const newLength = Math.min(value.length, MAX_LIST_LENGTH);
    for (let i = 0; i < newLength; ++i) {
      const truncatedString = truncateStringAtBoundaryWithError(this, value[i], MAX_STRING_LENGTH);
      truncatedString.then((v) => {
        truncatedList.push(v);
      });
    }
    const metric = new StringListMetric(truncatedList);
    Context.metricsDatabase.record(this, metric);
 });
```

glean/src/core/metrics/types/string_list.ts
```diff
@@ -0,0 +1,162 @@
...
+   * # Note
+   *
+   * - If the list is already of length `MAX_LIST_LENGTH`, log an error.
+   * - Truncates the value if it is longer than `MAX_STRING_LENGTH` bytes
+   * and logs an error.
+   *
+   * @param value The string to add.
+   */
+  async add(value: string): Promise<void> {
+    if (!this.shouldRecord(Context.uploadEnabled)) {
```
brizental, line 105: Same comment regarding the `dispatcher.launch` as I left in the `set` function.

brizental:
> Without `async` means I have to write something like
> 
> ```js
> truncatedValule = "";
>  const truncatedValuePromise = truncateStringAtBoundaryWithError(this, value, MAX_STRING_LENGTH);
> truncatedValuePromise.then((v) => {
>   truncatedValue = v;
> });
> ```

By wrapping the whole function body in a `dispatcher.launch` call, it will be inside an `async` function block so you can use `await` normally. You can see that pattern in action on the URL metric type `set` function (https://github.com/mozilla/glean.js/blob/main/glean/src/core/metrics/types/url.ts#L94-L113) or any other metric type.

The dispatcher orchestrates Glean's async tasks, so that they are run in the order the user called them. But that all happens in the background, without the user having to `await` on anything. That is a bit confusing, let me know if it's still unclear I am happy to try and clarify more :)

> However the code above led to failure of unit test cas

They failure is probably happening because the code is not using the dispatcher, so at the time the assertion is called the action that should precede has not completed yet.

brizental (changes requested):

glean/src/core/metrics/types/string_list.ts
ChinYing-Li: Hi @brizental , I just want to make sure that you still want me to merge `_private_addUndispatched` and `add`, is that correct? Or does it make more sense to keep `_private_addUndispatched`?

brizental: On a second look, I keep my initial opinion that it is not necessary to have the undispatched methods.

ChinYing-Li:
@brizental I've addressed all the issues, and would be great if you can take a look. Thank you so much for the guidance!

glean/src/core/metrics/types/string_list.ts
```diff
@@ -0,0 +1,164 @@
...
+   * @param value The string to add.
+   */
+  add(value: string): void {
+    Context.dispatcher.launch(async () => {
+      if (!this.shouldRecord(Context.uploadEnabled)) {
+        return;
+      }
+
+      const truncatedValue = await truncateStringAtBoundaryWithError(this, value, MAX_STRING_LENGTH);
+      let currentLen = -1;
```
brizental, line 108: Why did you choose to start this as `-1`? Why not `0`?

glean/src/core/metrics/types/string_list.ts
```diff
@@ -0,0 +1,164 @@
...
+            result = [value];
+          }
+          metric.set(result);
+          return metric;
+        };
+      })(truncatedValue);
+
+      await Context.metricsDatabase.transform(this, transformFn);
+
+      if (currentLen == MAX_LIST_LENGTH) {
```
brizental, line 132:
```suggestion
      if (currentLen >= MAX_LIST_LENGTH) {
```

Just to be extra sure, hehe.

glean/src/core/metrics/types/string_list.ts
```diff
@@ -0,0 +1,164 @@
...
+        await Context.errorManager.record(
+          this,
+          ErrorType.InvalidValue,
+          `String list length of ${currentLen}+1 exceeds maximum of ${MAX_LIST_LENGTH}.`
```
brizental, line 136:
```suggestion
          `String list length of ${currentLen} exceeds maximum of ${MAX_LIST_LENGTH}.`
```
?

glean/src/core/metrics/types/string_list.ts
```diff
@@ -0,0 +1,164 @@
...
+  async testGetValue(ping: string = this.sendInPings[0]): Promise<string[] | undefined> {
+    let metric: string[] | undefined;
+    await Context.dispatcher.testLaunch(async () => {
+      metric = await Context.metricsDatabase.getMetric<string[]>(ping, this);
+    });
+    return metric;
+  }
+}
+
+export default StringListMetricType;
```
brizental, line 164: Missing blank line at the onf file.

samples/web-extension/typescript/package-lock.json
```diff
@@ -117,9 +117,15 @@
       "dev": true
     },
     "node_modules/@types/node": {
+<<<<<<< HEAD
```
brizental, line 120: Let's take these changes out of this PR.

glean/tests/unit/core/metrics/string_list.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+    });
+
+    const testStringList: string[] = [];
+    metric.set(testStringList);
+    assert.deepStrictEqual(await metric.testGetValue("aPing"), testStringList);
+    assert.strictEqual(
+      await metric.testGetNumRecordedErrors(ErrorType.InvalidValue), 0
+    );
+  });
+});
```
brizental, line 183: Missing blank line at the end of file.

glean/tests/unit/core/metrics/string_list.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+      sendInPings: ["aPing"],
+      lifetime: Lifetime.Ping,
+      disabled: false,
+    });
+
+    const testStringList = [];
+    const testString = "test";
+    for(let i = 0; i < MAX_LIST_LENGTH - 1; ++i) {
+      testStringList.push(testString);
+    }
```
brizental, line 147: This block can be replaced by:

```js
const testStringList = Array(MAX_LIST_LENGTH - 2).fill("test")
```

That is also a bit less confusing, IMO. It took me a while to understand why we have `i < MAX_LIST_LENGTH - 1`.

brizental (changes requested):
This is looking great, I left a few nits, but nothing major. Once these are adressed this should be good to go.

Thanks for the work here @ChinYing-Li !

glean/src/core/metrics/types/string_list.ts
ChinYing-Li: `currentLen` is the size of the string list before pushing another string to it, so the error message should use `currentLen + 1`.

brizental: Ok, I see. In this case, the `+` needs to be inside the `{}`.

brizental:
Hey @ChinYing-Li I see you addressed the review comments, but there are three still missing :) Would you mind addressing those and rebasing this PR so I can merge? Thank you so much :)

Also, the `package-lock.json` can be dropped. That has already ben fixed in a different PR.

ChinYing-Li:
Hi @brizental , sorry for the delay! Please review and let me know if anything else needs to be fixed. Thank you.

ChinYing-Li:
Hi @brizental , Sure, I can update the Glean book!

PR #636, opened 2021-08-18
Bug 1724382 - Fix off-by-one intermittent error in EventsDatabase tests

brizental:
This seems to have fix it. I say it "seems" to have fixed it because this is how I tested:

- Ran the tests 10x with the original code. Got test failures 4 out of those 10 times.
- Applied the fixes.
- Ran the tests 10x with the new code. Got no failures.


Dexterp37 (approved):
💯

PR #647, opened 2021-08-19
Bug 1726327 - Implement the new text metric type

badboy:
The new text metric type is really just a long string,
but only allowed under specific circumstances.

---

Draft while I fix outstanding things:

* [x] glean_parser release with that metric: [v4.0.0](https://pypi.org/project/glean-parser/4.0.0/)
* [x] glean_parser update in glean.js
* [x] schema update: [#692](https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/692)

### DIFF
--- a/glean/src/core/metrics/types/text.ts
+++ b/glean/src/core/metrics/types/text.ts
@@ -1,0 +1,89 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import type { CommonMetricData } from "../index.js";
+import { isString, truncateStringAtBoundaryWithError } from "../../utils.js";
+import { MetricType } from "../index.js";
+import { Context } from "../../context.js";
+import { Metric } from "../metric.js";
+
+// The maximum number of characters for text.
+export const TEXT_MAX_LENGTH = 200 * 1024;
+
+export class TextMetric extends Metric<string, string> {
+  constructor(v: unknown) {
+    super(v);
+  }
+
+  /**
+   * Validates that a given value is within bounds.
+   *
+   * @param v The value to validate.
+   * @returns Whether or not v is valid text data.
+   */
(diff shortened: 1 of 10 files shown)


glean/src/core/metrics/types/text.ts
```diff
@@ -0,0 +1,89 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import type { CommonMetricData } from "../index.js";
+import { isString, truncateStringAtBoundaryWithError } from "../../utils.js";
+import { MetricType } from "../index.js";
+import { Context } from "../../context.js";
+import { Metric } from "../metric.js";
+
+// The maximum length for text: 200 kilobytes.
```
brizental, line 11:
```suggestion
// The maximum length for text: 200 characters.
```

If you want to check by kylobites then the string needs to be encoded before the check. We use a function for that `strToU8`, but we have been checking character length in URL and String metric types until now.

badboy: hm, fwiw the jsonschema defines the string length as the number of its characters according to the JSON rfc, which defines characters as unicode characters.
So ... uhm ... yeah, lets stick with characters.

badboy:
Ok, took me a moment to understand why the size check failed.
It turns out the check size once runs against the code in the current working directory.
And the second time it checks out the `main` branch to link in Glean, but then switches back and runs the benchmark code.

So it uses the updated benchmark code, which of course includes the changes to e.g. `benchmarks/size/webext/max.js` and thus fails.
It should really checkout main and run the benchmark there fully.

brizental (approved):

brizental:
Oh, also, are you also going to do the documentation stuff under Bug 1726327? If not, please file a follow-up for that :)

badboy:
> Oh, also, are you also going to do the documentation stuff under Bug 1726327? If not, please file a follow-up for that :)

[Followup bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1727299)

PR #658, opened 2021-08-23
Bug 1712920 - Implement rate limiting in ping upload

brizental:
This actually fixes a bug: there was no code path to re-start the PingUploader dispatcher once it was stopped due to max recoverable error being hit. None of our consumers have updated yet,s o if I make a release after this is merged we should be fine.

  - https://github.com/mozilla/glean/pull/1758

### DIFF
--- a/glean/src/core/upload/index.ts
+++ b/glean/src/core/upload/index.ts
@@ -148,7 +148,7 @@ class PingUploader implements PingsDatabaseObserver {
         log(
           LOG_TAG,
           [
-            "Attempted to upload a ping, but Glean is has reached maximum recoverable upload failures",
+            "Attempted to upload a ping, but Glean has reached maximum recoverable upload failures",
             "for the current uploading window.",
             `Will retry in ${(remainingTime || 0) / 1000}s.`
           ],
--- a/glean/src/core/upload/index.ts
+++ b/glean/src/core/upload/index.ts
@@ -26,7 +26,7 @@ import RateLimiter, { RateLimiterState } from "./rate_limiter.js";
 const LOG_TAG = "core.Upload";
 
 // Default rate limiter interval, in milliseconds.
-const RATE_LIMITER_INTERVAL = 60 * 1000;
+const RATE_LIMITER_INTERVAL_MS = 60 * 1000;
 // Default max pings per internal.
 const MAX_PINGS_PER_INTERVAL = 15;
 
--- a/glean/src/core/upload/index.ts
+++ b/glean/src/core/upload/index.ts
@@ -98,7 +98,7 @@ class PingUploader implements PingsDatabaseObserver {
     platform: Platform,
     private readonly pingsDatabase = Context.pingsDatabase,
     private readonly policy = new Policy(),
-    private readonly rateLimiter = new RateLimiter(RATE_LIMITER_INTERVAL, MAX_PINGS_PER_INTERVAL)
+    private readonly rateLimiter = new RateLimiter(RATE_LIMITER_INTERVAL_MS, MAX_PINGS_PER_INTERVAL)
   ) {
     this.processing = [];
     // Initialize the ping uploader with either the platform defaults or a custom
--- a/glean/src/core/upload/index.ts
+++ b/glean/src/core/upload/index.ts
@@ -21,8 +21,14 @@ import type PlatformInfo from "../platform_info.js";
 import { UploadResult } from "./uploader.js";
 import type Uploader from "./uploader.js";
 import { UploadResultStatus } from "./uploader.js";
+import RateLimiter, { RateLimiterState } from "./rate_limiter.js";
 
 const LOG_TAG = "core.Upload";
+
+// Default rate limiter interval, in milliseconds.
+const RATE_LIMITER_INTERVAL = 60 * 1000;
+// Default max pings per internal.
+const MAX_PINGS_PER_INTERVAL = 15;
 
 /**
  * Create and initialize a dispatcher for the PingUplaoder.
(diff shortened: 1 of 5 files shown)


glean/src/core/upload/index.ts
```diff
@@ -121,6 +128,35 @@ class PingUploader implements PingsDatabaseObserver {
...
+      this.dispatcher.resume();
+    } else {
+      this.dispatcher.stop();
+
+      if (rateLimiterState === RateLimiterState.Throttled) {
+        log(
+          LOG_TAG,
+          [
+            "Attempted to upload a ping, but Glean is currently throttled.",
+            `Pending pings will be processed in ${(remainingTime || 0) / 1000}s.`
```
brizental, line 142: I kinda of don't like this message, as it is promising something that is not true: the pings _may_ be processed after the remaining time, but it doesn't mean they will, there is no trigger that goes off once that time is over. 

On the other SDKs we do have one though, so I will file a bug to do the same here. See how that happens in the [Python](https://github.com/mozilla/glean/blob/main/glean-core/python/glean/net/ping_upload_worker.py#L156) or [Android](https://github.com/mozilla/glean/blob/main/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt#L120) SDKs.

Question is: is everyone confortable with the empty promise here until that is merged? I personally don't think it is a big deal, but best to flag it.

brizental: https://bugzilla.mozilla.org/show_bug.cgi?id=1727076 filed.

Dexterp37: > Question is: is everyone confortable with the empty promise here until that is done? I personally don't think it is a big deal, but best to flag it.

I don't think it's a big deal.

glean/src/core/upload/rate_limiter.ts
```diff
@@ -0,0 +1,128 @@
...
+ * An enum to represent the current state of the RateLimiter.
+ */
+export const enum RateLimiterState {
+  // The RateLimiter has not reached the maximum count and is still incrementing.
+  Incrementing,
+  // The RateLimiter has not reached the maximum count, but it is also not incrementing.
+  Stopped,
+  // The RateLimiter has reached the maximum count for the current interval.
+  //
+  // This variant contains the remaining time (in milliseconds)
```
Dexterp37, line 17: This is a bit confusing: the enum itself is not holding the remaining time in MS, it's just that we return the time from `getState`: could you clarify this a bit?

glean/src/core/upload/rate_limiter.ts
```diff
@@ -0,0 +1,128 @@
...
+import { isUndefined, getMonotonicNow } from "../utils.js";
+
+/**
+ * An enum to represent the current state of the RateLimiter.
+ */
+export const enum RateLimiterState {
+  // The RateLimiter has not reached the maximum count and is still incrementing.
+  Incrementing,
+  // The RateLimiter has not reached the maximum count, but it is also not incrementing.
+  Stopped,
```
Dexterp37, line 14: We seem to be returning the remaining time in this case, too: should we amend this comment or change `getState`?

glean/src/core/upload/rate_limiter.ts
```diff
@@ -0,0 +1,128 @@
...
+    } {
+    if (this.shouldReset()) {
+      this.reset();
+    }
+
+    const remainingTime = this.interval - this.elapsed;
+    if (this.stopped) {
+      return {
+        state: RateLimiterState.Stopped,
+        remainingTime,
```
Dexterp37, line 101: We're returning the remaining time here, do we need to fix the comment on the enum definition or remove this?

glean/src/core/upload/rate_limiter.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  get elapsed(): number {
+    if (isUndefined(this.started)) {
+      return NaN;
+    }
+
+    const now = getMonotonicNow();
+    const elapsed = now - this.started;
+
+    // It's very unlekily elapsed will be a negative number since we are using a monotonic timer
```
Dexterp37, line 47:
```suggestion
    // It's very unlikely elapsed will be a negative number since we are using a monotonic timer
```

glean/src/core/upload/rate_limiter.ts
```diff
@@ -0,0 +1,128 @@
...
+  private reset(): void {
+    this.started = getMonotonicNow();
+    this.count = 0;
+    this.stopped = false;
+  }
+
+  /**
+   * The rate limiter should reset if
+   *
+   * 1. It has never started;
```
Dexterp37, line 65: Why is this the case? If it has never started, it should already be in the stopped state, right?

Dexterp37 (changes requested):

glean/src/core/upload/rate_limiter.ts
brizental: Ack, this is because I copy pasted this comment from Rust, where enums may hold data. Which is not the case for TypeScript. I'll reword it.

glean/src/core/upload/rate_limiter.ts
brizental: Nah, I don't think we need to mention the remaining time at all in these enum comments.

glean/src/core/upload/rate_limiter.ts
brizental: Answered by the previous responses I think.

glean/src/core/upload/rate_limiter.ts
brizental: If it has never started it does not have a `startTime` so we need to generate it. I can reword it to say exactly this though. Full disclaimer, this was also copied from the Rust code, hehe. 

Maybe I should re-read these comments after sending the PR?!

Dexterp37 (approved):

PR #662, opened 2021-08-24
Bug 1727318 - Dispatcher can't be stopped while it's being shutdown

brizental:
I also found another bug on the rate limit implementaion while working on this bug, which is fixed in 5be95ad.

Detailed descriptions on the commit messages explain the changes made.

   - I think CHANGELOG changes are unnecessary here. This is part of rate limiting implementation, which is already there.

### DIFF
--- a/glean/src/core/dispatcher.ts
+++ b/glean/src/core/dispatcher.ts
@@ -30,7 +30,7 @@ const enum Commands {
   // The dispatcher will enqueue a new task.
   //
   // This command is always followed by a concrete task for the dispatcher to execute.
-  Task,
+  Task = "Task",
   // Same as the `Task` command,
   // but the task enqueued by this command is not cleared by the `Clear` command.
   //
@@ -39,15 +39,15 @@ const enum Commands {
   // `Shutdown` will still clear these tasks.
   //
   // Unless unavoidable, prefer using the normal `Task`.
-  PersistentTask,
+  PersistentTask = "PersistentTask",
   // The dispatcher should stop executing the queued tasks.
-  Stop,
+  Stop = "Stop",
   // The dispatcher should stop executing the queued tasks and clear the queue.
-  Clear,
+  Clear = "Clear",
   // The dispatcher will clear the queue and go into the Shutdown state.
-  Shutdown,
+  Shutdown = "Shutdown",
   // Exactly like a normal Task, but spawned for tests.
-  TestTask,
+  TestTask = "TestTask",
 }
 
 // A task the dispatcher knows how to execute.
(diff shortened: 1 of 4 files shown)


glean/src/core/dispatcher.ts
```diff
@@ -320,9 +324,18 @@ class Dispatcher {
...
    * In order to re-start the dispatcher, call the `resume` method.
+   *
+   * # Note
+   *
+   * In case a Shutdown command has been launched before this command,
+   * this command will result in the queue being cleared.
    */
   stop(): void {
-    this.launchInternal({ command: Commands.Stop }, true);
+    if (this.shuttingDown) {
```
Dexterp37, line 339: Wouldn't this disrupt the sequence of actions when shutting down? Shouldn't we still be executing tasks up until the "shutdown"?

brizental: So, the stop command disrupts the sequence of actions, with or without shutdown, by default. It is up to the caller to use `stop(false)` when they don't want stop to be enqueued in front of everything else.

brizental: I do agree this could be surprising behaviour, I was torn while implementing.

Dexterp37: I see, that's ok, sad but ok :)

glean/tests/unit/core/dispatcher.spec.ts
```diff
@@ -464,6 +464,44 @@ describe("Dispatcher", function() {
...
+      await new Promise<void>(resolve => setTimeout(() => resolve(), 100));
+      dispatcher.stop();
+    });
+
+    let executionCounter2 = 0;
+    const counterTask2 = (): Promise<void> => {
+      executionCounter2++;
+      return Promise.resolve();
+    };
+    // Launch another group of tasks
```
Dexterp37, line 493:
```suggestion
    // Launch another group of tasks.
```

glean/tests/unit/core/dispatcher.spec.ts
```diff
@@ -464,6 +464,44 @@ describe("Dispatcher", function() {
...
+    dispatcher = new Dispatcher();
+    dispatcher.flushInit();
+
+    let executionCounter1 = 0;
+    const counterTask1 = (): Promise<void> => {
+      executionCounter1++;
+      return Promise.resolve();
+    };
+
+    // Launch a group of tasks
```
Dexterp37, line 477:
```suggestion
    // Launch a group of tasks.
```

Dexterp37 (approved):
r+ with a changelog entry added.

PR #687, opened 2021-08-31
Bug 1727269 - Run the size benchmark against a full main checkout

badboy:
Instead of switching back and forth between branches this uses an
additional worktree.
This guarantees that the benchmark runs fully against code in the main
branch first, then fully against code in the current checkout.

This requires that the report code in the latest checkout needs to be
able to consume the report from the `main` branch,
so be careful when changing its format.

---

I decided to do this in an external script.
That seems to be easier than to cram this into individual npm scripts.

~~Also note: this will only start working properly AFTER this PR has been merged and can't be tested on this PR.~~

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -1,0 +1,19 @@
+# Glean.js size benchmarks
+
+Compare the size of the different Glean bundles across the local `main` branch and the current checkout.
+
+## How to run
+
+To run a local report:
+
+```
+npm run size:report:dry
+```
+
+To run the report and post the results to a GitHub pull request:
+
+```
+npm run size:report
+```
+
+_Note: This requires a `GITHUB_TOKEN` and the CircleCI environment to know which PR to post to._
--- a/benchmarks/run.sh
+++ b/benchmarks/run.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+set -xe
+
+WORKSPACE_ROOT="$( cd "$(dirname "$0")/.." ; pwd -P )"
+cd "$WORKSPACE_ROOT"
+
+# Benchmark the main branch
+tmpdir=$(mktemp -d)
+git worktree add --force "${tmpdir}" main
+pushd "${tmpdir}"
+npm --prefix ./benchmarks install
+npm --prefix ./benchmarks run size:build:main
+cp -a "${tmpdir}/benchmarks/dist" "${WORKSPACE_ROOT}/benchmarks"
+popd
+git worktree remove --force "${tmpdir}"
+
+# Benchmark the current code
+npm --prefix ./benchmarks install
+npm --prefix ./benchmarks run size:build
+
+# Post the details
+node --experimental-json-modules benchmarks/size/report.js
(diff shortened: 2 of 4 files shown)


benchmarks/run.sh
```diff
@@ -0,0 +1,23 @@
...
+
+# Benchmark the main branch
+tmpdir=$(mktemp -d)
+git worktree add "${tmpdir}" origin/main
+pushd "${tmpdir}"
+npm --prefix ./benchmarks install
+npm --prefix ./benchmarks run size:build:main
+cp -a "${tmpdir}/benchmarks/dist" "${WORKSPACE_ROOT}/benchmarks"
+popd
+git worktree remove --force "${tmpdir}"
```
brizental, line 16: TIL `git worktree` 🎉

benchmarks/README.md
```diff
@@ -0,0 +1,19 @@
...
+To run a local report:
+
+```
+DRY_RUN=1 ./run.sh
+```
+
+To run the report and post the results to a GitHub pull request:
+
+```
+./run.sh
```
brizental, line 16:
```suggestion
npm run size:report
```

benchmarks/README.md
```diff
@@ -0,0 +1,19 @@
+# Glean.js size benchmarks
+
+Compare the size of the different Glean bundles across the `main` branch and the current checkout.
+
+## How to run
+
+To run a local report:
+
+```
+DRY_RUN=1 ./run.sh
```
brizental, line 10: nit: Can we change these examples to the npm scripts instead?

```suggestion
npm run size:report:dry
```

benchmarks/run.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+set -xe
+
+WORKSPACE_ROOT="$( cd "$(dirname "$0")/.." ; pwd -P )"
+cd "$WORKSPACE_ROOT"
+
+# Benchmark the main branch
+tmpdir=$(mktemp -d)
+git worktree add "${tmpdir}" origin/main
```
brizental, line 10: Can we change this to `git worktree add "${tmpdir}" main`? In CircleCI it should work fine, but in my local machine `origin/main` will usually point to an old version of the repo (I don't really update my fork ever). 

We can then mention somewhere in the README.md that the size check is against the local main branch.

```suggestion
git worktree add "${tmpdir}" main
```

brizental (approved):
Thanks for adding a README there too 👍

benchmarks/run.sh
badboy: It's great! Have been using that for some things in the past.

moz-glean:
# Build size report

Merging https://github.com/mozilla/glean.js/pull/687 into [main](https://github.com/mozilla/glean.js) will:
* **Leave** the size of full **Qt/QML** bundle **unchanged**.

---

|--:|:---:|:---:|:---:|
| **Web Extension** |
| core only | 52 KB | 52 KB | 📉 0 bytes |
| full bundle | 76 KB | 76 KB | 📉 0 bytes |
| **Qt/QML** |
| core only | 134 KB | 134 KB | 📉 0 bytes |
| full bundle | 134 KB | 134 KB | 📉 0 bytes |

PR #689, opened 2021-08-31
Add chutten and mdboom to the Glean.js reviewers rotation

brizental:
As discussed in today's meeting.


travis79 (approved):
My first auto-assignment is to approve more auto-asignees!  😎

PR #692, opened 2021-09-01
Bug 1728128 - Ensure events database is initialized at a time when metrics can be recorded

brizental:
After tests, I confirmed none of the sample apps were recording the `glean.restarted` event when Glean was initialized :/ That was the case, because we were calling `eventsDatabase.initialize` before setting `Context.uploadEnabled`. That meant `Context.uploadEnabled` was `undefined` upon events database initialization and `undefined` being a falsy value, `.shouldRecord()` was returning `false`, which resulted in the `glean.restarted` and `execution_counter` metrics not being recorded.

In 8324569 I fix that and in 4cb6118 I add a test to ensure that if this behaviour ever changes we will see test failures.

The commits in between these two commits are there to make sure the test _does_ fail if the behaviour is changed. I actually found a bug in the way we were reseting Glean in tests. We were not reseting all the preoperties in the `Context`, only a few. That would prevent the test I added from actually failing when it had to. 

   - This is a bugfix, I guess a CHANGELOG entry is documentation enough.

### DIFF
--- a/glean/src/core/glean.ts
+++ b/glean/src/core/glean.ts
@@ -239,21 +239,18 @@ class Glean {
       Context.debugOptions = correctConfig.debug;
       Glean.instance._config = correctConfig;
 
-
-      // IMPORTANT!
-      // Any pings we want to send upon initialization should happen before these two lines.
-      //
-      // Clear application lifetime metrics.
-      await Context.metricsDatabase.clear(Lifetime.Application);
-      // Initialize the events database.
-      await Context.eventsDatabase.initialize();
-
       // We need to mark Glean as initialized before dealing with the upload status,
       // otherwise we will not be able to submit deletion-request pings if necessary.
       //
       // This is fine, we are inside a dispatched task that is guaranteed to run before any
       // other task. No external API call will be executed before we leave this task.
       Context.initialized = true;
+
+      // IMPORTANT!
+      // Any pings we want to send upon initialization should happen before this line.
+      //
+      // Clear application lifetime metrics.
+      await Context.metricsDatabase.clear(Lifetime.Application);
 
       // The upload enabled flag may have changed since the last run, for
       // example by the changing of a config file.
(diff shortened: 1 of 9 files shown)


glean/src/core/glean.ts
```diff
@@ -235,26 +240,19 @@ class Glean {
...
-      await Context.metricsDatabase.clear(Lifetime.Application);
-      // Initialize the events database.
-      await Context.eventsDatabase.initialize();
-
       // We need to mark Glean as initialized before dealing with the upload status,
       // otherwise we will not be able to submit deletion-request pings if necessary.
       //
       // This is fine, we are inside a dispatched task that is guaranteed to run before any
       // other task. No external API call will be executed before we leave this task.
       Context.initialized = true;
```
badboy, line 248: Is anything else depending on this value and could now cause problems because there's still intialization happening after this value is set to true?

glean/tests/unit/core/glean.spec.ts
```diff
@@ -548,4 +548,28 @@ describe("Glean", function() {
...
+
+    // Record this event, so that when we re-initialize the events database
+    // it will record a glean.restarted event on the `custom` ping events list.
+    event.record();
+
+    await Glean.testResetGlean(testAppId, true, undefined, false);
+
+    // Check that Glean was able to record the `glean.restarted` event on initialization.
+    const restartedEvent = getGleanRestartedEventMetric(["custom"]);
+    console.log(await restartedEvent.testGetValue("custom"));
```
badboy, line 569: Stray console.log, I think this can go? The below assert should explode when testGetValue is returning undefined

badboy (approved):
Looks good.

glean/tests/unit/core/glean.spec.ts
brizental: Good catch... I should maybe add a lint to remind me to remove my stray`console.log`s.

brizental:
```suggestion
```

PR #733, opened 2021-09-16
Bug 1728809 - Implement a simple Node.js sample app

brizental:
~This is a draft until https://github.com/mozilla/glean_parser/pull/390 makes it into a release and `glean_parser` is updated here.~ https://github.com/mozilla/glean_parser/releases/tag/v4.1.0

### DIFF
--- a/samples/node/main.js
+++ b/samples/node/main.js
@@ -9,9 +9,12 @@ import { custom } from "./generated/pings.js";
 import { appStarted } from "./generated/sample.js";
 
 export default () => {
+  const { GLEAN_DEBUG_VIEW_TAG } = process.env;
+  if (GLEAN_DEBUG_VIEW_TAG) {
+    Glean.setDebugViewTag(GLEAN_DEBUG_VIEW_TAG);
+  }
+
   Glean.setLogPings(true);
-  // Uncomment when you need to inspect the pings sent in the Glean Debug Ping Viewer.
-  // Glean.setDebugViewTag("gleanjs-node-sample");
   Glean.initialize("gleanjs-node-sample", true);
 
   appStarted.set();
--- a/samples/node/main.js
+++ b/samples/node/main.js
@@ -1,0 +1,19 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+"use strict";
+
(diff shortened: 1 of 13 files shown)


samples/node/main.js
```diff
@@ -0,0 +1,19 @@
...
+"use strict";
+
+import Glean from "@mozilla/glean/node";
+import { custom } from "./generated/pings.js";
+import { appStarted } from "./generated/sample.js";
+
+export default () => {
+  Glean.setLogPings(true);
+  // Uncomment when you need to inspect the pings sent in the Glean Debug Ping Viewer.
+  // Glean.setDebugViewTag("gleanjs-node-sample");
```
badboy, line 14: Make it depend on an app argument or an environment variable!

badboy (approved):
This is so boring compared to the 300 lines of code it takes for our iOS/Android sample app!

samples/node/main.test.js
```diff
@@ -0,0 +1,40 @@
...
+  const testAppId = `node.test.${this.title}`;
+
+  beforeEach(async function () {
+    await Glean.testResetGlean(testAppId);
+  });
+
+  describe("sample test", function () {
+    it("appStarted metric is recorded and ping is sent", async function () {
+      // Setup validator for sending of the custom ping.
+      const pingWasSent = custom.testBeforeNextSubmit(async () => {
```
badboy, line 22: Yey, big plus for having tests on the sample app!

samples/node/main.test.js
```diff
@@ -0,0 +1,40 @@
...
+        // Check no errors were recorded on that metric either.
+        assert.equal(
+          await appStarted.testGetNumRecordedErrors(ErrorType.InvalidValue),
+          0
+        );
+      });
+
+      // Call the sample script.
+      // This will record the metric and send the ping we are expecting.
+      void main();
```
badboy, line 34: huh, this looks like a function declaration in C.
Is the `void ...` pattern just for ignoring any return value?

brizental: Exactly! But it is actually not necessary and a leftover from development here. For a time the `main` function used to return a Promise, so I put the `void` there so the linter would be fine with me not `await`ing on it. Let me remove it... 
```suggestion
      main();
```

samples/node/main.js
brizental: Huh, I guess since we are in Node.js land we _can_ use env vars right. That is a nice suggestion, I'll change it.

PR #744, opened 2021-09-20
Bug 1727776 - Attempt to fix crash at submitPing on Qt in iOS

brizental:
**Opening this as a draft because, although I managed to find multiple potentital issues and fix them, we are still getting a crash on submitPing. The crash is 100% reproducible both in my iPhone and iPad.**

The error I get is:

```
Thread 1: EXC_BAD_ACCESS (code=2, address=0x16d117ff8)
```

And it happens specifically at the time we call `eventsStore.get` [here](https://github.com/mozilla/glean.js/blob/main/glean/src/core/metrics/events_database/index.ts#L231). Some other information on the crash:

- It only happens on real devices, not in simulators.
- We never even enter the `get` function. The crash happens when we call it.
- The error only happens if there is data to send, if we get "ping is empty. bailing out", there is no crash. However it's not necessary that the data we have is in the events store, even though the crash only happens there. See [this ping](https://debug-ping-preview.firebaseapp.com/pings/ios-gleanjs-sample) which I was able to send from my iPad after commenting the line linked to above. This indicates there is no issue getting data from the other stores, just the events store. The stores are all the same so this does not tell me much.
- During development I was able to send pings once or twice. Even though I was able to send pings I still got a crash, but since Xcode stops on crashes I was sometimes able to send the ping while stepping through the code. 

[This stack overflow question](https://stackoverflow.com/questions/61793803/stack-overflow-thread-1-exc-bad-access-code-2-address-0x16d09aa00) describes a bug very similar to the one we are having, but that user is coming from a very different setup and not even using Qt. Either way I am also not sure how to attempt a similar fix to his on our codebase.

[QTBUG-50319 - QML garbage collection deletes objects still in use](https://bugreports.qt.io/browse/QTBUG-94566) may also be related to this issue, but it seems to have beend fixed in newer versions. Since I did test the sample app in the newest version of Qt (6.1.3) and still got the crash, maybe it is not the root cause. There is some debate as to whether it was really fixed in the comments of that bug, so I have not discarded it completely.

Finally there is also [this Stack Overflow](https://stackoverflow.com/questions/15257946/declare-global-property-in-qml-for-other-qml-files#comment51835543_15265661) question. While not obviously related contains this quote:

> I must notice that .pragma library allowes JS file to be used by several QML files, but does not guarantee that there will be only one instance of the library. I've just encountered severe bug in my application that is due to my attempt of using JS library as global state container.

Although this quote is from 6 years ago, it makes me wonder if the weird behaviour we are getting is due to the fact that we expect Glean to be a singleton, but maybe it is not really. Not sure how possible that is though, since I am able to reproduce the issue in the sample app which has only a single view. I would expect to encounter singleton issues when navigating thorugh different views.


brizental:
Well, good news. Turns out I did not realize earlier but the crash I described is only happening in Debug mode. It is a crash created by Qt debug intrumentation code, specifically [QV4::Moth::VME::interpret](https://code.woboq.org/qt5/qtdeclarative/src/qml/jsruntime/qv4vme_moth.cpp.html#_ZN3QV44Moth3VME9interpretEPNS_13CppStackFrameEPNS_15ExecutionEngineEPKc) and I believe it happens due to a `null reference`. I will try describe this nicely and create a bug report on the Qt bug tracker, maybe there is a workaround that we can apply in our code for it which I can't see at the moment. 

Otherwise, since this is a crash in Qt's JS interpreter itself, I would say the way forward is documenting that Qt + iOs + Debug mode is not possible at the moment and moving forward from this issue.

See pings sent from:

- [The VPN appp](https://debug-ping-preview.firebaseapp.com/pings/vpn-from-ipad)
- [The Glean.js C++ sample app](https://debug-ping-preview.firebaseapp.com/pings/ios-gleanjs-sample)

brizental:
Related PR on the VPN repo https://github.com/mozilla-mobile/mozilla-vpn-client/pull/1854

brizental:
Closing in favor of https://github.com/mozilla/glean.js/pull/779

PR #768, opened 2021-09-28
Bug 1702468 - Simplify build config and remove code duplication from final lib 

brizental:
**Draft as this requires glean_parser changes.**

From the commit message:

> Previous to this commit we were building one whole new library for each
> platform. That was unnecessary, the only part that is indeed different
> per library is the platform/ module implementation.
>
> This commit simplifies that, by building the library only once and
> relying ecluisvely on the entry points to make sure each user has access
> to the desired platform.
> 
> Note 1: The Qt platform will still need to be built separate, that does
> not change.
> 
> Note 2: This is a breaking change because it breaks the glean_parser
> templates, because it changes the metrics and pings import paths. I
> could have made this a backwards compatible change, but I think at this
> point we are still unstanble enough that I can not do that.

---

This was one of _those_ moments when you look at your code from the past and think "What was I thinking?!". 

Anyways, glad to have simplified it :)


brizental:
> This is a breaking change because it breaks the glean_parser templates, because it changes the metrics and pings import paths. 

I guess actually it is not really a breaking change since glean_parser is bundled with the npm lib under the `glean` CLI, so this is actually all internal changes.

brizental:
(Please ignore lint fialure, dependabot has been messed up our package-lock file again).

Dexterp37 (changes requested):

Dexterp37 (approved):

PR #798, opened 2021-10-06
Bug 1733634 - Add audit step to CI and fix dependency issues

brizental:
The issues were fixed by running `npm run audit --force`. 

Nice suggestion, thanks @birdsarah !

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,7 +43,7 @@ jobs:
           command: npm --prefix ./glean run lint
       - run:
           name: Audit dependency tree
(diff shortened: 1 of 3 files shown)


chutten (approved):
Does any contribution documentation need to be amended to advise contributors they're expected to keep the deps audited?

brizental:
> Does any contribution documentation need to be amended to advise contributors they're expected to keep the deps audited?

At this point we don't have any developer documentation on adding new dependencies... But, I think for now the CI failing if auditing fails is good enough.

PR #822, opened 2021-10-11
Bug 1734730 - Fix docs build and add build CI job

brizental:
This probably got broken during one of dependabots updates, so now I added a CI job for us to know when dependabot breaks the build commands.

### DIFF
--- a/glean/package.json
+++ b/glean/package.json
@@ -63,7 +63,7 @@
     "build:lib": "tsc -p ./tsconfig/lib.json",
     "build:types": "tsc -p ./tsconfig/types.json",
     "build:qt": "rm -rf dist/qt && webpack --config webpack.config.qt.js && ../bin/prepare-qml-module.sh",
-    "build:docs": "rm -rf dist/docs && typedoc src/ --out dist/docs --tsconfig tsconfig/docs.json --theme minimal",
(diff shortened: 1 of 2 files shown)


glean/package.json
```diff
@@ -63,7 +63,7 @@
     "build:lib": "tsc -p ./tsconfig/lib.json",
     "build:types": "tsc -p ./tsconfig/types.json",
     "build:qt": "rm -rf dist/qt && webpack --config webpack.config.qt.js && ../bin/prepare-qml-module.sh",
-    "build:docs": "rm -rf dist/docs && typedoc src/ --out dist/docs --tsconfig tsconfig/docs.json --theme minimal",
+    "build:docs": "rm -rf dist/docs && typedoc src/**/*.ts src/**/**/*.ts src/**/**/**/*.ts --out dist/docs --tsconfig tsconfig/docs.json",
```
badboy, line 66: Is this necessary? `**` usually expands to one-or-more-levels already, so `**/**` is the same as `**`

brizental: Yeah I know... Not in this case though, lol. Not sure what is up, probably a bug in `typedoc` itself.

badboy: Guess it depends on who expands that.

badboy: Ah, see https://typedoc.org/guides/options/#entrypointstrategy
the strategy changed in 0.21 from `expand` to `resolve`.
Could that be the change why it broke?

PR #825, opened 2021-10-12
Bug 1734879 - Accept arch and os version as init params on Qt

brizental:
I know we discussed making these values enums maybe, but for os version that is just not possible and for architecture that would be hard, see https://doc.qt.io/qt-5/qsysinfo.html#currentCpuArchitecture, we don't have a list of all the possible values returned by the Qt function that retrieves that value.

I think this is fine though, because different from what we thought yesterday, GLAM does not use the architecture field for creating filters, it uses only OS and channel.

### DIFF
--- a/glean/src/core/config.ts
+++ b/glean/src/core/config.ts
@@ -29,6 +29,16 @@ export interface ConfigurationInterface {
   plugins?: Plugin[],
   // The HTTP client implementation to use for uploading pings.
   httpClient?: Uploader,
+  // Qt-only fields
+  //
+  // These values are not easily accessible from QML,
+  // so we expose them as init fields for the caller to fill them out.
+  //
+  // The architecture of the device (e.g. "arm", "x86").
+  readonly architecture?: string,
+  // The user-visible version of the operating system (e.g. "1.2.3").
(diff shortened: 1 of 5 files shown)


glean/src/core/config.ts
```diff
@@ -41,6 +51,12 @@ export class Configuration implements ConfigurationInterface {
   readonly appDisplayVersion?: string;
   // The server pings are sent to.
   readonly serverEndpoint: string;
+  // The architecture of the device (e.g. "arm", "x86").
+  readonly architecture?: string;
+  // The user-visible version of the operating system (e.g. "1.2.3").
+  // If the version detection fails, this metric gets set to Unknown.
```
badboy, line 57: As this is the docs exposed to users it reads a bit weird.
There's no "version detection" done by Glean, so there's nothing that can fail.
Should this be more like "If not set, this metrics gets set to `Unknown`?

brizental: Yeah, 🍝 copy pasta right there.

brizental, line 57:
```suggestion
```

PR #863, opened 2021-10-18
No bug - Make CI jobs depend on one another

brizental:
So that we don't run CI jobs when it does not really make sense to run them.

+ Change `samples-*` job names so they show up in an organized manner on the GitHub interface. This is just me being annoying.


mdboom (approved):
Reducing CI costs, ftw!

PR #967, opened 2021-11-18
Bug 1712914 - Remove debug options from configuration object

brizental:
The reason for this change is to be consistent with the other SDKs. None of them allow to pass debug options on init.

Best to make this change now that we don't have that many users.

  - This was never documented specifically because we did not want people to use it.

### DIFF
--- a/glean/tests/unit/plugins/encryption.spec.ts
+++ b/glean/tests/unit/plugins/encryption.spec.ts
@@ -89,9 +89,6 @@ describe("PingEncryptionPlugin", function() {
         plugins: [
           new PingEncryptionPlugin(await exportJWK(publicKey))
         ],
-        debug: {
-          logPings: true
-        },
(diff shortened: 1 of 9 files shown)


glean/tests/unit/plugins/encryption.spec.ts
```diff
@@ -93,9 +93,6 @@ describe("PingEncryptionPlugin", function() {
         plugins: [
           new PingEncryptionPlugin(await exportJWK(publicKey))
         ],
-        debug: {
```
Dexterp37, line 96: Don't we need a setLogPings(true) ?

brizental: Nah, logging was just a leftover from development on this one. It does nothing ot the test.

Dexterp37 (approved):

PR #983, opened 2021-11-22
Bug 1742063 - Use standalone uploader in uploader tests

brizental:

### DIFF
--- a/glean/src/core/context.ts
+++ b/glean/src/core/context.ts
@@ -226,6 +226,16 @@ export class Context {
   }
 
   static get platform(): Platform {
+    if (typeof Context.instance._platform === "undefined") {
+      log(
+        LOG_TAG,
+        [
+          "Attempted to access Context.platform before it was set. This may cause unexpected behaviour.",
+        ],
+        LoggingLevel.Error
+      );
+    }
+
     return Context.instance._platform;
   }
 }
--- a/glean/src/core/context.ts
+++ b/glean/src/core/context.ts
@@ -7,6 +7,7 @@ import type MetricsDatabase from "./metrics/database.js";
 import type EventsDatabase from "./metrics/events_database/index.js";
 import type PingsDatabase from "./pings/database.js";
 import type ErrorManager from "./error/index.js";
+import type Platform from "../platform/index.js";
 import Dispatcher from "./dispatcher.js";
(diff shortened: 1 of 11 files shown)


glean/src/core/context.ts
```diff
@@ -28,6 +29,7 @@ export class Context {
   private static _instance?: Context;
 
   private _dispatcher: Dispatcher;
+  private _platform!: Platform;
```
chutten, line 32: I'm not typescript guru, but doesn't this mean "not-nullish"? But it is guaranteed to be nullish until set (and there's nothing in the setter forbidding it from being set to nullish), so maybe I don't understand what this is about.

glean/tests/unit/core/upload/index.spec.ts
```diff
@@ -98,64 +114,77 @@ describe("PingUploader", function() {
...
     await fillUpPingsDatabase(MAX_PINGS_PER_INTERVAL + 5);
-    await Glean.shutdown();
+    await uploader.shutdown();
 
     assert.strictEqual(httpClient.count, MAX_PINGS_PER_INTERVAL);
   });
 
   it("correctly deletes pings when upload is successfull", async function() {
+    const uploader = new PingUploader(new Configuration(), pingsDatabase);
+    pingsDatabase.attachObserver(uploader);
```
chutten, line 139: Since this seems common across subtests here, should it be in a `beforeall` (or whatever it's called)?

chutten (approved):
Landable as-is, though I do have two (potentially-stupid) questions

glean/tests/unit/core/upload/index.spec.ts
brizental: In this case we prefer to create a new uploader per test case for isolation purposes. Also, even though it is _similar_ in each test case the configuration is actually different in most of them :)

glean/src/core/context.ts
brizental: Yeah... I think you are right. So the thing is, the platform _should_ never be null, because it is always set before Glean is initialized on the exposed API (https://github.com/mozilla/glean.js/blob/main/glean/src/index/base.ts#L33-L49). However, better be safe than sorry, I'll add a check there.

PR #1006, opened 2021-12-01
Bug 1716955: Initial implementation of the rate metric

ChinYing-Li:


ChinYing-Li:
Hello, 
I just realized that in the glean library (https://github.com/mozilla/glean), there's no unit test for the rate metric. Is this by purpose?

Dexterp37:
> Hello, I just realized that in the glean library (https://github.com/mozilla/glean), there's no unit test for the rate metric. Is this by purpose?

There's a [smoke test here](https://github.com/mozilla/glean/blob/4e56b157196350a4a2b605423dbfd1ce7c42d160/glean-core/rlb/src/private/rate.rs#L74) and a [schema test here](https://github.com/mozilla/glean/blob/1a8ad8003c220281e2f1cddc3b6aeabe7875cd35/glean-core/rlb/tests/schema.rs#L101). I believe that given the time pressure, no other test was added.

ChinYing-Li:
The PR is now ready for review; any suggestion is appreciated!

glean/src/core/metrics/types/rate.ts
```diff
@@ -0,0 +1,209 @@
...
+      })(amount);
+
+      await Context.metricsDatabase.transform(this, transformFn);
+    });
+  }
+
+  /**
+   * Test-only API.**
+   *
+   * Gets the currently stored value as a string.
```
chutten, line 187: It doesn't appear to return a string

glean/src/core/metrics/types/rate.ts
```diff
@@ -0,0 +1,209 @@
...
+   *
+   * # Note
+   *
+   * this function will return the unencoded URL for convenience.
```
chutten, line 191: I'm not sure this comment is correct for this type.

glean/tests/unit/core/metrics/rate.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+import RateMetricType from "../../../../src/core/metrics/types/rate";
+
+describe("RateMetric", function() {
+  const testAppId = `gleanjs.test.${this.title}`;
+
+  it("smoke test for rate metric", async function() {
+    await Glean.testResetGlean(testAppId);
+    Glean.setUploadEnabled(true);
+
+    const metric = new RateMetricType({
```
chutten, line 19: You add a test rate metric in the metrics.yaml, but here you manually create one. I don't understand why you don't use the rate metric from the metrics.yaml and save yourself the boilerplate.

chutten (commented):
This is looking good to me so far. I'll bring in @brizental to make sure I'm not allowing strange patterns in just because I'm unfamiliar with how metric types are implemented in the Glean JS SDK.

glean/src/core/metrics/types/rate.ts
```diff
@@ -0,0 +1,209 @@
...
+import { isNumber, isObject } from "../../utils.js";
+import type { JSONValue } from "../../utils.js";
+import { ErrorType } from "../../error/error_type.js";
+
+const LOG_TAG = "core.metrics.RateMetricType";
+
+export type RateInternalRepresentation = {
+  // numerator
+  numerator: number,
+  // The denominator
```
brizental, line 19: nit: These two comments are redundant.

glean/src/core/metrics/types/rate.ts
```diff
@@ -0,0 +1,209 @@
...
+  /**
+   * Increases the numerator by amount.
+   *
+   * # Note
+   *
+   * Logs an error if the `amount` is negative.
+   *
+   * @param amount The amount to increase by. Should be non-negative.
+   */
+  add_to_numerator(amount: number): void {
```
brizental, line 85: nit: 
```suggestion
  addToNumerator(amount: number): void {
```

glean/src/core/metrics/types/rate.ts
```diff
@@ -0,0 +1,209 @@
...
+  /**
+   * Increases the denominator by amount.
+   *
+   * # Note
+   *
+   * Logs an error if the `amount` is negative.
+   *
+   * @param amount The amount to increase by. Should be non-negative.
+   */
+  add_to_denominator(amount: number): void {
```
brizental, line 139: nit:
```suggestion
  addToDenominator(amount: number): void {
```

glean/tests/unit/core/metrics/rate.spec.ts
brizental: That metrics.yaml is only used by the integration tests. On the unit tests we are required to instantiate the metrics, so this is correct :)

glean/tests/unit/core/metrics/rate.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+    assert.strictEqual(await metric.testGetNumRecordedErrors(ErrorType.InvalidValue), 2);
+
+    assert.deepStrictEqual(await metric.testGetValue("aPing"), {numerator: 0, denominator: 0});
+
+    metric.add_to_numerator(22);
+    metric.add_to_denominator(7);
+
+    assert.deepStrictEqual(await metric.testGetValue("aPing"), {numerator: 22, denominator: 7});
+  });
+});
```
brizental, line 44: nit:
```suggestion
});

```

glean/tests/unit/core/metrics/rate.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+import { ErrorType } from "../../../../src/core/error/error_type";
+
+import Glean from "../../../../src/core/glean";
+import { Lifetime } from "../../../../src/core/metrics/lifetime";
+import RateMetricType from "../../../../src/core/metrics/types/rate";
+
+describe("RateMetric", function() {
+  const testAppId = `gleanjs.test.${this.title}`;
+
+  it("smoke test for rate metric", async function() {
```
brizental, line 15: Hm, I understand this is the only test we have for this metrics on glean-core. However, since we are not time constrained here I would prefer to have a more comprehensive test suite.

I am fine if you'd prefer to make that a follow-up PR :) And feel free to ask any questions with regards to the tests you want to add. 

You can also base yourself off of other metrics suites, such as the [BooleanMetric](https://github.com/mozilla/glean.js/blob/main/glean/tests/unit/core/metrics/boolean.spec.ts). At least the four tests you see there we have for all metric types, so this is a good baseline. Since this is a more complex metric type, there might be other tests that make sense here as well.

glean/package-lock.json
```diff
@@ -8799,40 +8799,6 @@
         }
       }
     },
-    "@typescript-eslint/scope-manager": {
```
brizental, line 8802: Hm, let's please drop these changes from this PR.

brizental (changes requested):
Hey @ChinYing-Li this is looking really great. I like the way you implemented the metric, very clean.

I left some comments. Implementation looks really good. I would like to have more tests as well. 

Once we are done with these changes we will also need to update [the docs](https://mozilla.github.io/glean/book/reference/metrics/rate.html). Feel free to simply file a follow up for that if you don't want to work on it.

glean/tests/unit/core/metrics/rate.spec.ts
brizental: I just remembered this metric may also record errors, so we should cover that in the test suite as well. See example in the [CounterMetric](https://github.com/mozilla/glean.js/blob/main/glean/tests/unit/core/metrics/counter.spec.ts#L82-L101) test suite.

ChinYing-Li: Just added a few test cases, taking reference from the Counter Metric. Let me know if you think there's any cases I missed.

glean/package-lock.json
brizental: Looks like the changes are still here, so I am unresolving. Perhaps rebasing the PR will make this change unnecessary.

glean/tests/unit/core/metrics/rate.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+    metric.addToDenominator(0);
+
+    assert.strictEqual(await metric.testGetNumRecordedErrors(ErrorType.InvalidValue), 0);
+    assert.deepStrictEqual(await metric.testGetValue("aPing"), { numerator: 0, denominator: 0 });
+
+    metric.addToNumerator(-1);
+    metric.addToDenominator(-1);
+
+    assert.strictEqual(await metric.testGetNumRecordedErrors(ErrorType.InvalidValue), 2);
+    assert.deepStrictEqual(await metric.testGetValue("aPing"), {numerator: 0, denominator: 0});
```
brizental, line 106: Hm, this is not the behaviour I would expect. If I try to set incorrect values, `testGetValue` should return `undefined`. I think this is problem on the implementation, good thing we are adding tests :D 

Indeed, [here](https://github.com/mozilla/glean/blob/main/glean-core/src/metrics/rate.rs#L117) you should return early after recording the error. [Like it is done in glean-core](https://github.com/mozilla/glean/blob/main/glean-core/src/metrics/rate.rs#L59-L68). Same goes for the `addToDenominator` function.

brizental: Also, let's add checks for error recordings here.

glean/src/core/metrics/types/rate.ts
```diff
@@ -0,0 +1,207 @@
...
+      await Context.metricsDatabase.transform(this, transformFn);
+    });
+  }
+
+  /**
+   * Increases the denominator by amount.
+   *
+   * # Note
+   *
+   * Logs an error if the `amount` is negative.
```
brizental, line 133: nit:
```suggestion
   * Records an `InvalidValue` an error if the `amount` is negative.
```

glean/tests/unit/core/metrics/rate.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+      disabled: false
+    });
+
+    metric.addToNumerator(10);
+    const expectedValue = { numerator: 10, denominator: 0 };
+
+    assert.deepStrictEqual(await metric.testGetValue("aPing"), expectedValue);
+    assert.deepStrictEqual(await metric.testGetValue("twoPing"), expectedValue);
+    assert.deepStrictEqual(await metric.testGetValue("threePing"), expectedValue);
+  });
```
brizental, line 85: Let's please use both `addToDenominator` and `addToNumerator` APIs in all there three first test cases.

glean/tests/unit/core/metrics/rate.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+    let expectedValue = { numerator: 2, denominator: 0 };
+    assert.deepStrictEqual(await metric.testGetValue("aPing"), expectedValue);
+    assert.deepStrictEqual(await metric.testGetValue("twoPing"), expectedValue);
+    assert.deepStrictEqual(await metric.testGetValue("threePing"), expectedValue);
+
+    metric.addToDenominator(2);
+    expectedValue = { numerator: 2, denominator: 2 };
+    assert.deepStrictEqual(await metric.testGetValue("aPing"), expectedValue);
+    assert.deepStrictEqual(await metric.testGetValue("twoPing"), expectedValue);
+    assert.deepStrictEqual(await metric.testGetValue("threePing"), expectedValue);
```
brizental, line 133: Let's add two more transformations here. To check that transforming the numerator and denominator after they were set once, also works as expected.

glean/tests/unit/core/metrics/rate.spec.ts
brizental: Unresolving because the new line at the end is still missing :)

glean/src/core/metrics/types/rate.ts
```diff
@@ -0,0 +1,207 @@
...
+  constructor(meta: CommonMetricData) {
+    super("rate", meta);
+  }
+
+  /**
+   * Increases the numerator by amount.
+   *
+   * # Note
+   *
+   * Logs an error if the `amount` is negative.
```
brizental, line 79:
```suggestion
   * Records an `InvalidValue` error if the `amount` is negative.
```

brizental (changes requested):
Thanks for adding the tests! Not much is missing, left some inline commments and through the tests found a bug in the implementation. Easy to fix though.

glean/package-lock.json
ChinYing-Li: Yes rebasing the PR eliminates the diff.

glean/tests/unit/core/metrics/rate.spec.ts
ChinYing-Li: The [smoke test in glean-core](https://github.com/mozilla/glean/blob/4e56b157196350a4a2b605423dbfd1ce7c42d160/glean-core/rlb/src/private/rate.rs#L74) actually expects "setting negative value" to be no-op. Please let me know if you'd still like to have "invalidate the metric if invalid value is passed" as the expected behavior.

brizental: Indeed, but it's not returning `undefined` (or `None` in their case), [because the metric was set to `0` initially](https://github.com/mozilla/glean/blob/4e56b157196350a4a2b605423dbfd1ce7c42d160/glean-core/rlb/src/private/rate.rs#L92-L93).

ChinYing-Li: The test case here also set to 0 initially; did I miss anything?

brizental: Woof, I am sorry _I_ was the one missing something. I missed the setting to zero.

Either way, it's still important to return early. I would even remove the lines where we set to zero, just to be sure we catch any problem with this behaviour in the future.

glean/src/core/metrics/types/rate.ts
```diff
@@ -0,0 +1,202 @@
...
+/*
+ * A rate metric.
+ *
+ * Used to determine the proportion of things via two counts:
+ * * A numerator defining the amount of times something happened,
+ * * A denominator counting the amount of times someting could have happened.
+ *
+ * Both numerator and denominator can only be incremented, not decremented.
+ */
+
```
brizental, line 61: nit: Extra blank line.
```suggestion
```

glean/src/core/metrics/types/rate.ts
```diff
@@ -0,0 +1,202 @@
...
+    Context.dispatcher.launch(async () => {
+      if (!this.shouldRecord(Context.uploadEnabled)) {
+        return;
+      }
+
+      if (amount < 0) {
+        await Context.errorManager.record(
+          this,
+          ErrorType.InvalidValue,
+          `Added negative value ${amount} to numerator.`
```
brizental, line 141:
```suggestion
          `Added negative value ${amount} to denominator.`
```

glean/src/core/metrics/types/rate.ts
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  /**
+   * Test-only API.**
+   *
+   * Gets the currently stored value as an object.
+   *
+   * # Note
+   *
+   * this function will return the RateInternalPresentation for convenience.
```
brizental, line 184:
```suggestion
   * This function will return the RateInternalPresentation for convenience.
```

brizental (approved):
Some final nits, otherwise, looks god!

Thanks so much for the work here @ChinYing-Li 🎉

brizental:
Oh, yeah. Also, please add a changelog entry for this.

glean/src/core/metrics/types/rate.ts
ChinYing-Li: Oof, so sorry for the stupid mistake! Thank you for pointing this out.

brizental: Nothing to be sorry about. Happens to all of us :) That's why we have reviews!

CHANGELOG.md
```diff
@@ -8,8 +8,10 @@
 
 * [#984](https://github.com/mozilla/glean.js/pull/984): BUGFIX: Return correct upload result in case an error happens while building a ping request.
 * [#988](https://github.com/mozilla/glean.js/pull/988): BUGFIX: Enforce rate limitation at upload time, not at ping submission time.
+
   * Note: This change required a big refactoring of the internal uploading logic.
 * [#994](https://github.com/mozilla/glean.js/pull/994): Automatically restart ping upload once the rate limit window is ended.
+* [#1006](https://github.com/mozilla/glean.js/pull/1006): Implement the rate metric.
```
brizental, line 14: Actually this should go on the `Unreleased changes` section :)

PR #1051, opened 2021-12-22
Bug 1746375 - Apply various changes to make Glean.js bundle smaller

brizental:
The changes that most significantly impacted the size of the bundle were in https://github.com/mozilla/glean.js/commit/64a747c2654346e32676346922f4364597ea291b. More details in the commit message.

I still think there is more to do here, but sending this in as a start.

As mentioned in https://bugzilla.mozilla.org/show_bug.cgi?id=1746375#c2, the `fflate` dependecy takes up a lot of space. We may reconsider it's usage, however I am not sure there is anything smaller out there. 

I also learned that we need to be mindful of adding new metric types. Every new metric type has a size impact, even when it's not being used due to the [METRIC_MAP](https://github.com/mozilla/glean.js/blob/612c82d42c13ecadb1db1346b7237d5f0d51f482/glean/src/core/metrics/utils.ts#L28-L41). See more details on https://bugzilla.mozilla.org/show_bug.cgi?id=1746375#c1. Even though the type itself may not be imported at least the Metric part (which has the payload validations) is. There might be a way to address this that I am not seeing right now... This may revive the idea of "metric types as plugins". How to make a metric type completely standalone when it depends on the MetricsDatabase and vice-versa?

Finally, one "fun" fact: The EventMetricType now is part of the minimum bundle, because of the glean.restarted event.


brizental:
Hum, seems like I broke the Web Storage. I'll look at it tomorrow.

moz-glean:
# Build size report

Merging https://github.com/mozilla/glean.js/pull/1051 into [main](https://github.com/mozilla/glean.js) will:
* **Decrease** the size of full **Web Extension** bundle build by `-8%`.
* **Decrease** the size of full **Website** bundle build by `-8%`.
* **Decrease** the size of full **Node.js** bundle build by `-7%`.
* **Decrease** the size of full **Qt/QML** bundle build by `-3%`.

---

|--:|:---:|:---:|:---:|
| **Web Extension** |
| core only | 67 KB | 55 KB | 📉 12 KB |
| full bundle | 89 KB | 82 KB | 📉 7.2 KB |
| **Website** |
| core only | 68 KB | 56 KB | 📉 12 KB |
| full bundle | 90 KB | 83 KB | 📉 7.4 KB |
| **Node.js** |
| core only | 66 KB | 54 KB | 📉 12 KB |
| full bundle | 94 KB | 87 KB | 📉 7.2 KB |
| **Qt/QML** |
| core only | 70 KB | 68 KB | 📉 2.2 KB |
| full bundle | 70 KB | 68 KB | 📉 2.2 KB |

101arrowz:
With regards to the `fflate` size issue: you might want to follow https://github.com/101arrowz/fflate/issues/115. If you have a nice reproduction similar to https://github.com/parcel-bundler/parcel/issues/7481, feel free to post it.

chutten (approved):
Seems like most of these changes (except the loss of decorators) aren't going to make the sdk any more difficult to maintain, which is nice.

A couple of questions:
* Can you please include in the commit messages the reasons why these changes make for a smaller bundle? The decorators one is an excellent example with "If we don't do this, typescript adds auxillary code which sums to 5kB"... but what about `namespace`? Why are those smaller? Why don't we use them for everything? Why are getters on the Glean singleton so expensive?
* Docs: These changes look like they're an ongoing shift in how we should think about package sizes. e.g. The next time we design a big component, we'll need to think about whether it should be `namespace` or not. Can the rationale behind these changes be encoded somehow so it'll be hard for future us to undo this work? Either as contribution docs or PR checklist items or something?

brizental:
> Docs: These changes look like they're an ongoing shift in how we should think about package sizes. e.g. The next time we design a big component, we'll need to think about whether it should be namespace or not. Can the rationale behind these changes be encoded somehow so it'll be hard for future us to undo this work? Either as contribution docs or PR checklist items or something?

Thanks, this is a great suggestion. Filed [a bug](
https://bugzilla.mozilla.org/show_bug.cgi?id=1748475) about as I think I am still learning about the subject myself. PR checklist would be good though. I like and use the one we have a lot.

PR #1065, opened 2022-01-03
Bug 1746375 - Dynamically add metric constructors to metric map 

brizental:
Sending this in as a draft because I would like to get some feedback on the approach before moving on. More details on the commit messages, but this is a part of attempting to make the Glean.js bundle size smaller.

The changes do not make as a drastic a dent in the size of the bundle as #1051, however these changes are guards against making the bundle much larger in the future. As we add more metric types the bundle wouldn't get bigger.

My main concern is outlined in aaa1ce0. 

> I am not entirely sure about this change, because it means that if a
> given metric is in the database, but was not added to the metric map
> (think if a metric is removed from the code, but has not been collected
> in a ping once the new code lands) it will be read as invalid by Glean.

This could cause some surprising behaviour for users. My question is: are the benefits of a smaller bundle worth this kinda weird, but hopefully tiny cornercase?

I have also filed [a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1748246) so we can instrument whenever we find invalid data in the database so we can know if this is really tiny or not.

Note that https://github.com/mozilla/glean.js/commit/7b5d7a2737ec84d025f7877f0aec78cfab9c598b is an attempt to make the cornercase not as big.

   - https://bugzilla.mozilla.org/show_bug.cgi?id=1748475

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 [Full changelog](https://github.com/mozilla/glean.js/compare/v0.30.0...main)
 
+* [#1065](https://github.com/mozilla/glean.js/pull/1065): Delete minimal amount of data when invalid data is found while collecting ping.
+  * Previous behaviour was to delete the whole ping when invalid data was found on the database,
+  new behaviour only deletes the actually invalid data and leave the rest of the ping intact.
+* [#1065](https://github.com/mozilla/glean.js/pull/1065): Only import metric types into the library when they are used either by the user or Glean itself.
+  * Previously the code required to deserialize metric data from the database was always imported by the library even if the metric type was never used by the client. This effort will decrete the size of the Glean.js bundles that don't import all the metric types.
 # v0.30.0 (2022-01-10)
 
 [Full changelog](https://github.com/mozilla/glean.js/compare/v0.29.0...v0.30.0)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
   * Previous behaviour was to delete the whole ping when invalid data was found on the database,
   new behaviour only deletes the actually invalid data and leave the rest of the ping intact.
 * [#1065](https://github.com/mozilla/glean.js/pull/1065): Only import metric types into the library when they are used either by the user or Glean itself.
-  * Previously the code required to deserialize metric data from the database was always imported by the library even if the metric type was never used by the client. This effort will decrete the size of the Glean.js bundles that don't import all the metric types.
+  * Previously the code required to deserialize metric data from the database was always imported by the library even if the metric type was never used by the client. This effort will decrease the size of the Glean.js bundles that don't import all the metric types.
 # v0.30.0 (2022-01-10)
 
 [Full changelog](https://github.com/mozilla/glean.js/compare/v0.29.0...v0.30.0)
--- a/glean/src/core/context.ts
+++ b/glean/src/core/context.ts
@@ -48,9 +48,9 @@ export class Context {
   // Whether or not Glean is in testing mode.
   private _testing = false;
   // A map of metric types and their constructors.
-  // This map is dinamically filled everytime a metric type is constructed.
+  // This map is dynamically filled everytime a metric type is constructed.
   //
-  // If a metric is not on this map it cannot be serialized from the database.
+  // If a metric is not in this map it cannot be deserialized from the database.
   private _supportedMetrics: {
     [type: string]: new (v: unknown) => Metric<JSONValue, JSONValue>
   } = {};
--- a/glean/src/core/context.ts
+++ b/glean/src/core/context.ts
@@ -7,6 +7,8 @@ import type EventsDatabase from "./metrics/events_database/index.js";
 import type PingsDatabase from "./pings/database.js";
 import type ErrorManager from "./error/index.js";
 import type Platform from "../platform/index.js";
+import type { Metric } from "./metrics/metric.js";
+import type { JSONValue } from "./utils.js";
 import Dispatcher from "./dispatcher.js";
 import log, { LoggingLevel } from "./log.js";
 import type { Configuration } from "./config.js";
@@ -45,6 +47,13 @@ export class Context {
   private _initialized = false;
   // Whether or not Glean is in testing mode.
   private _testing = false;
+  // A map of metric types and their constructors.
+  // This map is dinamically filled everytime a metric type is constructed.
+  //
+  // If a metric is not on this map it cannot be serialized from the database.
+  private _supportedMetrics: {
+    [type: string]: new (v: unknown) => Metric<JSONValue, JSONValue>
+  } = {};
 
   // The moment the current Glean.js session started.
   private _startTime: Date;
@@ -241,5 +250,28 @@ export class Context {
 
   static isPlatformSet(): boolean {
     return !!Context.instance._platform;
+  }
+
+  static getSupportedMetric(type: string): (new (v: unknown) => Metric<JSONValue, JSONValue>) | undefined {
+    return Context.instance._supportedMetrics[type];
+  }
+
+  /**
+   * Adds a new constructor to the supported metrics map.
+   *
+   * If the metric map already contains this constructor, this is a no-op.
+   *
+   * @param type A string identifying the given metric type.
+   * @param ctn The metric constructor.
+   */
+  static addSupportedMetric(
+    type: string,
+    ctn: new (v: unknown) => Metric<JSONValue, JSONValue>
+  ): void {
+    if (type in Context.instance._supportedMetrics) {
+      return;
+    }
+
+    Context.instance._supportedMetrics[type] = ctn;
   }
 }
--- a/glean/src/core/context.ts
+++ b/glean/src/core/context.ts
@@ -262,16 +262,16 @@ export class Context {
    * If the metric map already contains this constructor, this is a no-op.
    *
    * @param type A string identifying the given metric type.
-   * @param ctn The metric constructor.
+   * @param ctor The metric constructor.
    */
   static addSupportedMetric(
     type: string,
-    ctn: new (v: unknown) => Metric<JSONValue, JSONValue>
+    ctor: new (v: unknown) => Metric<JSONValue, JSONValue>
   ): void {
     if (type in Context.instance._supportedMetrics) {
       return;
     }
 
-    Context.instance._supportedMetrics[type] = ctn;
+    Context.instance._supportedMetrics[type] = ctor;
   }
 }
--- a/glean/src/core/glean.ts
+++ b/glean/src/core/glean.ts
@@ -29,7 +29,7 @@ namespace Glean {
   // The below properties are exported for testing purposes.
   //
   // Instances of Glean's core metrics.
-  export const coreMetrics = new CoreMetrics();
+  export let coreMetrics = new CoreMetrics();
   // Instances of Glean's core pings.
   export const corePings = new CorePings();
   // An instance of the ping uploader.
@@ -463,6 +463,10 @@ namespace Glean {
     uploadEnabled = true,
     config?: ConfigurationInterface
   ): Promise<void> {
+    // Core metrics need tp be re-initialized so that
+    // the supportedMetrics map is re-created.
+    coreMetrics = new CoreMetrics();
+
     Context.testing = true;
 
     setPlatform(TestPlatform);
--- a/glean/src/core/glean.ts
+++ b/glean/src/core/glean.ts
@@ -463,7 +463,7 @@ namespace Glean {
     uploadEnabled = true,
     config?: ConfigurationInterface
   ): Promise<void> {
-    // Core metrics need tp be re-initialized so that
+    // Core metrics need to be re-initialized so that
     // the supportedMetrics map is re-created.
     coreMetrics = new CoreMetrics();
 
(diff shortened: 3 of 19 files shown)


moz-glean:
# Build size report

Merging https://github.com/mozilla/glean.js/pull/1065 into [main](https://github.com/mozilla/glean.js) will:
* **Decrease** the size of full **Web Extension** bundle build by `-3%`.
* **Decrease** the size of full **Website** bundle build by `-3%`.
* **Decrease** the size of full **Node.js** bundle build by `-3%`.
* **Leave** the size of full **Qt/QML** bundle **unchanged**.

---

|--:|:---:|:---:|:---:|
| **Web Extension** |
| core only | 67 KB | 57 KB | 📉 9.8 KB |
| full bundle | 89 KB | 86 KB | 📉 3.1 KB |
| **Website** |
| core only | 68 KB | 58 KB | 📉 9.8 KB |
| full bundle | 90 KB | 87 KB | 📉 3.1 KB |
| **Node.js** |
| core only | 66 KB | 56 KB | 📉 9.8 KB |
| full bundle | 94 KB | 91 KB | 📉 3.1 KB |
| **Qt/QML** |
| core only | 70 KB | 71 KB | 📈 302 bytes |
| full bundle | 70 KB | 71 KB | 📈 302 bytes |

moz-glean:
# Build size report

Merging https://github.com/mozilla/glean.js/pull/1065 into [main](https://github.com/mozilla/glean.js) will:
* **Leave** the size of full **Website** bundle **unchanged**.
* **Leave** the size of full **Node.js** bundle **unchanged**.
* **Leave** the size of full **Qt/QML** bundle **unchanged**.

---

|--:|:---:|:---:|:---:|
| **Web Extension** |
| core only | 55 KB | 54 KB | 📉 1.9 KB |
| full bundle | 82 KB | 82 KB | 📉 102 bytes |
| **Website** |
| core only | 56 KB | 54 KB | 📉 1.9 KB |
| full bundle | 83 KB | 83 KB | 📉 100 bytes |
| **Node.js** |
| core only | 54 KB | 52 KB | 📉 1.9 KB |
| full bundle | 87 KB | 87 KB | 📉 97 bytes |
| **Qt/QML** |
| core only | 68 KB | 68 KB | 📈 291 bytes |
| full bundle | 68 KB | 68 KB | 📈 291 bytes |

glean/src/core/context.ts
```diff
@@ -45,6 +47,13 @@ export class Context {
   private _initialized = false;
   // Whether or not Glean is in testing mode.
   private _testing = false;
+  // A map of metric types and their constructors.
+  // This map is dinamically filled everytime a metric type is constructed.
```
badboy, line 51:
```suggestion
  // This map is dynamically filled everytime a metric type is constructed.
```

(should be changed in the commit message too)

glean/src/core/context.ts
```diff
@@ -45,6 +47,13 @@ export class Context {
...
+  //
+  // If a metric is not on this map it cannot be serialized from the database.
```
badboy, line 53:
```suggestion
  // If a metric is not in this map it cannot be serialized from the database.
```

Also it should be "deserialized", shouldn't it?

glean/src/core/context.ts
```diff
@@ -242,4 +251,27 @@ export class Context {
...
+   * Adds a new constructor to the supported metrics map.
+   *
+   * If the metric map already contains this constructor, this is a no-op.
+   *
+   * @param type A string identifying the given metric type.
+   * @param ctn The metric constructor.
+   */
+  static addSupportedMetric(
+    type: string,
+    ctn: new (v: unknown) => Metric<JSONValue, JSONValue>
```
badboy, line 269: I haven't seen `ctn` for `constructor` before. `ctor` would be more common.

glean/src/core/glean.ts
```diff
@@ -458,6 +458,10 @@ namespace Glean {
     uploadEnabled = true,
     config?: ConfigurationInterface
   ): Promise<void> {
+    // Core metrics need tp be re-initialized so that
```
badboy, line 461:
```suggestion
    // Core metrics need to be re-initialized so that
```

badboy (commented):
How does the final bundle actually get built? Is this under control of the user embedding Glean?
Will this be determined by the metric types used in metrics.yaml?

It might be worth listing out when we lose what data.
E.g. something like

Version A: initial. metric types [string, bool, datetime]. Records all 3. No ping sent.
Version A+1: removed datetime. metric types [string, bool]. Sends ping, contains string and bool, old datetime is deleted from the database.

(Am I correct with that example?)

My first reaction is: I think this is fine then. Version A+1 wouldn't record a new value. In the current model we would get a value from a previous run, but no new ones anyway.
Losing that one-time stale value doesn't seem to dramatic.

brizental:
> How does the final bundle actually get built? Is this under control of the user embedding Glean?

So in the benchmarks example we are using webpack. We are not in control of how the user builds it, but the most common way to use Glean would be to use a bundler such as webpack or rollup.

We also don't provide a pre-built library, so I would say that might be the only way to use it.

> Will this be determined by the metric types used in metrics.yaml?

Exactly. The glean_parser generated template imports the metric types defined in the metrics.yaml and instantiates the metric. With this change, unless a given metric type's code is imported, none of it is added to the final bundle. Which was not the case before, because we were always importing at least the Metric part used for database deserialization.

> It might be worth listing out when we lose what data.
> E.g. something like
>
> Version A: initial. metric types [string, bool, datetime]. Records all 3. No ping sent.
> Version A+1: removed datetime. metric types [string, bool]. Sends ping, contains string and bool, old datetime is deleted from the database.

Perfect, this is right.

> (Am I correct with that example?)
> 
> My first reaction is: I think this is fine then. Version A+1 wouldn't record a new value. In the current model we would get a value from a previous run, but no new ones anyway.
> Losing that one-time stale value doesn't seem to dramatic.

Yeah I tend to agree with this, but I am working on assumptions here. I was wondering if you have seen such a pattern in other places.

glean/src/core/context.ts
brizental: Heh, I knew there was a proper short way to write it... Later I also realize `ctn` is not even correct, it would have to be `cnt`.

brizental:
> Version A: initial. metric types [string, bool, datetime]. Records all 3. No ping sent.
> Version A+1: removed datetime. metric types [string, bool]. Sends ping, contains string and bool, old datetime is deleted from the database.

Actually in this case it would still report the datetime, because that metric is used internally by Glean and thus always added to the metric map.

badboy:
> > Version A: initial. metric types [string, bool, datetime]. Records all 3. No ping sent.
> > Version A+1: removed datetime. metric types [string, bool]. Sends ping, contains string and bool, old datetime is deleted from the database.
> 
> Actually in this case it would still report the datetime, because that metric is used internally by Glean and thus always added to the metric map.

ehehehe, pick a metric type that's not used by Glean then.

> > Will this be determined by the metric types used in metrics.yaml?
> 
> Exactly. The glean_parser generated template imports the metric types defined in the metrics.yaml and instantiates the metric. With this change, unless a given metric type's code is imported, none of it is added to the final bundle. Which was not the case before, because we were always importing at least the Metric part used for database deserialization.

Bear with me, I really haven't used JS tooling in a long time. Trying to understand how a JS app is shipped.

1. `glean` is a dependency (as per package.json) and all its files are somewhere in `node_modules`
1. `glean_parser` is invoked by the user at some point, generating a `metrics.js` (or similar). That file has all the `import { UrlMetric } from "./types/url.js";`, etc. lines in there
1. webpack/rollup/whatever-bundler goes through all referenced files to bundle them up. (Assuming no text metric) because `./types/text.js` is not `import`ed, webpack/* will not include it at all. It finally generates one big `bundle.js` with all the other code.
1. `bundle.js` is what's loaded in the final application/website/webext

Is my understanding correct?
If so, then [removing these lines are what is the "important" part](https://github.com/mozilla/glean.js/pull/1065/files#diff-6319d5f7d9e2933d6509715cd99505c483d5761c4bebf097ba55ec5d22bdbc50L10-L20) I guess.

> > (Am I correct with that example?)
> > My first reaction is: I think this is fine then. Version A+1 wouldn't record a new value. In the current model we would get a value from a previous run, but no new ones anyway.
> > Losing that one-time stale value doesn't seem to dramatic.
> 
> Yeah I tend to agree with this, but I am working on assumptions here. I was wondering if you have seen such a pattern in other places.

Not really. glean-core always has all metrics (and we can't ever remove metric types really because our database serialization depends on it).
So for once-valid data in the database we can always turn that into JSON.
However for data we can't deserialize we [simply skip it](https://github.com/mozilla/glean/blob/3e839a38f11525064e432b39000804bed19e6985/glean-core/src/database/mod.rs#L417-L420).
We don't even delete such data.
There should be no code path where we write data that we can't later decode, but these are files on disk afterall, so some external process could mess with the data. We have no numbers on whether that ever happens really.

brizental:
> Bear with me, I really haven't used JS tooling in a long time. Trying to understand how a JS app is shipped.
> 
>     1. `glean` is a dependency (as per package.json) and all its files are somewhere in `node_modules`
> 
>     2. `glean_parser` is invoked by the user at some point, generating a `metrics.js` (or similar). That file has all the `import { UrlMetric } from "./types/url.js";`, etc. lines in there
> 
>     3. webpack/rollup/whatever-bundler goes through all referenced files to bundle them up. (Assuming no text metric) because `./types/text.js` is not `import`ed, webpack/* will not include it at all. It finally generates one big `bundle.js` with all the other code.
> 
>     4. `bundle.js` is what's loaded in the final application/website/webext
> 
> 
> Is my understanding correct? If so, then [removing these lines are what is the "important" part](https://github.com/mozilla/glean.js/pull/1065/files#diff-6319d5f7d9e2933d6509715cd99505c483d5761c4bebf097ba55ec5d22bdbc50L10-L20) I guess.

No bearing required, your understanding is perfect. Yes, that is the crucial part of this PR.

> > > (Am I correct with that example?)
> > > My first reaction is: I think this is fine then. Version A+1 wouldn't record a new value. In the current model we would get a value from a previous run, but no new ones anyway.
> > > Losing that one-time stale value doesn't seem to dramatic.
> > 
> > 
> > Yeah I tend to agree with this, but I am working on assumptions here. I was wondering if you have seen such a pattern in other places.
> 
> Not really. glean-core always has all metrics (and we can't ever remove metric types really because our database serialization depends on it). So for once-valid data in the database we can always turn that into JSON. However for data we can't deserialize we [simply skip it](https://github.com/mozilla/glean/blob/3e839a38f11525064e432b39000804bed19e6985/glean-core/src/database/mod.rs#L417-L420). We don't even delete such data. 

Because it is just so damn easy for people to mess with the database data in the browser I think I'll stick to deleting the wrongly formatted data for now. There is [a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1682273) to think about this behaviour and it's problems more deeply in the future. (Fun fact: this is the oldest bug in Glean.js' backlog!).

It might be worth instrumenting it in the future though. I'll keep that in mind.

> There should be no code path where we write data that we can't later decode, but these are files on disk afterall, so some external process could mess with the data. We have no numbers on whether that ever happens really.

In this PR I essentially add a possible code path for users to write data the database can't later decode. But from what I understand we are alright with that.

---

Anyways, thanks a lot for the thorough discussion. I'll clean this up and promote it to a non-draft PR. I'll make sure to add some docs here or in a later PR capturing this discussion.

glean/src/core/context.ts
brizental:
```suggestion
  // If a metric is not in this map it cannot be deserialized from the database.
```

badboy:
> Because it is just so damn easy for people to mess with the database data in the browser I think I'll stick to deleting the wrongly formatted data for now. There is [a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1682273) to think about this behaviour and it's problems more deeply in the future. (Fun fact: this is the oldest bug in Glean.js' backlog!).

yeah, I think the fact that glean core is not deleting data is a bug.

> Anyways, thanks a lot for the thorough discussion. I'll clean this up and promote it to a non-draft PR. I'll make sure to add some docs here or in a later PR capturing this discussion.

Agreed. Docs can help here, but I also think it's not even that surprising to users: you delete your metrics -> data's gone.

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 [Full changelog](https://github.com/mozilla/glean.js/compare/v0.30.0...main)
 
+* [#1065](https://github.com/mozilla/glean.js/pull/1065): Delete minimal amount of data when invalid data is found while collecting ping.
+  * Previous behaviour was to delete the whole ping when invalid data was found on the database,
+  new behaviour only deletes the actually invalid data and leave the rest of the ping intact.
+* [#1065](https://github.com/mozilla/glean.js/pull/1065): Only import metric types into the library when they are used either by the user or Glean itself.
+  * Previously the code required to deserialize metric data from the database was always imported by the library even if the metric type was never used by the client. This effort will decrete the size of the Glean.js bundles that don't import all the metric types.
```
badboy, line 9:
```suggestion
  * Previously the code required to deserialize metric data from the database was always imported by the library even if the metric type was never used by the client. This effort will decrease the size of the Glean.js bundles that don't import all the metric types.
```

badboy (approved):

PR #1095, opened 2022-01-17
Bug 1698091, 1698089 - Document how to create plugins and events

brizental:
[Rendered](https://github.com/brizental/glean.js/blob/1698089-plugins-docs/docs/implementation/plugins-events.md)

### DIFF
--- a/docs/implementation/plugins-events.md
+++ b/docs/implementation/plugins-events.md
@@ -1,0 +1,49 @@
+# Plugins
+
+Plugins provide a way for Glean.js to support especialized use cases without bloating
+the size of the library. Each plugin can be attached to a Glean core event to augment
+or modify its behaviour.
+
+A Glean plugin is a class that extends the [`Plugin`](https://mozilla.github.io/glean.js/classes/plugins.default.html) class. Each plugin instance exposes an
+`action` API, to be triggered by the event. The signature of the `action` API will depend on
+the event the plugin is attached to.
+
+## Core events
+
+Events denote a specific moment in code that may have a plugin attached to it.
+
+A Glean event is an instance of the [`CoreEvent`](https://mozilla.github.io/glean.js/classes/core_events.CoreEvent.html) class. These instances expose a `trigger` API,
+for triggering the action of a plugin that may be attached to it. Each event defines a different
+signature for the `trigger` API.
+
+## When should we implement a feature as a plugin?
+
+When we are requested a feature that would make more sense as a plugin than as a core feature
+we create a new plugin and event for it. However, when is that?
+
+Plugins should always be the choice when we are implementing a feature that has a very especialized
+use case and would be left unused by most users.
+
+The [`PingEncryptionPlugin`](https://mozilla.github.io/glean.js/classes/plugins_encryption.default.html)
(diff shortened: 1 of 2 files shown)


docs/implementation/plugins-events.md
```diff
@@ -0,0 +1,49 @@
+# Plugins
+
+Plugins provide a way for Glean.js to support especialized use cases without bloating
```
chutten, line 3:
```suggestion
Plugins provide a way for the Glean JS SDK to support specialized use cases without bloating
```

docs/implementation/plugins-events.md
```diff
@@ -0,0 +1,49 @@
...
+Events denote a specific moment in code that may have a plugin attached to it.
+
+A Glean event is an instance of the [`CoreEvent`](https://mozilla.github.io/glean.js/classes/core_events.CoreEvent.html) class. These instances expose a `trigger` API,
+for triggering the action of a plugin that may be attached to it. Each event defines a different
+signature for the `trigger` API.
+
+## When should we implement a feature as a plugin?
+
+When we are requested a feature that would make more sense as a plugin than as a core feature
+we create a new plugin and event for it. However, when is that?
```
chutten, line 22:
```suggestion
When a requested feature would make more sense as a plugin than as a core feature
we should create a new plugin and event for it. However, when is that?
```

docs/implementation/plugins-events.md
```diff
@@ -0,0 +1,49 @@
...
+
+Plugins should always be the choice when we are implementing a feature that has a very especialized
```
chutten, line 24:
```suggestion
Plugins should always be the choice when we are implementing a feature that has a very specialized
```

docs/implementation/plugins-events.md
```diff
@@ -0,0 +1,49 @@
...
+use case and would be left unused by most users.
+
+The [`PingEncryptionPlugin`](https://mozilla.github.io/glean.js/classes/plugins_encryption.default.html)
+is a good example of that. That plugin encrypts all outgoing pings when enabled. It is a fairly large
+plugin due to the encryption dependencies it includes in the bundle and it is only used by a small
+amount of clients. It makes sense to make it a stand-alone feature.
```
chutten, line 30: Seems as though there are two criteria we are considering, then:
1) Left unused by most users of the Glean JS SDK
2) Is big (code size)

chutten (approved):
Minor spelling/grammar stuff, mostly. No meaning is lost if it's landed as-is.

I did have to read some of the other linked resources to understand quite exactly how it works (Event gathers Context which, when triggered, passes to the Action on the Plugin. The Plugin then (synchronously) returns a Result which is given back to the Event which can provide it to wherever it is in the SDK). But that's not a bad thing.

docs/implementation/plugins-events.md
brizental: Actually, I think left unused is the main reason to follow. Even if the feature is not big, a lot of small things can sum up to a bug thing... Maybe this part can cause confusion about that. I can clarify.

brizental: Added a small note :)

PR #1097, opened 2022-01-17
RFC: Bug 1741583: Initial implementation of Experiment plugin

ChinYing-Li:
Currently the one newly added unit test does not pass, but I'd like to seek people's advice to see if I am on the right track. Any suggestion is appreciated (Just saw the plugin docs!)

Signed-off-by: ChinYing-Li <chinying.li@mail.utoronto.ca>

### DIFF
--- a/glean/src/core/events/index.ts
+++ b/glean/src/core/events/index.ts
@@ -5,8 +5,9 @@
 import type Plugin from "../../plugins/index.js";
 import log, { LoggingLevel } from "../log.js";
 
-import type { PingPayload } from "../pings/ping_payload.js";
+import type {PingInfo, PingPayload} from "../pings/ping_payload.js";
 import type { JSONObject } from "../utils.js";
+import { ExperimentData } from "../metrics/types/experiment";
 
 const LOG_TAG = "core.Events";
 
@@ -76,13 +77,19 @@ export class CoreEvent<
  */
 const CoreEvents: {
   afterPingCollection: CoreEvent<[PingPayload], Promise<JSONObject>>,
+  afterPingInfoCollection: CoreEvent<[PingInfo, ExperimentData?], Promise<JSONObject>>,
   [unused: string]: CoreEvent
 } = {
   // Event that is triggered immediatelly after a ping is collect and before it is recorded.
   //
   //  - Context: The `PingPayload` of the recently collected ping.
   //  - Result: The modified payload as a JSON object.
-  afterPingCollection: new CoreEvent<[PingPayload], Promise<JSONObject>>("afterPingCollection")
+  afterPingCollection: new CoreEvent<[PingPayload], Promise<JSONObject>>("afterPingCollection"),
+  // Event that is triggered immediatelly after a ping_info is collect and before it is recorded.
+  //
+  //  - Context: The `PingPayload` of the recently collected ping.
+  //  - Result: The modified payload as a JSON object.
+  afterPingInfoCollection: new CoreEvent<[PingInfo, ExperimentData?], Promise<JSONObject>>("afterPingInfoCollection")
 };
 
 export default CoreEvents;
--- a/glean/src/core/pings/maker.ts
+++ b/glean/src/core/pings/maker.ts
@@ -12,6 +12,7 @@ import CoreEvents from "../events/index.js";
 import { Lifetime } from "../metrics/lifetime.js";
 import { Context } from "../context.js";
 import log, { LoggingLevel } from "../log.js";
+import { ExperimentData } from "../metrics/types/experiment.js";
 
 const LOG_TAG = "core.Pings.Maker";
 
@@ -109,9 +110,10 @@ export async function getStartEndTimes(ping: CommonPingData): Promise<{ startTim
  *
  * @param ping The ping to build the `ping_info` section for.
  * @param reason The reason for submitting this ping.
+ * @param experiment_data
  * @returns The final `ping_info` section in its payload format.
  */
-export async function buildPingInfoSection(ping: CommonPingData, reason?: string): Promise<PingInfo> {
+export async function buildPingInfoSection(ping: CommonPingData, reason?: string, experiment_data?: ExperimentData): Promise<PingInfo> {
   const seq = await getSequenceNumber(ping);
   const { startTime, endTime } = await getStartEndTimes(ping);
 
@@ -125,6 +127,7 @@ export async function buildPingInfoSection(ping: CommonPingData, reason?: string
     pingInfo.reason = reason;
   }
 
+  await CoreEvents.afterPingInfoCollection.trigger(pingInfo, experiment_data);
   return pingInfo;
 }
 
--- a/glean/src/plugins/experiment.ts
+++ b/glean/src/plugins/experiment.ts
@@ -1,0 +1,98 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Plugin from "./index.js";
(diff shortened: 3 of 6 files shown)


glean/src/plugins/experiment.ts
```diff
@@ -0,0 +1,98 @@
...
+    if (extra != null && Object.keys(extra).length > MAX_EXPERIMENTS_EXTRAS_SIZE) {
+      const msg = `Extra hash map length ${Object.keys(extra).length} exceeds maximum of ${MAX_EXPERIMENTS_EXTRAS_SIZE}`;
+      await Context.errorManager.record(metric, ErrorType.InvalidValue, msg);
+
+      const temp: {[key: string]: string} = {};
+      for (const [key, value] of Object.entries(extra).slice(0, MAX_EXPERIMENTS_EXTRAS_SIZE)) {
+        const truncated_key = (key.length <= MAX_EXPERIMENTS_IDS_LEN) ? key : await truncateStringAtBoundaryWithError(metric, key, MAX_EXPERIMENTS_IDS_LEN);
+        temp[truncated_key] = (value.length <= MAX_EXPERIMENT_VALUE_LEN) ? value : await truncateStringAtBoundaryWithError(metric, value, MAX_EXPERIMENT_VALUE_LEN);
+      }
+      truncated_extras = temp;
```
brizental, line 61: Let's move this whole logic do a `setActive` API on the experiment metric itself. This way it is jsut like in glean-core.

glean/src/plugins/experiment.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    try {
+      await Context.metricsDatabase.clear(Lifetime.Application, metric.name);
+    } catch (e) {
+      log(
+        LOG_TAG,
+        `Failed to set experiment as inactive ${e}`,
+        LoggingLevel.Error
+      )
+    }
```
brizental, line 81: Same, let's move this logic to a `setInactive` API on the metric type itself to keep things just like in glean-core. And then have this function just call that one. These two functions should be analogous to these ones https://github.com/mozilla/glean/blob/main/glean-core/src/lib.rs#L752-L780

glean/src/plugins/experiment.ts
```diff
@@ -0,0 +1,98 @@
...
+    const metric = new ExperimentMetricType({
+      category: "",
+      name: `${ping_info.seq}#experiment_data`,  // Not sure if this is a good idea.
+      sendInPings: [PING_INFO_STORAGE],
+      lifetime: Lifetime.User,
+      disabled: false
+    });
+    await this.setExperimentActive(metric, experiment_data.branch, experiment_data.extra);
+    ping_info.experiment_data = experiment_data;
+    return ping_info;
```
brizental, line 94: The action should not activate the experiment. Only the calling of the `setExperimentActive` or `setExperimentInactive` APIs. On this action we should attempt to get the experimenta dat and modify the ping info with it. If there is no experiment data, we simply leave the PingInfo un-modified.

glean/src/core/events/index.ts
```diff
@@ -76,13 +77,19 @@ export class CoreEvent<
...
   //
   //  - Context: The `PingPayload` of the recently collected ping.
   //  - Result: The modified payload as a JSON object.
-  afterPingCollection: new CoreEvent<[PingPayload], Promise<JSONObject>>("afterPingCollection")
+  afterPingCollection: new CoreEvent<[PingPayload], Promise<JSONObject>>("afterPingCollection"),
+  // Event that is triggered immediatelly after a ping_info is collect and before it is recorded.
+  //
+  //  - Context: The `PingPayload` of the recently collected ping.
+  //  - Result: The modified payload as a JSON object.
+  afterPingInfoCollection: new CoreEvent<[PingInfo, ExperimentData?], Promise<JSONObject>>("afterPingInfoCollection")
```
brizental, line 92: The context on this event should not know anything about the experiment data itself. We can change this to

```suggestion
  afterPingInfoCollection: new CoreEvent<[PingInfo], Promise<JSONObject>>("afterPingInfoCollection")
```

Then the plugin will take that ping_info and modify it as needed. Does it make sense?

glean/src/core/pings/maker.ts
```diff
@@ -125,6 +127,7 @@ export async function buildPingInfoSection(ping: CommonPingData, reason?: string
     pingInfo.reason = reason;
   }
 
+  await CoreEvents.afterPingInfoCollection.trigger(pingInfo, experiment_data);
```
brizental, line 130: Let's not modify the object and instead return a new one with the applied modifications. We can do exactly like we did on the `afterPingCollection` event: https://github.com/mozilla/glean.js/blob/main/glean/src/core/pings/maker.ts#L253-L269. If there is no modified payload that would mean there is no attached plugin to the event and we can just use the original one, also like we do on the `afterPingCollection` event: https://github.com/mozilla/glean.js/blob/main/glean/src/core/pings/maker.ts#L274

brizental (commented):
Great start! Left some comments on the approach. The main problem I would say is attempting to pass the experiment data from the `buildPingInfo` function to the event. The event itself should not know about that, it should pass the `PingInfo` to the plugin and let it do whatever it wants with it. Let me know if the comments make sense, happy to clarify anything.

I skipped over commenting on any nits and minor things for later when the implementation is complete. Anyways, again, great work. Thanks for taking on this one!

ChinYing-Li:
I am trying to correctly implement the constructor function of `ExperimentMetricType`, and would like to call `truncateStringAtBoundaryWithError` in its constructor. However, this is not possible as `truncateStringAtBoundaryWithError` takes `MetricType` as parameter. Is it possible to refactor `truncateStringAtBoundaryWithError` (or even `Context.errorManager.record` such that they don't need to be called with an existing `MetricType`? example: [the glean library](https://github.com/mozilla/glean/blob/8a3ac6c8e8211a9d3f5c623725b83ebe649fa618/glean-core/src/util.rs#L137)

brizental:
Hey @ChinYing-Li on the example you linked there, the function does take the `CommonMetricData` which is enough on the glean-core to use for recording to a metrics type, so in essence it also takes the metric.

If you are in the constructor of the `ExperimentMetricType`, why can't you pass the metric type to that function as well?

ChinYing-Li:
The `metric` parameter we want to use in the constructor of `ExperimentMetricType` should be the `ExperimentMetricType` that's currently being constructed i.e. `this`: this allows us to correctly associated the error with the instance of `MetricType`. I don't think this is possible, though.

brizental:
> The `metric` parameter we want to use in the constructor of `ExperimentMetricType` should be the `ExperimentMetricType` that's currently being constructed i.e. `this`: this allows us to correctly associated the error with the instance of `MetricType`. I don't think this is possible, though.

Why is it not possible? From what I understand it should be.

One thing I noticed is that `truncateStringAtBoundaryWithError` returns a promise. That would be problematic because you can't make constructors async functions. Is that maybe the error you are getting? Maybe we could consider having an `initialize` function for this metric? Not sure how to go about using that, but it's a possible pattern to deal with this problem. Another possibility is to create a `recordAsync` function for recording errors, that uses the dispatched version of the `add` method... Hm, would need to think about the consequences of doing that, however off the top of my mind it should be ok. 

Let me know if this is helpful, happy to clarify things further.

brizental:
Hey @ChinYing-Li are you planning to finish this patch? No worries if not, I am just asking because in that case I will finish this one up.

ChinYing-Li:
@brizental Apologies for the delay, I will continue to work on this patch!

ChinYing-Li:
@brizental Sorry that I haven't found time to complete this PR; please feel free to take over!

brizental:
> @brizental Sorry that I haven't found time to complete this PR; please feel free to take over!

Thanks for letting us know @ChinYing-Li :D

PR #1130, opened 2022-01-26
Bug 1751402 - Guarantee getMonotonicNow doesn't return a negative value

brizental:
We were previously extracting the TIME_ORIGIN from Date.now()
in case performance.now was not available in the environment.

That could create negative timestamps, because Date.now is not
guaranteed to be monotonically increasing. (Yeah... it is
contradictiory to have a function called getMonotonicNow, but not
guarantee monotonically increasing timestamps for some platforms.
At least there is a TODO comment to address the incoherence at
some point).

Subtracting the TIME_ORIGIN was essentially pointless and I don't really
know why that was the case before. That function is used by the Timestamp
metric type and for event timestamps which are both about elapsed time,
so the absoulte timestamp should not matter.

EXTRA: Address stray TODO comment on TimespanMetricType test suite.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * [#1130](https://github.com/mozilla/glean.js/pull/1130): BUGFIX: Guarantee event timestamps
 cannot be negative numbers.
-  * Timestamps were observed to be negative in a few occurrences, for platforms that do not provide the `performance.now` API, namely QML, and in which we fallback to the `Date.now` API.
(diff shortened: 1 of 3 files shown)


chutten:
Oops, sorry I didn't get to this yesterday as I planned.

Is it the storage that's responsible for setting the first event's timestamp to 0, then?

brizental:
> Oops, sorry I didn't get to this yesterday as I planned.

No worries, jan-erik asked about it that is why I marked him as reviewer :D

> Is it the storage that's responsible for setting the first event's timestamp to 0, then?

Yeah it is: https://github.com/mozilla/glean.js/blob/main/glean/src/core/metrics/events_database/index.ts#L286

PR #1135, opened 2022-01-28
Bug 1741586 - Create browser compat tests for web platform and run on CI

brizental:
EXTRA: Get better coverage for Chromium based browsers. See f0520aa

### Pull Request checklist ###

- [x] **Quality**: Make sure this PR builds and runs cleanly.
  - Inside the `glean/` folder, run:
    - `npm run test` Runs _all_ tests
    - `npm run lint` Runs _all_ linters
- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [x] **Changelog**: This PR includes a changelog entry to `CHANGELOG.md` or an explanation of why it does not need one
- [x] **Documentation**: This PR includes documentation changes, an explanation of why it does not need that or a follow-up bug has been filed to do that work

PR #1136, opened 2022-01-28
Bug 1741586 - Create browser compat tests for web platform and run on CI

brizental:
EXTRA: Get better coverage for Chromium based browsers. See f0520aa

Reopening this PR after accidentally deleting the branch... See #1135

The tests work, but are not enabled for PRs. See: https://app.circleci.com/pipelines/github/mozilla/glean.js/3290/workflows/627f2268-b957-4a3c-9fbd-684fda6b1d04/jobs/13551


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 cannot be negative numbers.
   * Timestamps were observed to be negative in a few occurrences, for platforms that do not provide the `performance.now` API, namely QML, and in which we fallback to the `Date.now` API.
   * If event timestamps are negative pings are rejected by the pipeline.
+* [#1132](https://github.com/mozilla/glean.js/pull/1132): Retry ping request on network error with `keepalive: false`. This is sometimes the issue Chrome browsers below v81.
```
chutten, line 9: Grammar "This is sometimes the issue Chrome browser below v81" ... maybe...

```suggestion
* [#1132](https://github.com/mozilla/glean.js/pull/1132): Retry ping request on network error with `keepalive: false`. This is sometimes an issue on Chrome browsers below v81.
```

benchmarks/README.md
```diff
@@ -16,4 +22,45 @@ To run the report and post the results to a GitHub pull request:
...
 
-_Note: This requires a `GITHUB_TOKEN` and the CircleCI environment to know which PR to post to._
+> **Note**: This script requires the `GITHUB_TOKEN` environment variable
+> and the CircleCI environment to know which PR to post to.
+> It is not really meant to be run outside of CircleCI.
+
+### Docs
+
+Builds the Glean.js library for all the targeted platforms and document
+the size of the library and of each extra metrics and plugins.
```
chutten, line 32:
```suggestion
the size of the library and of each extra metric and plugin.
```

benchmarks/README.md
```diff
@@ -16,4 +22,45 @@ To run the report and post the results to a GitHub pull request:
...
+npm run size:docs:dry
+```
+
+To run the script and update the documentation.
+
+```
+npm run size:docs
+```
+
+## Browser compatibility
```
chutten, line 46: Huh. So browser compatibility is considered a benchmark? I always thought benchmarks were perf things.

benchmarks/compat/index.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=edge">
```
chutten, line 5: Thanks to this I've now looked at the flow chart on https://docs.microsoft.com/en-us/openspecs/ie_standards/ms-iedoco/380e2488-f5eb-4457-a07a-0cb1b6e4b4b5

benchmarks/compat/tests/browserstack.test.js
```diff
@@ -0,0 +1,95 @@
...
+import webdriver from "selenium-webdriver";
+
+import { runWebTest, fastSelenium, App } from "./utils.js"
+
+fastSelenium();
+
+const app = new App();
+app.start();
+
+// All of the minimum supported versions were found throw manual testing.
```
chutten, line 15:
```suggestion
// All of the minimum supported versions were found through manual testing.
```

benchmarks/compat/tests/browserstack.test.js
```diff
@@ -0,0 +1,95 @@
...
+const BROWSERS = [
+  { name: "firefox", minVersion: "60" },
+  { name: "safari", minVersion: "11.1" },
+  { name: "chrome", minVersion: "66" },
+  // Oldest version of edge available in BrowserStack.
+  // We are not sure it does not work in even older versions.
+  { name: "edge", minVersion: "80" },
+];
+
+// Start BrowserStack.Local so that the remote machines can access  localhost.
```
chutten, line 25:
```suggestion
// Start BrowserStack.Local so that the remote machines can access localhost.
```

benchmarks/compat/tests/browserstack.test.js
```diff
@@ -0,0 +1,95 @@
...
+const bs = new browserstack.Local();
+await new Promise(resolve => {
```
chutten, line 27: Gotta love that top-level `await`

docs/reference/compat.md
```diff
@@ -0,0 +1,32 @@
+# Compatibility
+
+## Websites
+
+The Glean JavaScript SDK for websites supports all the latest versions of major browsers.
+
+The following list, details the minimum supported version for each major browser.
+All of the listed versions are explicitly tested.
+
+| Browser | Minimum Supported Version |
```
chutten, line 10: Do we also want to document what the latest tested version is?

benchmarks/compat/index.js
```diff
@@ -0,0 +1,35 @@
...
+
+Glean.setSourceTags(["automation"]);
+Glean.initialize("glean-compat-benchmark", true);
+
+metrics.pageLoaded.set();
+benchmark.submit();
+
+// !BIG HACK!
+//
+// Ovewrite the console.info function in order to know when (and if) the benchmark ping was sent.
```
chutten, line 19: Hm. It is indeed a big hack, but one that'll be fragile in the correct way (if we change our messages at all, this test will fail, unless we add the phrase "successfully sent 200." to the end of a log message of a failure state).

It may also break if the pipeline is having problems and returns us a `5xx`. This might be a good thing. If we want to include that in the test's scope we should document that even intermittent failures should be examined for their response codes. (Hopefully it'll be easy to get to the Glean SDK's logs to get this information. )

The only other way I can think of to still include a ping upload (not just submission) in the smoke test would be to set a localhost ping server and then pass a message (or remote-control the browser and run JS in the page context) to add the success text to `#msg`. Is there a reason to not do it this way? (And then maybe we can convince BrowserStack to fail the test if it tries to make non-local connections, like m-c's CI does).

chutten (approved):
Looks like it does what it intends to, and I have only smallish questions and suggestions. Could land as-is without causing problems.

benchmarks/README.md
brizental: Well, to be honest I don't know. Wikipedia agrees with you... https://en.wikipedia.org/wiki/Benchmark_(computing). I am happy to change the name of this folder to something else. Any suggestions?

benchmarks/compat/index.js
brizental: Yeah, create a ping server just for this test case would be the non-hacky solution... But also the more involved one. This time I chose the lazy path, heh. 

In case there is a pipeline bug we can access the console or networks logs on the testing browser to figure out what whas the error code / what happened. In BrowserStack we have access to that and even though I intercept the console.info, they still end up in the console, see https://github.com/mozilla/glean.js/pull/1136/files#diff-64051b0b88d31fa148b9b4bdeaec65fed18bebb91779f7720aab69d9f5c752f4R30.

docs/reference/compat.md
brizental: That changes too frequently for us to keep this doc up to date... I can make it clear that we actually test in the "latest" version as well as in the minimum supported version.

benchmarks/README.md
brizental: Renamed it to "reports". Think that is more accurate I guess.

brizental: Ok, changed the name of this to "automation"... Kinda weird, but I guess it works. Let me know what you think.

PR #1223, opened 2022-02-28
Bug 1742642 - Add command to retrieve glean parser version on the Glean CLI

brizental:
Yes, users could potentially use "glean --version"
but that would install glean_parser. This command allows
to retrieve the glean_parser version without installing it.

See: https://bugzilla.mozilla.org/show_bug.cgi?id=1742642#c4

Example: 

```
% ./node_modules/.bin/glean --glean-parser-version
5.0.1
```

  - https://bugzilla.mozilla.org/show_bug.cgi?id=1741829

### DIFF
--- a/glean/src/cli.ts
+++ b/glean/src/cli.ts
@@ -230,6 +230,11 @@ function stopSpinner(spinner: NodeJS.Timeout) {
  * @param args the arguments passed to this process.
  */
 async function run(args: string[]) {
+  if (args.includes("--glean-parser-version")) {
+    console.log(GLEAN_PARSER_VERSION);
+    process.exit(1);
+  }
+
   try {
     await setup();
   } catch (err) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,8 @@
 
 [Full changelog](https://github.com/mozilla/glean.js/compare/v0.31.0...main)
 
-* [#1220](https://github.com/mozilla/glean.js/pull/1220): Refactor virtual environment behavior to support virtual environments that aren't in the project root. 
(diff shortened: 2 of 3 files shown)


brizental:
@stevejalim Does this solution address what you mention in https://bugzilla.mozilla.org/show_bug.cgi?id=1742642#c4?

brizental:
wrt [2fc6ead](https://github.com/mozilla/glean.js/pull/1223/commits/2fc6ead767afd0f0a38f2cf8f62bec556fd6421a) I updated all the releases in the first page of our releases notes to include the glean_parser version (I thought that should be enough). Docs + example should be enough for us to remember to add it on every new release.

glean/src/cli.ts
```diff
@@ -236,6 +236,11 @@ function stopSpinner(spinner: NodeJS.Timeout) {
  * @param args the arguments passed to this process.
  */
 async function run(args: string[]) {
+  if (args.includes("--glean-parser-version")) {
+    console.log(GLEAN_PARSER_VERSION);
+    process.exit(1);
```
badboy, line 241: Error exit? Should probably be 0.

```suggestion
    process.exit(0);
```

badboy (approved):

glean/src/cli.ts
brizental: Oops.

stevejalim:
@brizental Hey, nice! That looks cool. I like the idea of using `glean` itself to tell us what version of the parser is in use, rather than having to extract it from a specific file via a separate script.  

It looks like this should also be possible:

`pip install -U glean_parser==$(./node_modules/.bin/glean --glean-parser-version)`

stevejalim (approved):
Nice one 🎸

PR #1271, opened 2022-03-22
Bug 1760556 - Update internal ping validation function to account for collectionDate e prop

brizental:
This was causing a bug in the following situation:

1. User submits a ping that contains custom headers
2. Restarts the app, the app scans the pending pings directory
3. Ping is thrown away because validation function only accepts pings with 2 to 3 properties and since v0.19.0 the internal pings have 3 to 4 properties due to the addition of the `collectionDate` property.

### DIFF
--- a/glean/src/core/pings/database.ts
+++ b/glean/src/core/pings/database.ts
@@ -49,7 +49,7 @@ export interface PingInternalRepresentation extends JSONObject {
  *          stating whether `v` is in the correct ping internal representation.
  */
 export function isValidPingInternalRepresentation(v: unknown): v is PingInternalRepresentation {
-  if (isObject(v) && [3, 4].includes(Object.keys(v).length)) {
+  if (isObject(v)) {
     const hasValidCollectionDate = "collectionDate" in v && isString(v.collectionDate) && isNumber(new Date(v.collectionDate).getTime());
     const hasValidPath = "path" in v && isString(v.path);
     const hasValidPayload = "payload" in v && isJSONValue(v.payload) && isObject(v.payload);
--- a/glean/src/core/pings/database.ts
+++ b/glean/src/core/pings/database.ts
@@ -4,7 +4,7 @@
 
 import type Store from "../storage/index.js";
 import type { JSONObject} from "../utils.js";
-import { isObject, isJSONValue, isString } from "../utils.js";
+import { isNumber, isObject, isJSONValue, isString } from "../utils.js";
 import log, { LoggingLevel } from "../log.js";
 import { DELETION_REQUEST_PING_NAME } from "../constants.js";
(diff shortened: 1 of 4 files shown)


glean/src/core/pings/database.ts
```diff
@@ -47,11 +49,12 @@ export interface PingInternalRepresentation extends JSONObject {
  *          stating whether `v` is in the correct ping internal representation.
  */
 export function isValidPingInternalRepresentation(v: unknown): v is PingInternalRepresentation {
-  if (isObject(v) && (Object.keys(v).length === 2 || Object.keys(v).length === 3)) {
+  if (isObject(v) && [3, 4].includes(Object.keys(v).length)) {
```
Dexterp37, line 52: Do we need to strictly check the number of properties here? This seems like the kind of thing that could bite us again

glean/src/core/pings/database.ts
```diff
@@ -161,7 +164,11 @@ class PingsDatabase {
         if (isValidPingInternalRepresentation(ping)) {
           finalPings[identifier] = ping;
         } else {
-          log(LOG_TAG, "Unexpected data found in pings database. Deleting.", LoggingLevel.Warn);
+          log(
+            LOG_TAG,
+            `Unexpected data found in pings database: ${JSON.stringify(ping, null, 2)}. Deleting.`,
+            LoggingLevel.Warn
+          );
```
Dexterp37, line 171: I wish we had an health ping. This is the kind of error that can get unnoticed unless people look at the logs, and exactly the reason why we had an health ping. /rant

Dexterp37:
@brizental did this cause data loss?

brizental:
> @brizental did this cause data loss?

Potentially, but only on the Glean Dictionary. For VPN, regrets-reporter and Rally it should not have caused any data loss. I was just checking this and none of these apps enable any of the (debugging) features that would result in custom headers in production, so this bug would not be hit.

Even on the Glean Dictionary this bug should not have been hit, were it not for some other bug in there which I currently investigating over on https://github.com/mozilla/glean-dictionary/issues/1188.

glean/src/core/pings/database.ts
brizental: Yeap, I thought about that while implementing the fix... not sure why I did not go with it. The main drawback is that it could result in larger than necessary data in storage, but we can worry about it if it ever happens (which IMO is unlikely). I'll change this.

glean/src/core/pings/database.ts
brizental: Should I file a bug? 👀

Dexterp37: There's one alreayd, don't worry :)

Dexterp37 (approved):

PR #1452, opened 2022-08-24
Bug 1723090 - Delete trailing glean.restarted events

rosahbruno:
### Description ###
Add functionality to delete all trailing `glean.restarted` events. Now before returning the formatted events for the ping, we pop items from the array until we hit one that isn't a `restarted` event. If the last item is not a `glean.restarted` event, then it will not do anything.

Glean Book Documentation PR - https://github.com/mozilla/glean/pull/2172


rosahbruno:
I want to get a second opinion on what should be updated in the docs. If you look under `glean.restarted` there is verbiage that I am not entirely sure is accurate anymore. https://github.com/mozilla/glean.js/blob/9186c5aa38ab6a35b184137e6c8f8985b771e5d8/docs/reference/metrics.md#all-pings

We now strip the first restarted event from the events array as well as all trailing events. I may lack enough context to know for sure, but I would think that there is a possibility now where that event doesn't get passed at all.

I am going to do some testing with the sample projects to try and confirm as well.

glean/tests/unit/core/metrics/events_database.spec.ts
```diff
@@ -400,8 +398,13 @@ describe("EventsDatabase", function() {
...
       assert.strictEqual("test", (snapshot[i * 2] as JSONObject)["category"]);
       assert.strictEqual(`stichting_test_${i}`, (snapshot[i * 2] as JSONObject)["name"]);
-      assert.strictEqual("glean", (snapshot[(i * 2) + 1] as JSONObject)["category"]);
-      assert.strictEqual("restarted", (snapshot[(i * 2) + 1] as JSONObject)["name"]);
+
+      // We no longer keep trailing restarted events, so in this scenario, we need to ignore
+      // the final element of the snapshot since it previously had a restarted event.
+      if (snapshot[(i * 2) + 1]) {
+        assert.strictEqual("glean", (snapshot[(i * 2) + 1] as JSONObject)["category"]);
+        assert.strictEqual("restarted", (snapshot[(i * 2) + 1] as JSONObject)["name"]);
```
badboy, line 406: might want to pull out the `snapshot[(i * 2) + 1]` into a variable with a descriptive name

glean/tests/unit/core/metrics/events_database.spec.ts
```diff
@@ -455,8 +458,13 @@ describe("EventsDatabase", function() {
...
       assert.strictEqual("test", (snapshot[i * 2] as JSONObject)["category"]);
       assert.strictEqual(`time_travel_${i}`, (snapshot[i * 2] as JSONObject)["name"]);
-      assert.strictEqual("glean", (snapshot[(i * 2) + 1] as JSONObject)["category"]);
-      assert.strictEqual("restarted", (snapshot[(i * 2) + 1] as JSONObject)["name"]);
+
+      // We no longer keep trailing restarted events, so in this scenario, we need to ignore
+      // the final element of the snapshot since it previously had a restarted event.
+      if (snapshot[(i * 2) + 1]) {
+        assert.strictEqual("glean", (snapshot[(i * 2) + 1] as JSONObject)["category"]);
+        assert.strictEqual("restarted", (snapshot[(i * 2) + 1] as JSONObject)["name"]);
```
badboy, line 466: same here

glean/tests/unit/core/metrics/events_database.spec.ts
```diff
@@ -509,8 +517,13 @@ describe("EventsDatabase", function() {
...
```
badboy, line 525: and here

glean/tests/unit/core/pings/maker.spec.ts
```diff
@@ -273,4 +273,47 @@ describe("PingMaker", function() {
...
+      } else {
+        // Record a restarted event (easiest way)
+        //
+        // Un-initialize and re-initialize manually instead of using testResetGlean
+        // in order to have control over the startTime at initialization.
+        await testUninitializeGlean(false);
+        // Move the clock backwards by one hour.
+        //
+        // This will generate incoherent timestamps in events at collection time
+        // and record an `InvalidValue` error for the `glean.restarted` event.
```
badboy, line 307: Why is this something we want?

glean/src/core/metrics/events_database/index.ts
```diff
@@ -402,6 +406,27 @@ class EventsDatabase {
   async clearAll(): Promise<void> {
     await this.eventsStore.delete([]);
   }
+
+  private isRestartedEvent(event: RecordedEvent): boolean {
+    return !!event?.get()?.extra?.[GLEAN_REFERENCE_TIME_EXTRA_KEY];
```
badboy, line 411: I would have expected that we check category and name, but I guess this extra value always exists and the key is unique, and so this might be sufficient enough.

glean/tests/unit/core/metrics/events_database.spec.ts
rosahbruno: Good call, will fix shortly

glean/tests/unit/core/pings/maker.spec.ts
rosahbruno: This was just a consistent way that I found to generate a `glean.restarted` event. I will take another pass and see how I can generate one without creating an error

glean/src/core/metrics/events_database/index.ts
rosahbruno: Originally I was checking the category name and it was working the same, then I was looking through the rest of the code, and everywhere else we check for a `glean.restarted` event we check it that way. If you think that the category and name are more reliable data points to check, I can update it to check those instead.

badboy: Nah, if we're consistent with other checks that seems fine for me. Maybe add a comment though.

glean/tests/unit/core/pings/maker.spec.ts
rosahbruno: @badboy 
I found a more straightforward way in my latest commit for triggering those restarts. The only thing that I am unsure of is whether or not I need to worry about these messages when running the tests.

I played around a bit and it looks like I can resolve most of these by jumping ahead in time so events aren't given the same timestamp, but not sure if it is worth the extra messing around when everything appears to be working the same way after my updates.

```
(Glean.core.metrics.Event) glean.restarted: Invalid time offset between application sessions found for ping "aPing". Ignoring.
```

glean/src/core/metrics/events_database/index.ts
rosahbruno: Fixed in 7c06436a736cf9354ef9ec45845434f304a40dc0

glean/tests/unit/core/metrics/events_database.spec.ts
rosahbruno: Fixed in 54fbc1e7e825bce6db734f45132fee63f42b82d2

glean/tests/unit/core/metrics/events_database.spec.ts
rosahbruno: Fixed in 54fbc1e7e825bce6db734f45132fee63f42b82d2

glean/tests/unit/core/metrics/events_database.spec.ts
rosahbruno: Fixed in 54fbc1e7e825bce6db734f45132fee63f42b82d2

badboy:
Just noticed that I wrote some comments, but didn't submit them -_-

badboy (commented):
Code looks good, but PR is still in draft mode. If you're ready turn it into a full PR please.

glean/tests/unit/core/pings/maker.spec.ts
badboy: So other tests use a different way to fake a restart: https://github.com/mozilla/glean.js/blob/ac4c56a31121ec74fd63db8b40854c6e3381b205/glean/tests/unit/core/metrics/events_database.spec.ts#L336-L340

Should this just be used here too?

badboy:
> I want to get a second opinion on what should be updated in the docs. If you look under `glean.restarted` there is verbiage that I am not entirely sure is accurate anymore. [`9186c5a`/docs/reference/metrics.md#all-pings](https://github.com/mozilla/glean.js/blob/9186c5aa38ab6a35b184137e6c8f8985b771e5d8/docs/reference/metrics.md#all-pings)
> 
> We now strip the first restarted event from the events array as well as all trailing events. I may lack enough context to know for sure, but I would think that there is a possibility now where that event doesn't get passed at all.
> 
> I am going to do some testing with the sample projects to try and confirm as well.

I think that wording is still correct, though is leaving out a bit of context.
In _general_ `glean.restarted` can end up in any ping with events, without the user needing to define anything extra.
Whether a specific ping with events has that particular event depends on whether the application was actually restarted.
Even if we strip the first and all trailing restarted events it might still occur in the middle.

glean/tests/unit/core/pings/maker.spec.ts
rosahbruno: When I was looking around I saw a few different ways to trigger a restart. I can give this one a shot and hopefully, it doesn't generate that error.

rosahbruno:
> Code looks good, but PR is still in draft mode. If you're ready turn it into a full PR please.

Was just waiting on some feedback for the docs part since I knew that was going to still need updating before publishing for real. I will update the docs with that additional context, look at triggering restarts without an error, then publish this.

@badboy thanks for the multiple reviews!

glean/tests/unit/core/pings/maker.spec.ts
badboy: Might be worth filing a bug to eventually come up with a single way and use that consistently through out tests.
For now I'm fine with either way.

badboy:
> > Code looks good, but PR is still in draft mode. If you're ready turn it into a full PR please.
> 
> Was just waiting on some feedback for the docs part since I knew that was going to still need updating before publishing for real. I will update the docs with that additional context, look at triggering restarts without an error, then publish this.

yeah, I'm not sure how to concisely add that for the single metric.
We explain `glean.restarted` a tiny bit more [in the Glean SDKs book](https://mozilla.github.io/glean/book/user/pings/custom.html#the-gleanrestarted-event).

rosahbruno:
> > > Code looks good, but PR is still in draft mode. If you're ready turn it into a full PR please.
> > 
> > 
> > Was just waiting on some feedback for the docs part since I knew that was going to still need updating before publishing for real. I will update the docs with that additional context, look at triggering restarts without an error, then publish this.
> 
> yeah, I'm not sure how to concisely add that for the single metric. We explain `glean.restarted` a tiny bit more [in the Glean SDKs book](https://mozilla.github.io/glean/book/user/pings/custom.html#the-gleanrestarted-event).

@badboy 
I added a small clarification in the `metrics.md` docs just to note that those leading and trailing restarted events will be omitted. I can add a note in the Glean SDKs book too, since there is already that JS specific section.

a8f8ed4cdc572c7c58b6c580e38cc47b95ca70d8

glean/src/core/testing/index.ts
```diff
@@ -29,3 +29,5 @@ export async function testResetGlean(
   await testUninitializeGlean(clearStores);
   await testInitializeGlean(applicationId, uploadEnabled, config);
 }
+
+export { testRestartGlean };
```
rosahbruno, line 33: Clarification for this

A more common pattern in JS/TS is to bubble all of your folder/module exports up to an `index.ts` file, "barrel file", and export everything from there.

That way when you are accessing different functions/variables/constants you only have to drill down to the top level of the folder when importing. So instead of pulling 3 things from index, 4 things from utils, etc, you re-export everything you need from the index then you can just import it all via `import {} from '../core/testing`.

This also helps to keep things more organized and modular. You start to see very quickly what may be tightly coupled to a module that doesn't need to be.

glean/src/core/testing/index.ts
```diff
@@ -29,3 +29,6 @@ export async function testResetGlean(
   await testUninitializeGlean(clearStores);
   await testInitializeGlean(applicationId, uploadEnabled, config);
 }
+
+export * from "./utils.js";
```
rosahbruno, line 33: Clarification for this

A more common pattern in JS/TS is to bubble all of your folder/module exports up to an index.ts file, "barrel file", and export everything from there.

That way when you are accessing different functions/variables/constants you only have to drill down to the top level of the folder when importing. So instead of pulling 3 things from index, 4 things from utils, etc, you re-export everything you need from the index then you can just import it all via import {} from '../core/testing.

This also helps to keep things more organized and modular. You start to see very quickly what may be tightly coupled to a module that doesn't need to be.

glean/tests/unit/core/pings/maker.spec.ts
rosahbruno: @badboy 
This should be resolved in the latest commit. I create a helper function inside of `core/testing` that lets you pass a specified time interval you want to jump (forwards or backwards) - the default is set at a minute. I will log a bug for cleaning up all the places in our tests where we create `restarted` events to use this helper instead of forcing it through other means.

b5fe0a1914e38d16b67feaa39676e39b9e5c4f09

rosahbruno:
PR is now ready for review. I've made all PR changes as well as create a documentation update PR for the Glean Book.

badboy (approved):

PR #1475, opened 2022-09-08
Bug 1716954 - Implement the Timing Distribution Metric

rosahbruno:
This PR implements the timing distribution metric into Glean.js. The majority of the logic is a direct port from the rust implementation, but there were a lot of things that had to be adjusted or implemented a little differently because it was in Typescript.

I added a substantial amount of unit tests, some ported almost exactly from Glean and some custom ones. I also **tested the new metric in all of the sample projects** and ran into no immediately obvious issues.

Notable things that were changed
- Ported over all the functionality for timing distribution metric
- Ported over all the functionality for histograms (Functional, Linear, Exponential)
- Ported over all the tests for both the timing distribution and the histograms
- Configured project for being able to debug tests with `ts-mocha`
- Updated the `getting_started` guide
- Added a new counter to `context` to get as close to an atomic counter as we can
- Added a custom `binarySearch` function as an alternative to the one built-in to Rust. It is specific for bucketing and will return a value that lets you determine which bucket it is closest to if there is no exact match.
- Bucketing interface that each of the histogram types implements
- Added the new metric to the QT entry file
- Updated global schema test to include the new timing distribution metric
- Updated `metrics.yaml` to include data for new timing distribution metric

Created a PR for updating the Glean Book docs - https://github.com/mozilla/glean/pull/2187


rosahbruno:
@badboy 
When I was updating the Glean docs I noticed that the `measure` API is not provided in all languages, is this something that we want to implement for Glean.js?

glean/src/core/metrics/types/timing_distribution.ts
```diff
@@ -0,0 +1,588 @@
...
+//    always a 1:1 match with the actual durations, so trying to store our keys or
+//    any other histogram data just causes more headaches.
+//
+type TimingDistributionInternalRepresentation = number[];
+
+export type TimingDistributionPayloadRepresentation = {
+  // A map containing the bucket index mapped to the accumulated count.
+  //
+  // This can contain buckets with a count of `0`.
+  values: Record<TimerId, number>;
```
badboy, line 61: Why is this `TimerId` here? Arent' those just the bucket indices?

glean/src/core/metrics/types/timing_distribution.ts
```diff
@@ -0,0 +1,588 @@
...
+    const nonNumericKey = (v as number[]).find((key) => isNaN(+key));
+    if (nonNumericKey) {
+      return {
+        type: MetricValidation.Error,
+        errorType: ErrorType.InvalidValue,
+        errorMessage: `Expected all durations to be numbers, got ${nonNumericKey}`,
+      };
+    }
+
+    const negativeDuration = (v as number[]).find((key) => !isNaN(+key) && +key < 0);
```
badboy, line 142: Why `+key`? Is this just to ensure this is a number? (I don't do nearly enough js/ts, so I will occasionally ask basic questions like this to understand)

glean/src/core/metrics/types/timing_distribution.ts
```diff
@@ -0,0 +1,588 @@
...
+          .catch();
+        duration = maxSampleTime;
+      } else {
+        // do nothing, we have a valid duration
+      }
+
+      if (!this.shouldRecord(Context.uploadEnabled)) {
+        return;
+      }
+
```
badboy, line 258: We already do this check at the beginning. I think we don't need it here anymore
(I saw we do the same on glean-core, but I don't see why. I think this is a left-over from some earlier refactoring).
```suggestion
```

glean/tests/unit/core/metrics/timing_distribution.spec.ts
```diff
@@ -0,0 +1,619 @@
...
+
+    const id = 100;
+    const duration = 60;
+
+    // Flip around the timestamps, this should result in a negative value which should be
+    // discarded.
+    metric.setStart(id, duration);
+    metric.setStopAndAccumulate(id, 0);
+
+    assert.equal(await metric.testGetValue("aPing"), undefined);
```
badboy, line 162: Above should record an error, right? Let's check for that.

```suggestion
    assert.equal(await metric.testGetValue("aPing"), undefined);
    assert.equal(1, await metric.testGetNumRecordedErrors(ErrorType.InvalidValue));
```

glean/tests/unit/core/metrics/timing_distribution.spec.ts
```diff
@@ -0,0 +1,619 @@
...
+    // Check that we got the right sum and number of samples.
+    assert.equal(snapshot?.sum, 6);
+
+    // We should get a sample in each of the first 3 buckets.
+    assert.equal(1, snapshot?.values[1]);
+    assert.equal(1, snapshot?.values[2]);
+    assert.equal(1, snapshot?.values[3]);
+
+    // 1 error should be reported.
+    assert.equal(0, await metric.testGetNumRecordedErrors(ErrorType.InvalidValue));
```
badboy, line 222: 1 error, right?

```suggestion
    assert.equal(1, await metric.testGetNumRecordedErrors(ErrorType.InvalidValue));
```

glean/tests/unit/core/metrics/timing_distribution.spec.ts
```diff
@@ -0,0 +1,619 @@
...
+    const maxSampleTime = 1000 * 1000 * 1000 * 60 * 10;
+
+    metric.accumulateRawSamplesNanos([0, 1, maxSampleTime + 1]);
+
+    const snapshot = await metric.testGetValue("aPing");
+
+    // Check that we got the right sum and number of samples.
+    assert.equal(snapshot?.sum, 2 + maxSampleTime);
+
+    // We should get a sample in 3 buckets.
```
badboy, line 353: 2 buckets, just 2 values in one of them.

glean/tests/unit/core/metrics/timing_distribution.spec.ts
```diff
@@ -0,0 +1,619 @@
...
+      },
+      TimeUnit.Nanosecond
+    );
+
+    const id = metric.start();
+    metric.stopAndAccumulate(id);
+
+    const testValue = await metric.testGetValue("aPing");
+    assert.strictEqual(Object.keys(testValue?.values || []).length, 1);
+    assert.ok(testValue?.sum || 0 > 0, "The sum of the distribution should be greater than 0.");
```
badboy, line 442: Huh, I read this as `(testValue?.sum) || (0 > 0)` and was seriously confused why one would write that :D 
No change necessary though.

glean/tests/unit/core/metrics/timing_distribution.spec.ts
```diff
@@ -0,0 +1,619 @@
...
+
+    await testRestartGlean();
+
+    const testValue2 = await metric.testGetValue("aPing");
+    assert.strictEqual(Object.keys(testValue2?.values || []).length, 1);
+    assert.ok(testValue2?.sum || 0 > 0, "The sum of the distribution should be greater than 0.");
+  });
+
+  it("unique IDs are persisted through restarts", async function () {
```
badboy, line 451: This is actually interesting as this is different from how it behaves on glean-core.
We don't persist started timers there, so started timers are lost after restarts.

I'm not opposed to just leaving this as is here, but it should be documented.

badboy: Can you point me to where we persist the global timerid counter?

badboy (changes requested):
Good work here!
Got a couple of nits and a few questions we should resolve before landing this.

glean/src/core/metrics/types/timing_distribution.ts
rosahbruno: Yep, +key is a just quick conversion to a number. This conversion is left over from the original implementation where I was storing the map rather than an array of numbers. So good call out, this extra casting shouldn't be needed anymore, I will remove.

glean/tests/unit/core/metrics/timing_distribution.spec.ts
rosahbruno: Haha yeah, the optionals are annoying, so unless I add the fallback it will treat it as possibly undefined and yell at me, so it looks a bit odd. Let me see if there is a way to make these look a little nicer

glean/tests/unit/core/metrics/timing_distribution.spec.ts
rosahbruno: This was something I should've gotten some more clarification on I think. In the Rust implementation instead of using just an integer, it uses `AtomicUsize` instead. Since JS doesn't really have a similar implementation, I figured this would be the closest to that.

If you don't think we need to be this worried about assigning timer IDs, then that is fine. My worry was just that once we start a timer and if something happens or there is a restart, we are always going to grab the first ID again and we could end up overwriting previously saved info.

BUT, I could be making incorrect assumptions here or thinking about it wrong too.

glean/src/core/metrics/types/timing_distribution.ts
rosahbruno: The `TimerId` type is nothing more than a copy of the `number` type, it just exists for better readablity. So in reality, that statement is the same as `values: Record<number, number`.

But looking at it now, I don't really think it does us any good to have that alias, its obvious its an ID because it is the first argument for the `Record` and likely just makes things more confusing

glean/tests/unit/core/metrics/timing_distribution.spec.ts
badboy: But in real-world use the only way Glean "restarts" is that someone reloads the webpage (assuming the glean.js-on-web usecase).
Any in-memory state is lost then, including `_timingDistributionId` in the context, right?
If so, I am confused why the test works. Maybe the "restart" machinery doesn't clear out that context correctly?

Are not-yet-stopped timers persisted? (that is: is their `timer id -> start time` mapping storred in the client?)

glean/src/core/metrics/types/timing_distribution.ts
badboy: But this is `bucket index -> count`, not `timer id -> start time` (or similar), right?

rosahbruno: Yeah, that's true, that's a good point. I am going to just remove `TimerId` alias altogether anyways, it is being used incorrectly and doesn't really help make anything easier to understand

rosahbruno: Removed in 8270f955be485847c86e9ed914a8cbe03611ba1c

glean/tests/unit/core/metrics/timing_distribution.spec.ts
rosahbruno: Yep, copied this comment directly from Rust, so we can likely update that too at some point
https://github.com/mozilla/glean/blob/main/glean-core/tests/timing_distribution.rs#L392

rosahbruno: Fixed 95632814c29afb4ff917264c89510f362e9450b5

glean/tests/unit/core/metrics/timing_distribution.spec.ts
rosahbruno: Yes, tested this locally and it works as expected.

glean/tests/unit/core/metrics/timing_distribution.spec.ts
rosahbruno: Yes! Good call, this was due to me incorrectly setting the error type. I need to make another change for this to work, so I will just push a commit with this and the other change.

rosahbruno: Fixed https://github.com/mozilla/glean.js/pull/1475/commits/ecbeedf5aa971056f95b85f7d61b8e7b5ac76763

glean/tests/unit/core/metrics/timing_distribution.spec.ts
rosahbruno: This would actually be better as just `assert.ok(!!testValue?.sum)` since truthy'ing the value makes sure that any integer is greater than 0. I will update to use that instead of the more confusing option

rosahbruno: I updated these checks to use the truthy check now instead. I also updated all the places where we manually checked the length of our values map (`Object.keys(values).length`) to use the `count` property for our `DistributionData`, which accomplishes the same thing - just much more easily.

glean/tests/unit/core/metrics/timing_distribution.spec.ts
rosahbruno: I don't want to answer without looking into it, let me take a look and see what I can find

rosahbruno: So I took a look, and what we talked about earlier was correct - they aren't persisting, the restart just isn't clearing out the important context.

Since this is not actually "persisting" anything and we don't need to worry about that, I refactored the implementation to just use a class variable to store the ID. Now instead of reading and writing from the global context, we just update the `timerId` in the class itself.

I re-ran all the tests and everything is still working as expected without the more global ID. So as long as you don't see any issues with just managing that ID in the class itself, this should be better.

glean/tests/unit/core/metrics/timing_distribution.spec.ts
```diff
@@ -0,0 +1,563 @@
...
+      TimeUnit.Nanosecond
+    );
+
+    const id = metric.start();
+    metric.cancel(id);
+
+    assert.strictEqual(await metric.testGetValue("aPing"), undefined);
+  });
+
+  it("attempting to start/accumulate when glean upload is disable is a no-op", async function () {
```
badboy, line 397:
```suggestion
  it("attempting to start/accumulate when glean upload is disabled is a no-op", async function () {
```

glean/tests/unit/core/metrics/timing_distribution.spec.ts
```diff
@@ -0,0 +1,563 @@
...
+        sendInPings: ["aPing"],
+      },
+      TimeUnit.Nanosecond
+    );
+
+    const id1 = metric.start();
+    Glean.setUploadEnabled(false);
+    metric.stopAndAccumulate(id1);
+
+    const id2 = metric.start();
```
badboy, line 433: Can you add a comment that this is the one that records? `start` always works, we only check upload-enabled on `stop`

badboy (approved):
few last nits, but then it's ready to go

glean/tests/unit/core/metrics/timing_distribution.spec.ts
rosahbruno: Good call, added a clarifying comment
https://github.com/mozilla/glean.js/pull/1475/commits/be8bbf182444056a6d7ccb9bc864d8bcb8099d8f

PR #1514, opened 2022-09-20
Bug 1716952 - Implement the Memory Distribution Metric

rosahbruno:
Implements the memory distribution metric

Notable changes
- Ported over all the functionality for the memory distribution metric
- Ported over all the tests for memory distributions
- Created distributions helper file containing types/functions for our distributions
- Created `MemoryUnit` enum and conversion helper
- Wrote additional tests for memory distributions
- Added the new metric to the QT entry file
- Updated global schema test to include the new memory distribution metric
- Updated `metrics.yaml` to include data for new memory distribution metric
- fixed some typos

PR for updating the Glean Book docs - https://github.com/mozilla/glean/pull/2203


rosahbruno:
@badboy 
I am testing this out in the web project and had a question about the persistence of data. If I call `accumulate`, refresh the page, call `accumulate` again, then submit the ping I get two values in my histogram. Is this expected? Or once the user refreshes their page should the old data be cleared?

I know we chatted a bit about how this should work with timing distributions, so wanted to get your thoughts.

badboy:
> @badboy I am testing this out in the web project and had a question about the persistence of data. If I call `accumulate`, refresh the page, call `accumulate` again, then submit the ping I get two values in my histogram. Is this expected? Or once the user refreshes their page should the old data be cleared?
> 
> I know we chatted a bit about how this should work with timing distributions, so wanted to get your thoughts.

2 values is correct (if this is a ping lifetime).
`accumulate` here is the recording function, so we persist that.

(for timing distributions `stop` is the recording function. Only calling `start`, but never `stop` means we're not recording anything, that's why we "loose" data there. that's no issue for memory distributions, because we only have a single function to call)

rosahbruno:
@badboy 
Ok great, thanks so much!

glean/src/core/metrics/time_unit.ts
```diff
@@ -23,4 +23,30 @@ enum TimeUnit {
   Day = "day",
 }
 
+/**
+ * Converts a number from any `TimeUnit` to nanoseconds.
+ *
+ * @param duration Difference between start and stop time stamps
+ * @param timeUnit Time unit for the duration
```
Dexterp37, line 30:
```suggestion
 * @param duration Difference between start and stop time stamps.
 * @param timeUnit Time unit for the duration.
```

glean/tests/integration/schema/metrics.yaml
```diff
@@ -230,3 +230,17 @@ for_testing:
     send_in_pings:
       - testing
     time_unit: nanosecond
+  memory_distribution:
+    type: memory_distribution
+    description: |
+      Sample memory distribution metric.
+    bugs:
+      - https://bugzilla.mozilla.org/000000
```
Dexterp37, line 238:
```suggestion
      - https://bugzilla.mozilla.org/1716952
```

glean/tests/integration/schema/metrics.yaml
```diff
@@ -230,3 +230,17 @@ for_testing:
...
+    data_reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=000000#c3
```
Dexterp37, line 240:
```suggestion
      - https://bugzilla.mozilla.org/show_bug.cgi?id=1716952
```

glean/src/core/metrics/distributions.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+/**
+ * Create a snapshot of the histogram with a time unit.
+ *
+ * Utility function for testing.
+ *
+ * **Caution**
+ * This cannot use `Histogram.snapshot_values` and needs to use the more
+ * specialized snapshot function.
```
Dexterp37, line 28: Is this still true? Do we even have this function in Glean.js? I believe this description was copy-pasted from the Rust implementation but does not really apply here.

glean/src/core/metrics/distributions.ts
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * @param hist Histogram to get the snapshot of
+ * @returns Snapshot of the current histogram
```
Dexterp37, line 31:
```suggestion
 * @param hist Histogram to get the snapshot of.
 * @returns Snapshot of the current histogram.
```

glean/src/core/metrics/types/memory_distribution.ts
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    const negativeDuration = (v as number[]).find((key) => key < 0);
+    if (negativeDuration) {
+      return {
+        type: MetricValidation.Error,
+        errorType: ErrorType.InvalidValue,
+        errorMessage: `Expected all durations to be greater than 0, got ${negativeDuration}`,
+      };
+    }
```
Dexterp37, line 73: This is not a timing distribution and so we're not expecting durations at all. Does this apply here?

Dexterp37: Ok, after reading the rest of the code, this might still apply. Let's not call it "durations" though, maybe just "negativeSample", as it is otherwise confusing.

Dexterp37 (changes requested):
This looks great overall. There's a few minor things to fix, but it's very close to be merge-able.

glean/src/core/metrics/types/memory_distribution.ts
rosahbruno: Good callout, yes I will rename this.

rosahbruno: Fixed in [`74c8631`](https://github.com/mozilla/glean.js/pull/1514/commits/74c8631ddee81306e4315f017a2b0bda8aebb81b)

glean/src/core/metrics/distributions.ts
rosahbruno: Looking at the rust implementation, this `snapshot` function is utilized by both timing and memory distributions. When working with a custom distribution, we don't use this function, we just use `snapshotValues` instead.

So, yes I think we still need this function because the case with custom distributions should still be valid. This is a good callout though because I need to update the custom distribution PR to behave the same way.

I will fix the comment since it references the rust function name rather than the JS one.

Here is what I found in terms of where this is true
- Timing Distribution - https://github.com/mozilla/glean/blob/main/glean-core/src/metrics/timing_distribution.rs#L73
- Memory Distribution - https://github.com/mozilla/glean/blob/main/glean-core/src/metrics/memory_distribution.rs#L42
- Custom Distribution - https://github.com/mozilla/glean/blob/main/glean-core/src/metrics/custom_distribution.rs#L32

rosahbruno: Removed in https://github.com/mozilla/glean.js/pull/1514/commits/d0d2d45bf9b0402d4aa39fe54c48417fce4c7fec

rosahbruno:
@Dexterp37 
Thanks for all the feedback! This is ready for another review.

Dexterp37 (approved):

PR #1695, opened 2023-06-07
Glean.js web rework

rosahbruno:
I am going to go back through and break this up into more digestible commits, but for the sake of getting something up, here is everything.

**Summary**
The gist of the changeset involves making a new structure for files (async/sync/shared) for any internal processes that now need to run synchronously for web. The reason that everything was async before was because the storage (IndexedDB) is all asynchronous. The web implementation now uses LocalStorage instead, which is synchronous.

95% of this code isn't actually changed, it is just moved around. The only real things that include new code are
- Shared interfaces for async/sync internal processes
- Storage layer for web
- Uploader for web

**Callouts**
- The `async/shared/sync` structure causes issues for individual metrics because of how generated files work.
- There are still TODOs in here for things that I need to work out


glean/src/platform/browser/web/storage.ts
```diff
@@ -2,187 +2,84 @@
...
+import log, { LoggingLevel } from "../../../core/log.js";
+import {
+  deleteKeyFromNestedObject,
+  getValueFromNestedObject,
+  updateNestedObject
+} from "../../../core/storage/utils.js";
 
-const DATABASE_NAME = "Glean";
+const LOG_TAG = "platform.web.Storage";
 const STORE_NAME = "Main";
```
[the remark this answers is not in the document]
rosahbruno: Done, changed this to `Gleanjs`.

https://github.com/mozilla/glean.js/pull/1695/files#diff-16e18cf640e972ca20c053b77babbc469491815ff5de3e9803e8b672cbaf368bR17

glean/src/core/upload/uploader.ts
```diff
@@ -38,7 +38,7 @@ export class UploadResult {
 }
 
 /**
- * Uploader abstract class, actualy uploading logic varies per platform.
+ * Uploader abstract class, actually uploading logic varies per platform.
```
travis79, line 41: "actually" ->
"actual"

glean/src/core/upload/worker/sync.ts
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  /**
+   * Attempts to upload a ping.
+   *
+   * @param ping The ping object containing headers and payload.
+   * @returns The status number of the response or `undefined` if unable to attempt upload.
+   */
+  private async attemptPingUpload(ping: QueuedPing): Promise<UploadResult> {
+    // TODO
```
travis79, line 89: TODO without a bug to help remind us of the todo

glean/src/core/upload/worker/sync.ts
```diff
@@ -0,0 +1,148 @@
...
+   */
+  work(
+    getUploadTask: () => UploadTask,
+    processUploadResponse: (ping: QueuedPing, result: UploadResult) => void
+  ): void {
+    try {
+      const task = getUploadTask();
+      switch (task.type) {
+      case UploadTaskTypes.Upload:
+        // TODO
```
travis79, line 127: Another TODO without a bug

glean/src/plugins/encryption.ts
```diff
@@ -18,6 +18,8 @@ import CoreEvents from "../core/events/index.js";
 const JWE_ALGORITHM = "ECDH-ES";
 const JWE_CONTENT_ENCODING = "A256GCM";
 
+// TODO: Determine if we still need this plugin.
```
travis79, line 21: TODO without a bug

travis79 (commented):
Whew! That's was a serious PR stack! Good job on keeping things reviewable, most of my comments are just cosmetic things.

glean/src/core/metrics/database/sync.ts
```diff
@@ -0,0 +1,267 @@
...
+
+  /// PRIVATE ///
+  /**
+   * Returns the store instance for a given lifetime.
+   *
+   * @param lifetime The lifetime related to the store we want to retrieve.
+   * @returns The store related to the given lifetime.
+   * @throws If the provided lifetime does not have a related store.
+   */
+  private _chooseStore(lifetime: Lifetime): SynchronousStore {
```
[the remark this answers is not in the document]
rosahbruno: This one likely is just around from copying. I will update the rest of the functions to match. Since Typescript has the ability to mark functions as private, the underscores likely aren't necessary.

glean/src/platform/browser/web/storage.ts
```diff
@@ -2,187 +2,82 @@
...
-          const updatedObj = updateNestedObject(value, index, transformFn);
-          const request = store.put(updatedObj, this.rootKey);
-          request.onsuccess = resolve;
-          request.onerror = () => reject(request.error);
-        });
-      },
-      DatabaseTransactionModes.ReadWrite,
-    );
+  update(index: StorageIndex, transformFn: (v?: JSONValue) => JSONValue): void {
+    // TODO: Fix issue where stores are written to nested objects.
```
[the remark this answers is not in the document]
rosahbruno: Nope, this is still something I am going to fix in this PR, just haven't fixed it yet.

glean/src/plugins/encryption.ts
rosahbruno: These TODOs will still be addressed in this PR, so hopefully won't need their own bug.

glean/src/core/context.ts
```diff
@@ -2,18 +2,31 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import type MetricsDatabase from "./metrics/database.js";
-import type EventsDatabase from "./metrics/events_database/index.js";
-import type PingsDatabase from "./pings/database.js";
-import type ErrorManager from "./error/index.js";
-import type Platform from "../platform/index.js";
+import type MetricsDatabase from "./metrics/database/async.js";
+import type MetricsDatabaseSync from "./metrics/database/sync.js";
```
badboy, line 6: When I looked earlier at the size report I saw a bit of an increase.
I guess the fact that we unconditionally include both code paths here might be one of the contributing factors. This should be looked at in https://bugzilla.mozilla.org/show_bug.cgi?id=1839937

badboy (approved):

glean/src/core/context.ts
rosahbruno: Yeah, unfortunately in the context we need to include both no matter what since we are determining there which implementation we are going to use. But yes, I will spend some time on this in the new bug and see if there is a way we can avoid any major size changes.

PR #1728, opened 2023-07-11
Bug 1839934 - Migrate client_id & first_run_date

rosahbruno:
Migrate the two user lifetime internal metrics (client_id & first_run_date) whenever a user runs the app for the first time. Once the migration has happened, we set a flag in LocalStorage to tell Glean not to try again.

### DIFF
--- a/glean/src/core/internal_metrics/sync.ts
+++ b/glean/src/core/internal_metrics/sync.ts
@@ -243,6 +243,13 @@ export class CoreMetricsSync {
           } else {
             this.initializeFirstRunDate();
           }
+
+          // Get previous sequence number.
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+          const seq = req.result?.["glean_ping_info"]?.["counter"]?.["submission#sequence"] as number;
+          if (!!seq) {
+            localStorage.setItem("MIGRATION_SEQ", seq.toString());
+          }
         };
 
         req.onerror = () => {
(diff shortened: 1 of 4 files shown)


glean/src/core/internal_metrics/sync.ts
```diff
@@ -187,4 +192,67 @@ export class CoreMetricsSync {
...
+   */
+  private initializeUserLifetimeMetrics(): void {
+    this.initializeClientId();
+    this.initializeFirstRunDate();
+  }
+
+  /**
+   * Migrates the core metrics from the old IDB storage to LocalStorage on first launch.
+   */
+  private migrateCoreMetricsFromIdb(): void {
```
chutten, line 207: I'd be tempted to prefix this with `maybe` since the precondition isn't "We know there are metrics to migrate".

Also, softeng time: I think the more proper pattern might be to return boolean from this about whether migration was successful and let the caller determine whether/when/how to initialize the user lifetime metrics. It's a little weird seeing calls to `initializeUserLifetimeMetrics` inside here _and_ outside in `initialize`.

chutten (approved):
Could land as-is, but I have some potential tweaks that may improve things a smidge.

glean/src/core/internal_metrics/sync.ts
rosahbruno: Yes good callout, I will adjust the comments there.

I don't think we can return anything from this function because of the listeners. Since they are going to run whenever the event happens, we likely can't check that everything actually completed or not as a whole and then return a value.

I think for now, until I figure out a nicer way to deal with this and the listeners, I am just going to add some comments explaining the behavior and calling out why things are structured like they are.

PR #1750, opened 2023-08-08
Update to glean-parser 8.1 and allow any patch version

badboy:
I copied the script from https://github.com/mozilla/glean/blob/3c1c083238876dd6cc4e7577c66b73d309ab83d8/gradle-plugin/src/main/groovy/mozilla/telemetry/glean-gradle-plugin/GleanGradlePlugin.groovy#L61

This now uses `~=` for the glean-parser install, so we don't need to update glean.js for every glean_parser patch update.

### DIFF
--- a/docs/guides/update_glean_parser.md
+++ b/docs/guides/update_glean_parser.md
@@ -3,8 +3,10 @@
 To update the version of glean_parser used by the Glean.js, run the `bin/update-glean-parser-version.sh` script, providing the version as a command line parameter:
 
 ```bash
-bin/update-glean-parser-version.sh 1.28.3
+bin/update-glean-parser-version.sh 1.28
 ```
+
+It's enough to specify `MINOR.MAJOR`.
 
 This will update the version in all the required places and create a commit with the changes. After that, you just need to create a pull request with the changes.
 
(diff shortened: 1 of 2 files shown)


badboy:
Probably needs a fix to `bin/update-glean-parser-version.sh` to only set the `MAJOR.MINOR` version correctly.

What to test:

* Does it pick glean_parser 8.1.1 correctly on a clean source?
* Manually install glean_parser 8.1.0 -- does it work?

rosahbruno:
> Probably needs a fix to `bin/update-glean-parser-version.sh` to only set the `MAJOR.MINOR` version correctly.

It looks like this command only updates two spots
1. `cli.ts` - it will set whatever you pass in. If we run the script with version 8.0, it will set the value of `GLEAN_PARSER_VERSION` to 8.0
2. `qt/requirements.txt` - we are not currently supporting any projects on QT and plan to likely decommission support, so likely not worth extra work there to deal with this

> 
> What to test:
> 
>     * Does it pick glean_parser 8.1.1 correctly on a clean source?
> 
>     * Manually install glean_parser 8.1.0 -- does it work?

I was able to `git clean` my repo and run and it picked up 8.1.1. I then manually pulled in 8.1.0 and it seemed to work correctly too.

rosahbruno (approved):

rosahbruno (changes requested):
Actually, before we merge. I think we should update these docs to specify that running this command should only use `<MAJOR>.<MINOR>` now.

https://github.com/mozilla/glean.js/blob/main/docs/guides/update_glean_parser.md?plain=1#L6

PR #1751, opened 2023-08-10
Bug 1848200 - add migration flag to init

rosahbruno:
Conditionally migrate IDB data only if it already exists. If the project has never used an older version of Glean, then we want to avoid migration because of a potential race condition when submitting pings on first page load.

### DIFF
--- a/glean/src/core/glean/sync.ts
+++ b/glean/src/core/glean/sync.ts
@@ -2,8 +2,10 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import { CLIENT_INFO_STORAGE, KNOWN_CLIENT_ID } from "../constants.js";
 import type { ConfigurationInterface } from "../config.js";
+import type PlatformSync from "../../platform/sync.js";
+
+import { CLIENT_INFO_STORAGE, KNOWN_CLIENT_ID } from "../constants.js";
 import { Configuration } from "../config.js";
 import PingUploadManager from "../upload/manager/sync.js";
 import { isBoolean, isString, sanitizeApplicationId } from "../utils.js";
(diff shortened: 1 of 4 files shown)


glean/src/core/glean/sync.ts
```diff
@@ -42,7 +43,15 @@ namespace Glean {
    */
   function onUploadEnabled(): void {
     Context.uploadEnabled = true;
-    (Context.coreMetrics as CoreMetricsSync).initialize();
```
Dexterp37, line 45: Why don't we keep this here and just add a parameter to this function?

travis79: @Dexterp37 Moving this to the function below more closely follows the pattern and naming used in glean-core, but I don't have any strong opinions and defer to your opinions on this.

rosahbruno: True, I will update

rosahbruno:
> What happens if Local Storage is cleared? will this continue to try and migrate metrics? 

If LocalStorage is cleared, it will try and run just once. Currently, we aren't migrating most metrics - we are just migrating `client_id` and `first_run_date`. So on a second migration the flag will be set and the values will be set to the same thing as before.

> How long does the migration code need to be in tree, have you considered this and planned (and filed a bug) to remove it at some time in the future?

That is something I've been thinking about. I'd like to maybe chat with you during our next 1:1 to get some thoughts here. In the mean time, I will file a bug so that this isn't lost in the shuffle.

Dexterp37:
> > How long does the migration code need to be in tree, have you considered this and planned (and filed a bug) to remove it at some time in the future?
> 
> That is something I've been thinking about. I'd like to maybe chat with you during our next 1:1 to get some thoughts here. In the mean time, I will file a bug so that this isn't lost in the shuffle.

Only as long as MDN hasn't migrated.

glean/src/core/glean/sync.ts
rosahbruno: @Dexterp37 
Do you have issues with leaving this as is based on Travis' comment? It doesn't matter to me either way.

Dexterp37: No, no concern.

PR #1772, opened 2023-09-29
Bugs 1855792, 1855795 - window.Glean updates

rosahbruno:
I tested this by pushing up a branch with the `dist` folder checked in and running this version directly inside of the `mdn` project. The `window is not defined` error goes away when wrapped in the `typeof window !== undefined` check.

### DIFF
--- a/automation/compat/tests/utils.js
+++ b/automation/compat/tests/utils.js
@@ -74,6 +74,30 @@ export async function runWebTest(driver) {
     await driver.get(`http://localhost:${PORT}/`);
     // Give it time to send the ping request.
     const successTextContainer = await driver.findElement(By.id("msg"));
+
+    const areGleanWindowVarsSet = await driver.executeScript(() => {
+      // Verify that all Glean `window` vars are properly set.
+      if (
+        window &&
+        window.Glean &&
+        window.Glean.setDebugViewTag &&
+        window.Glean.setLogPings &&
+        window.Glean.setSourceTags
+      ) {
+        return true;
+      }
+
+      // One of our expected values wasn't set, so the test should fail.
+      return false;
+    }).catch(() => {
+      // In case `window` isn't available or something unexpected happens.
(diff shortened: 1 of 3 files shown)


Dexterp37 (approved):

automation/compat/tests/utils.js
```diff
@@ -74,6 +74,30 @@ export async function runWebTest(driver) {
     await driver.get(`http://localhost:${PORT}/`);
     // Give it time to send the ping request.
     const successTextContainer = await driver.findElement(By.id("msg"));
+
+    const areGleanWindowVarsSet = await driver.executeScript(() => {
+      // Verify that all Glean `window` vars are properly set.
+      if (
+        window &&
```
caugner, line 81: @rosahbruno Might this need to be `typeof window !== "undefined"`?

rosahbruno: These are actually our browserstack tests. So for this scenario, we want our tests to fail if for some reason we can't find the `window` object.

The place where we believe this is coming from is fixed in [this PR](https://github.com/mozilla/glean.js/pull/1788). For whatever reason, our helper function `isUndefined` isn't working right and is actually leading to the error being thrown. As a fix, we just check directly against `window` instead.

I tested this by pulling down MDN and running `yarn dev` with v2.0.4 and reproducing the error. When I retested and did the same thing with the code I pushed to that new branch, I was able to successfully run `yarn dev` and run the app locally.

Hopefully this should resolve the errors you are seeing whenever we cut a new release!

caugner: Awesome, thank you! 🙌

caugner: PS: My gut feeling says that passing an undefined variable to `isUndefined()` just throws an error before you get into the function, because it uses undefined variable whereas an explicit `typeof varname !== "undefined"` never throws.

PR #1808, opened 2023-11-08
Only sync Glean.js

rosahbruno:
This PR updates Glean.js to be completely synchronous with only a `web` implementation.

The 3 other platforms are going to undergo changes
- QT was deprecated in #1810 
- Node is being temporarily removed in #1813 
- Web Extension is being temporarily removed in #1814 

Both the node and web extension platforms are going to be added back in later. They are being removed now so we can focus on web and build a new architecture that allows for platform OR framework specific implementations with their own auto-instrumentation.

Important updates included in this PR
- Removal of `async`/`sync` file structure
- Removal of the `dispatcher` since it is not needed in the sync implementation
- Removal of `plugins` as they are asynchronous and not used in any active projects
- Removal of tests that are async specific and no longer fit in Glean.js


Dexterp37:
@rosahbruno a few observations on this:

* Please remove the "builtin metrics" from this PR. We should tackle this as a follow-up after this refactor.
* Please break out the QT code removal out of this PR, to simplify reviewing (instead of removing the docs from the Glean Book, we should mark that as deprecated, mention the last supported version and file a bug for moving these docs to an "Obsolete" part of the book)

rosahbruno:
> @rosahbruno a few observations on this:
> 
>     * Please remove the "builtin metrics" from this PR. We should tackle this as a follow-up after this refactor.
> 
>     * Please break out the QT code removal out of this PR, to simplify reviewing (instead of removing the docs from the Glean Book, we should mark that as deprecated, mention the last supported version and file a bug for moving these docs to an "Obsolete" part of the book)

Yep, this makes sense. I will break these into separate PRs.

> 
> 
> Note that I see 13 unit tests failing locally, not 2:
> 
>     * Some of them are failing because of `isWindowObjectUnavailable` in internal_metrics.ts/initialize
> 
>     * Some of them are failing because of the newly automatically instrumented page load events (e.g. `"send the 'events' ping when max capacity is hit"`)
> 
> 
> By splitting the PR as advised above we should be more confident in the changes we make.

I will take a look

rosahbruno:
This PR is now rebased and includes the changes from #1810. It also includes and builds off the changes made in #1813 and #1814.

glean/src/core/metrics/database.ts
```diff
@@ -79,9 +175,31 @@ class MetricsDatabaseSync {
...
+   * Gets  and validates the persisted payload of a given metric in a given ping.
+   *
+   * If the persisted value is invalid for the metric we are attempting to retrieve,
+   * the persisted value is deleted and `undefined is returned.
+   *
+   * This behaviour is not consistent with what the Glean SDK does, but this is on purpose.
+   * On the Glean SDK we panic when we can't serialize the persisted value,
+   * that is because this is an extremely unlikely situation for that environment.
+   *
+   * Since Glean.js will run on the browser, it is easy for a consumers / developers
```
travis79, line 188:
```suggestion
   * Since Glean.js will run on the browser, it is easy for consumers / developers
```

glean/src/core/error/error_type.ts
```diff
@@ -1,7 +1,6 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-
```
travis79, line 4: Not sure if deleting this whitespace line was intentional or not?

travis79 (approved):
No blocking comments or arguments from me. That was a lot of `await` and `async` to remove!

glean/src/core/error/error_type.ts
rosahbruno: This was the result of adding some stuff to this file, then removing it because having all the error things in a single file caused a bunch of circular dependencies. I added the space back in!

PR #1810, opened 2023-11-08
Remove QT

Dexterp37:
This removes QT and rolls back some of the changes that were made to the APIs in order to support the QT platform (FYI @brizental as your team was the only consumer of this API).

TODO:

- [x] Add a breaking change to the changelog
- [x] Verify that everything else works in a product
- [x] Check if it's ok to remove the unsafe `generateUUIDv4` fallback that was introduced due to Qt. **(It was not ok, as we might still need this for webextensions that this version still supports)**.


automation/size/docs.js
```diff
@@ -1,127 +1,125 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
```
travis79, line 1: It's weird that it thinks you changed this entire file when it's mostly the same. Did the line endings change or something?

travis79 (commented):
This is awesome! I'm a little concerned about the files that appear to have every line changed, did we change line endings from Linux-y to Windows-y maybe?

travis79 (approved):

rosahbruno (approved):
It looks like we may be able to drop the `sqlite` dev dependency in the `/glean` `package.json` since it was only being used by QT.

Also looks like there may be some KnownOperatingSystems we may be able to drop now, but probably not a big deal if we didn't.
https://github.com/mozilla/glean.js/blob/main/glean/src/core/platform_info/shared.ts#L23-L26

Besides that, everything looks good. I will try and run this in a project and make sure everything is working as expected.

rosahbruno:
@Dexterp37 
I ran this version of Glean with the debug ping viewer via this branch https://github.com/rosahbruno/glean.js/tree/remove-qt-dist and everything worked as expected 👍

Dexterp37:
> It looks like we may be able to drop the `sqlite` dev dependency in the `/glean` `package.json` since it was only being used by QT.

Good call, remove that!

> Also looks like there may be some KnownOperatingSystems we may be able to drop now, but probably not a big deal if we didn't. https://github.com/mozilla/glean.js/blob/main/glean/src/core/platform_info/shared.ts#L23-L26

Yes, saw that and thought to remove it as I wasn't really sure and the risk didn't seem worth it.

> Besides that, everything looks good. I will try and run this in a project and make sure everything is working as expected.

Thanks!

PR #1813, opened 2023-11-10
Remove current node implementation

rosahbruno:
To properly configure Glean.js to be only synchronous, we are going to drop the existing support for node first. The other PR I am working on #1808 is rebuilding the Glean internals. To be able to do this, we first need to remove QT, node, and the webext implementations.

By removing these first, we are going to be able to build new platforms/packages for each in a more reusable way that scales nicer than what we are currently doing.

The reason we are doing this now and then going to put it back in is because to be able to drop the `async`/`sync`/`shared` file structure, we need to remove all the platforms except for web.


Dexterp37 (approved):
I'm ok with moving forward with this, as discussed during our workweek. 

We should make sure that this has no impact on the one consumer we have ([Treeherder](https://github.com/mozilla/treeherder)). @badboy checked and it seems this is just UI.

We should probably merge this, cut a new alpha release and see.

PR #1834, opened 2023-11-21
Bug 1777182 - Add a `sendBeacon` uploader

Dexterp37:
Before merging this PR:

- [x] Remove the changes to the sample app
- [x] Figure out with @scholtzan why these submissions are not going through

The use of that API has some implications to the rest of the ecosystem (ingestion, tooling) so we likely want to have this optionally enabled  for a while. Unfortunately, there's no way to simple test this without using it on a product. So here's what I'd recommend:

1. Review and merge this (not used anywhere!)
2. Cut a new prerelease version
3. Use the new prerelease in the debug view or the glean dictionary, and add URL parameters to conditionally turn this on
4. Validate if this works in the pipeline without changing anything ?

### DIFF
--- a/glean/src/core/upload/ping_request.ts
+++ b/glean/src/core/upload/ping_request.ts
@@ -1,0 +1,51 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { gzipSync, strToU8 } from "fflate";
+import { PingBodyOverflowError } from "./ping_body_overflow_error.js";
+
+/**
+ * Represents a payload to be transmitted by an uploading mechanism.
+ */
+export default class PingRequest<BodyType extends string | Uint8Array> {
+  constructor (
+    readonly identifier: string,
+    readonly headers: Record<string, string>,
+    readonly payload: BodyType,
+    readonly maxBodySize: number
+  ) {}
+
+  asCompressedPayload(): PingRequest<string | Uint8Array> {
+    // If this is already gzipped, do nothing.
+    if (this.headers["Content-Encoding"] == "gzip") {
+      return this;
+    }
+
+    const stringifiedBody = JSON.stringify(this.payload);
+    // We prefer using `strToU8` instead of TextEncoder directly,
+    // because it will polyfill TextEncoder if it's not present in the environment.
+    // Environments that don't provide TextEncoder are IE and most importantly QML.
+    const encodedBody = strToU8(stringifiedBody);
+
+    let finalBody: string | Uint8Array;
+    const headers = this.headers;
+    try {
+      finalBody = gzipSync(encodedBody);
+      headers["Content-Encoding"] = "gzip";
+      headers["Content-Length"] = finalBody.length.toString();
+    } catch {
+      // Fall back to whatever we had.
+      return this;
+    }
+
+    if (finalBody.length > this.maxBodySize) {
+      throw new PingBodyOverflowError(
+        `Body for ping ${this.identifier} exceeds ${this.maxBodySize} bytes. Discarding.`
+      );
+    }
+
+    return new PingRequest<string | Uint8Array>(this.identifier, headers, finalBody, this.maxBodySize);
+  }
+}
+
--- a/glean/src/core/upload/ping_request.ts
+++ b/glean/src/core/upload/ping_request.ts
@@ -22,11 +22,10 @@ export default class PingRequest<BodyType extends string | Uint8Array> {
       return this;
     }
 
-    const stringifiedBody = JSON.stringify(this.payload);
     // We prefer using `strToU8` instead of TextEncoder directly,
     // because it will polyfill TextEncoder if it's not present in the environment.
-    // Environments that don't provide TextEncoder are IE and most importantly QML.
-    const encodedBody = strToU8(stringifiedBody);
+    // For example, IE doesn't provide TextEncoder.
+    const encodedBody = strToU8(this.payload as string);
 
     let finalBody: string | Uint8Array;
     const headers = this.headers;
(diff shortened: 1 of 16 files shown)


Dexterp37:
@badboy @rosahbruno do you have thoughts on the testing plan?

glean/src/core/upload/ping_request.ts
```diff
@@ -0,0 +1,51 @@
...
+  asCompressedPayload(): PingRequest<string | Uint8Array> {
+    // If this is already gzipped, do nothing.
+    if (this.headers["Content-Encoding"] == "gzip") {
+      return this;
+    }
+
+    const stringifiedBody = JSON.stringify(this.payload);
+    // We prefer using `strToU8` instead of TextEncoder directly,
+    // because it will polyfill TextEncoder if it's not present in the environment.
+    // Environments that don't provide TextEncoder are IE and most importantly QML.
```
badboy, line 28:
```suggestion
    // Environments that don't provide TextEncoder are IE and most importantly QML.
```

While you are here, probably worth changing this.
We don't have QML support anymore.

badboy (commented):
testing in production!
Yeah, we don't rely on glean dictionary data for anything, so testing it there gets us the most real-world testing.

scholtzan:
Instead of having `sendBeacon` send an escaped JSON string, it actually shouldn't get escaped. This is working:

```javascript
navigator.sendBeacon("https://incoming.telemetry.mozilla.org/submit/debug-ping-view/events/1/e2588a68-6801-45c0-86a8-03f6784b0168", '{"events":[{"category":"sample","name":"random_event","timestamp":0}],"ping_info":{"seq":4,"start_time":"2023-11-29T18:50+01:00","end_time":"2023-11-29T19:05+01:00","reason":"max_capacity"},"client_info":{"telemetry_sdk_build":"4.0.0-pre.0","client_id":"cae639c1-7462-49e6-a8a7-7eced5656820","first_run_date":"2023-11-27+01:00","os":"Windows","os_version":"Unknown","architecture":"Unknown","locale":"it-IT","app_build":"Unknown","app_display_version":"Unknown"}}');
```

glean/src/entry/web.ts
```diff
@@ -6,4 +6,5 @@ import platform from "../platform/browser/web/index.js";
 import { base } from "./base.js";
 
 export { default as Uploader, UploadResult, UploadResultStatus } from "../core/upload/uploader.js";
+export {default as BrowserSendBeaconUploader} from "../platform/browser/sendbeacon_uploader.js";
```
rosahbruno, line 9: I know we chatted before about the having the two uploaders work together and we would use whatever was available. I just wanted to confirm that was going into a separate PR right?

I'm guessing for this we are going to land and test by just setting the sendbeacon uploader on the Glean Dictionary?

glean/src/platform/browser/sendbeacon_uploader.ts
```diff
@@ -0,0 +1,42 @@
...
+import log, { LoggingLevel } from "../../core/log.js";
+import Uploader from "../../core/upload/uploader.js";
+import { DEFAULT_UPLOAD_TIMEOUT_MS, UploadResultStatus, UploadResult } from "../../core/upload/uploader.js";
+
+const LOG_TAG = "platform.browser.SendBeaconUploader";
+
+class BrowserSendBeaconUploader extends Uploader {
+  timeoutMs: number = DEFAULT_UPLOAD_TIMEOUT_MS;
+
+  // eslint-disable-next-line @typescript-eslint/require-await
```
rosahbruno, line 16: Hmm.. I want to think about this a little bit and if we should really be ignoring async/await stuff. I know we do it a little differently for fetch, I will come back and comment on this part tomorrow.

rosahbruno (commented):
Everything looks good. The testing plan also makes sense, we talked about this offline too.

The only call outs I have here are
1. Just making sure that we plan on using the supported uploader later once we verify that sendbeacon is working.
2. The ignore-await stuff in sendbeacon_uploader feels like maybe we could do it differently, but I am going to take a look and get back to you tomorrow.

glean/src/entry/web.ts
Dexterp37: Yes, that work is going to be done after we trying this out in the wild for a bit and understand the pipeline implications. I suspect we're going to have another uploader likely composing the two.

glean/src/platform/browser/sendbeacon_uploader.ts
Dexterp37: Sure! I couldn't come up with a better way, but I don't think we can do it any differently I'm afraid.

Dexterp37:
All right, I believe this is finally ready for review folks! I'm able to see the live data using [this local hack](https://gist.github.com/Dexterp37/6c45cb4684528e828e96a5e6c683773f) and with a query like:

```sql
select * FROM `moz-fx-data-shared-prod.debug_ping_view_live.events_v1` AS e where date(submission_timestamp) >= "2023-11-30" order by submission_timestamp desc
```

glean/src/core/upload/worker.ts
```diff
@@ -70,13 +70,18 @@ class PingUploadWorker {
...
-      const needsHeaders = (
-        (Context.config.sourceTags === undefined) && (Context.config.debugViewTag === undefined)
-      ) ? false : true;
-      const safeUploader = needsHeaders ? Context.platform.uploader : this.uploader;
+      let safeUploader = this.uploader;
+      if (!this.uploader.supportsCustomHeaders()) {
+        // Some options require us to submit custom headers. Unfortunately not all the
+        // uploaders support them (e.g. `sendBeacon`). In case headers are required, switch
+        // back to the default uploader that, for now, supports headers.
+        const needsHeaders = (
```
rosahbruno, line 78: I think this should just be a boolean without the need for a ternary, but not a big deal.

rosahbruno (approved):
I think everything here looks good! Excited to get this landed and start testing

PR #1835, opened 2023-11-23
Bug 1867126 - automatic page load instrumentation

rosahbruno:
### Description
Configure auto instrumentation for page_load events.

Built-in `page_load` events can now be recorded in 2 ways
1. Automatically via the `enableAutoPageLoadEvents` parameter in the Glean configuration object. Whenever this is set to true, page load events will be collected automatically whenever `initialize` is called.
2. Manually via the `@mozilla/glean/metrics` object. We now expose `pageLoad` which is a function that a user can call at any time to record the same page load event that would have been collected automatically.

**Automatic instrumentation**
```js
Glean.initialize("app-name", true, { enableAutoPageLoadEvents: true });
```

Example: https://debug-ping-preview.firebaseapp.com/pings/glean-from-website/9f6996dc-a336-46eb-8c93-a1278046ae10#L10

**Manually call page_load**
```js
import GleanMetrics from "@mozilla/glean/metrics";

// ...

GleanMetrics.pageLoad({
  // Any of the values can be overwritten. If they are not overwritten,
  // the values pulled from the browser will be used.
  //
  // If `enableAutoPageLoadEvents` is true, then this call is a no-op.
  title: "Title Override"
});
```

Example: https://debug-ping-preview.firebaseapp.com/pings/glean-from-website/81fd045b-c8da-4229-8464-2860b054ece7#L10


glean/package.json
```diff
@@ -11,7 +11,8 @@
     "./private/ping": "./dist/core/pings/ping_type.js",
     "./uploader": "./dist/core/upload/uploader.js",
     "./testing": "./dist/core/testing/index.js",
-    "./web": "./dist/entry/web.js"
+    "./web": "./dist/entry/web.js",
+    "./built_in_events": "./dist/core/built_in_events/index.js"
```
Dexterp37, line 15: Don't know where else to put this comment, but what about exporting a `GleanMetrics` "namespace" instead of a "built_in_events"? We might have other things in the future

glean/src/core/built_in_events/index.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import type { PageLoadValues } from "./page_load.js";
+
+import { Context } from "../context.js";
+import { recordPageLoadEvent } from "./page_load.js";
+
+/**
+ * Wrapper for `recordPageLoadEvent` that checks if page load events
+ * are already being automatically recorded, if so we don't want to
+ * record them in two different ways.
```
Dexterp37, line 13: Why is that we don't want to allow two different things at the same time? I'd expect the ability to be able to record page events to be somewhat useful in case automated instrumentation fails on some odd edge case.

I'd imagine folks using automatic instrumentation would not really think about using this anyway.

Removing this gatekeeping would allow us to just directly expose the event.

glean/src/core/built_in_events/page_load.ts
```diff
@@ -0,0 +1,50 @@
...
+  title?: string;
+}
+
+/**
+ * Creates a `glean.page_load` event metric.
+ *
+ * @returns A metric type instance.
+ */
+function getGleanPageLoadEventMetric(): EventMetricType {
+  return new EventMetricType(
```
Dexterp37, line 21: What about keeping the same pattern we're using for external products or other metrics? 

1) Create the file containing the generated events (either using the glean_parser or manually, we don't have to regenerate it, we can do the same thing we do with other internal metrics)
2) Include the metrics file in some other file implementing the logic. It feels like this is a blend of the two things.

glean/src/metrics.yaml
```diff
@@ -413,3 +413,31 @@ glean:
...
     expires: never
+  page_load:
+    type: event
+    description: |
+      A event triggered whenever a page is loaded.
+
+      **This event by default is collected automatically. This can be
+      configured by the client and turned off. Glean provides a separate
+      API for collecting the same page load data if the client does not
+      want to collect page loads automatically.**
```
badboy, line 424: It's default-off. This text reads as if it is on though.

glean/src/core/built_in_events/index.ts
badboy: I agree that the check is not necessarily needed here. The docs should call out that usually calling it if `automaticallyRecordPageLoads` is true is not needed though.

I do not understand what @Dexterp37 means with the automated instrumentation cases though, could you expand on that?

Dexterp37: @badboy I could imagine web apps being a hybrid between single-page and multi-page requiring some tailoring, but maybe that's not that common? No strong opinion honestly

badboy: I see, though in that case you probably want to only use the manual way anyway

glean/src/core/built_in_events/page_load.ts
rosahbruno: I originally was going to include this in the `internal_metrics.ts` file where we declare the rest of the metrics recorded by Glean. I do agree it is a bit of the hybrid between the two.

We haven't really done anything like this in Glean yet, and for each of these built_in events, we are going to have only a few things that each event actually needs
1. The optional payload interface
2. A function to create the metric
3. A function to record the event

This was my attempt to create something self-contained that handled everything related to recording that type of event.

> Create the file containing the generated events (either using the glean_parser or manually, we don't have to regenerate it, we can do the same thing we do with other internal metrics)

I am not against this, but it feels a bit clunky checking in and then bundling the generated files. When I was doing this with the initial implementation of the page_loads it felt a bit weird.

> Include the metrics file in some other file implementing the logic. It feels like this is a blend of the two things.

Do you mean just taking the things out of this file and adding this directly into wherever it gets used rather than having everything in its own file like this?

glean/package.json
rosahbruno: This makes sense, I will look into creating a new namespace.

glean/src/core/built_in_events/index.ts
rosahbruno: Yeah the thought behind is was that this is more tailored for generic web projects. In the future, like we talked about at the workweek, we could have an implementation for a specific SPA framework so that they wouldn't need to worry about this.

The goal of this check was just to ensure that we aren't giving people the tools to shoot themselves in the foot. As far as frameworks or projects that allow you to have both SPA and MPAs, I am not entirely sure how common something like that may be.

I think removing the check is fine, I just don't want to create a scenario where users are accidentally recording multiple page load events. I guess if I update the documentation to call out how all this works, then we hopefully shouldn't run into any issues of not having this check.

rosahbruno: I removed this check now.

glean/src/core/built_in_events/index.ts
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * @param values Override values for event keys.
```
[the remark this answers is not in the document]
rosahbruno: These should be available via auto-complete in their IDE. We export the types now (in my [latest commit](https://github.com/mozilla/glean.js/pull/1835/commits/5737a88ba4a97c74956dcaeb22e59566bc343da3)), so they should see the types they need to supply.

These can also be added to the docs as well.

badboy (approved):

kirill-demtchouk:
Love this! ❤️

PR #1849, opened 2023-12-12
Bug 1868748 - truncate event extra strings to 500 bytes

rosahbruno:

### DIFF
--- a/glean/src/core/metrics/types/event.ts
+++ b/glean/src/core/metrics/types/event.ts
@@ -11,14 +11,14 @@ import {
   getMonotonicNow,
   isString,
   testOnlyCheck,
-  truncateStringAtBoundaryWithError
+  truncateStringAtBytesBoundaryWithError
 } from "../../utils.js";
 import { Context } from "../../context.js";
 import { ErrorType } from "../../error/error_type.js";
 import { MetricValidationError } from "../metric.js";
 
 const LOG_TAG = "core.metrics.EventMetricType";
-const MAX_LENGTH_EXTRA_KEY_VALUE = 100;
+const MAX_BYTES_EXTRA_KEY_VALUE = 500;
 
 /**
  * Base implementation of the event metric type,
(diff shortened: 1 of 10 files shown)


glean/src/core/metrics/types/event.ts
```diff
@@ -11,14 +11,14 @@ import {
...
-  truncateStringAtBoundaryWithError
+  truncateStringAtBytesBoundaryWithError
 } from "../../utils.js";
 import { Context } from "../../context.js";
 import { ErrorType } from "../../error/error_type.js";
 import { MetricValidationError } from "../metric.js";
 
 const LOG_TAG = "core.metrics.EventMetricType";
-const MAX_LENGTH_EXTRA_KEY_VALUE = 100;
+const MAX_BYTES_EXTRA_KEY_VALUE = 500;
```
travis79, line 21: Big ⭐ for more descriptive naming

glean/src/core/utils.ts
```diff
@@ -241,6 +241,39 @@ export function truncateStringAtBoundaryWithError(metric: MetricType, value: unk
...
+ * If the string required truncation, records an error through the error
+ * reporting mechanism.
+ *
+ * @param metric The metric to record an error to, if necessary,
+ * @param value The string to truncate.
+ * @param maxBytes The max number of bytes to truncate to.
+ * @returns A string with at most `maxLength` bytes.
+ * @throws In case `value` is not a string.
+ */
+export function truncateStringAtBytesBoundaryWithError(metric: MetricType, value: unknown, maxBytes: number): string {
```
travis79, line 256: Are you planning on completely replacing the `truncateStringAtBoundaryWithError` function with this?

rosahbruno: This old function gets used in a bunch of different places and I am now realizing that in all instances we aren't _really_ checking the length of strings by bytes. In all spots we are just using a max character number, which in most instances assuming the strings only include normal characters would be close enough, but I think the more full approach is to always check bytes length, like we are now.

So long answer short, I think that it might be worth replacing that entire function with the new one and always properly checking for bytes length rather than just character length.

rosahbruno: So this turned into a yes and no.

I compared the places where we were using `truncateStringAtBoundaryWithError` and in most of the places we actually should've been checking for bytes, we were only checking for string length.

Going off of the Glean Book, URLs are supposed to be 8192 characters before we convert to bytes. The boundary for all the other metrics is bytes.

Now we use the `truncateStringAtBytesBoundaryWithError` function everywhere except for with URLs where we continue to use `truncateStringAtBoundaryWithError`.

travis79 (approved):
This is looking really good. I especially like the more descriptive naming of the constants.

PR #1850, opened 2023-12-16
Bug 1862955 - Gleanjs session implementation

rosahbruno:
PR for updating Glean schema: https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/795

### DIFF
--- a/automation/compat/index.js
+++ b/automation/compat/index.js
@@ -8,19 +8,12 @@ import Glean from "@mozilla/glean/web";
 import { benchmark } from "./generated/pings.js";
 import * as metrics from "./generated/sample.js";
 
-Glean.setSourceTags(["automation"]);
-Glean.initialize("glean-compat-benchmark", true, {
-  enableAutoPageLoadEvents: true
-});
-
-metrics.pageLoaded.set();
-benchmark.submit();
-
 // !BIG HACK!
 //
 // Overwrite the console.info function in order to know when (and if) the benchmark ping was sent.
 // If a success ping message is logged we show that in the document.
 let pingSubmissionCount = 0;
+let sessionId = "";
 console.info = function () {
   var message = "";
   for (var i = 0; i < arguments.length; i++) {
@@ -30,6 +23,7 @@ console.info = function () {
       message += " ";
     }
   }
+
   console.log(message);
   if (/successfully sent 200.$/.test(message)) {
     pingSubmissionCount++;
@@ -38,8 +32,40 @@ console.info = function () {
     // 1. The built-in page_load event, which submits an events ping.
     // 2. The benchmark ping.
     if (pingSubmissionCount == 2) {
-      var elem = document.getElementById("msg");
+      var elem = document.getElementById("ping_msg");
       elem.innerHTML = "Pings submitted successfully.";
     }
   }
+
+  const sessionRegex = /"session_id": .+"/;
+  const sessionInfo = sessionRegex.exec(message);
+  if (!!sessionInfo) {
+    const currSessionId = sessionInfo?.[0].split(`"`)?.[3];
+    if (!!sessionId) {
+      if (currSessionId !== sessionId) {
+        var elem = document.getElementById("session_msg");
+        elem.innerHTML = "Session IDs updated successfully.";
+      } else {
+        console.log("Something went wrong...");
+      }
+    }
+
+    sessionId = currSessionId;
+  }
 }
+
+Glean.setSourceTags(["automation"]);
+
+// This needs to be set so we can pull the session ID from the log messages.
+Glean.setLogPings(true);
+
+Glean.initialize("glean-compat-benchmark", true, {
+  enableAutoPageLoadEvents: true,
+  // Setting the override to 0 means every action will trigger
+  // a new session. We use this to check that the session ID
+  // changes whenever a session has expired.
+  sessionLengthInMinutesOverride: 0
+});
+
+metrics.pageLoaded.set();
+benchmark.submit();
--- a/automation/compat/index.js
+++ b/automation/compat/index.js
@@ -33,7 +33,7 @@ console.info = function () {
     // 2. The benchmark ping.
     if (pingSubmissionCount == 2) {
       var elem = document.getElementById("ping_msg");
-      elem.innerHTML = "Pings submitted successfully.";
+      elem.innerText = "Pings submitted successfully.";
     }
   }
 
@@ -44,7 +44,7 @@ console.info = function () {
     if (!!sessionId) {
       if (currSessionId !== sessionId) {
         var elem = document.getElementById("session_msg");
-        elem.innerHTML = "Session IDs updated successfully.";
+        elem.innerText = "Session IDs updated successfully.";
       } else {
         console.log("Something went wrong...");
       }
--- a/glean/src/core/config.ts
+++ b/glean/src/core/config.ts
@@ -55,7 +55,7 @@ export interface ConfigurationInterface {
   readonly enableAutoElementClickEvents?: boolean,
   // Experimentation identifier to be set in all pings
   experimentationId?: string,
-  // Allow the client to explicitly define the length of a session in MINUTES.
+  // Allows custom session length, in minutes. The default value is 30 minutes.
   readonly sessionLengthInMinutesOverride?: number,
 }
 
--- a/glean/src/core/config.ts
+++ b/glean/src/core/config.ts
@@ -55,6 +55,8 @@ export interface ConfigurationInterface {
   readonly enableAutoElementClickEvents?: boolean,
   // Experimentation identifier to be set in all pings
   experimentationId?: string,
+  // Allow the client to explicitly define the length of a session in MINUTES.
+  readonly sessionLengthInMinutesOverride?: number,
 }
 
 // Important: the `Configuration` should only be used internally by the Glean singleton.
@@ -69,6 +71,7 @@ export class Configuration implements ConfigurationInterface {
   readonly enableAutoPageLoadEvents?: boolean;
   readonly enableAutoElementClickEvents?: boolean;
   experimentationId?: string;
+  readonly sessionLengthInMinutesOverride?: number;
 
   // Debug configuration.
   debug: DebugOptions;
@@ -85,6 +88,7 @@ export class Configuration implements ConfigurationInterface {
     this.enableAutoPageLoadEvents = config?.enableAutoPageLoadEvents;
     this.enableAutoElementClickEvents = config?.enableAutoElementClickEvents;
     this.experimentationId = config?.experimentationId;
+    this.sessionLengthInMinutesOverride = config?.sessionLengthInMinutesOverride;
 
     this.debug = {};
 
--- a/glean/src/core/internal_metrics.ts
+++ b/glean/src/core/internal_metrics.ts
@@ -223,7 +223,7 @@ export class CoreMetrics {
       this.generateNewSession();
     }
 
-    // Set the session new session last active time in LocalStorage.
+    // Update the last-active timestamp in LocalStorage to the current time.
     localStorage.setItem("glean_session_last_active", Date.now().toString());
   }
 
--- a/glean/src/core/internal_metrics.ts
+++ b/glean/src/core/internal_metrics.ts
@@ -273,7 +273,7 @@ export class CoreMetrics {
   }
 
   private generateNewSession(): void {
-    this.sessionId.set(generateUUIDv4());
+    this.sessionId.generateAndSet();
     this.sessionCount.add();
   }
 
--- a/glean/src/core/internal_metrics.ts
+++ b/glean/src/core/internal_metrics.ts
@@ -183,7 +183,8 @@ export class CoreMetrics {
 
   /**
    * Update local stored session information for Glean. This is called whenever
-   * the app is initialized and on every read/write to storage.
+   * the app on initialization and just after every read/write/delete to/from
+   * storage.
    *
    * There are a few scenarios to handle depending on what we already have
    * stored about the session and how long it has been since the last action.
--- a/glean/src/core/internal_metrics.ts
+++ b/glean/src/core/internal_metrics.ts
@@ -190,8 +190,8 @@ export class CoreMetrics {
    *
    * SCENARIOS:
    *
-   * 1. If this is the first session, then we set a new session ID and a
-   * lastActive timestamp.
+   * 1. If this is the first session (there is no existing session ID),
+   * then we set a new session ID and a lastActive timestamp.
    *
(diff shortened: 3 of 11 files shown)


glean/src/metrics.yaml
```diff
@@ -122,6 +122,40 @@ glean.internal.metrics:
       - glean-team@mozilla.com
     expires: never
 
+  session_id:
+    type: uuid
+    description: |
+      A UUID uniquely identifying the client's current session.
+    send_in_pings:
+      - glean_client_info
+    lifetime: user
```
chutten, line 131: Innnnteresting. Not `application`? Oh! Right! Because `application` doesn't preserve across navigation, does it?

glean/src/metrics.yaml
```diff
@@ -122,6 +122,40 @@ glean.internal.metrics:
       - glean-team@mozilla.com
     expires: never
 
+  session_id:
+    type: uuid
+    description: |
+      A UUID uniquely identifying the client's current session.
```
chutten, line 128: Define or link to a definition of what a session is, please.

glean/src/metrics.yaml
```diff
@@ -122,6 +122,40 @@ glean.internal.metrics:
...
+    data_sensitivity:
+      - technical
+    notification_emails:
+      - glean-team@mozilla.com
+    expires: never
+
+  session_count:
+    type: counter
+    description: |
+      A running counter of the number of sessions for this client.
```
chutten, line 145: Define or link to a definition of what a session is, please.

glean/src/core/internal_metrics.ts
```diff
@@ -159,6 +181,54 @@ export class CoreMetrics {
...
+  /**
+   * Update local stored session information for Glean. This is called whenever
+   * the app is initialized and on every read/write to storage.
+   *
+   * There are a few scenarios to handle depending on what we already have
+   * stored about the session and how long it has been since the last action.
+   *
+   * SCENARIOS:
+   *
+   * 1. If this is the first session, then we set a new session ID and a
```
chutten, line 193:
```suggestion
   * 1. If this is the first session (there is no existing session ID), then we set a new session ID and a
```

glean/src/core/internal_metrics.ts
```diff
@@ -202,6 +272,11 @@ export class CoreMetrics {
     }
   }
 
+  private generateNewSession(): void {
+    this.sessionId.set(generateUUIDv4());
```
chutten, line 276: We don't have `generateAndSet()` in gleanjs?

glean/src/core/sessions.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+export function hasSessionBeenInactiveForOverThirtyMinutes(): boolean {
+  const lastActive = localStorage.getItem("glean_session_last_active");
+  const lastActiveDate = new Date(Number(lastActive));
+
+  // Create a date 30 minutes ago to compare to our lastActiveDate.
+  //
+  // 60000 - number of milliseconds in a minute
+  // 30 - the number of minutes that can pass before a session is inactive
+  const thirtyMinutesAgo = new Date(Date.now() - (60000 * 30));
```
chutten, line 19: Thoughts on 

```js
let earlier = new Date();
earlier.setMinutes(earlier.getMinutes() - 30);
```

chutten (changes requested):
We're firm on 30mins? Might make testing difficult (btw, where are the tests?) unless we make that parametrized. (in addition to the existing testing difficulties with using `Date`)

glean/src/metrics.yaml
rosahbruno: Well this can change depending on the client. On a client where they are using basic HTML or something of that nature, it will initialize every page load. For a single page application, if will not initialize every page load. I think we need to cover as many cases as possible and just make the metric tied to the user lifetime so that we can maintain the value across different initializations.

glean/src/core/internal_metrics.ts
rosahbruno: Fixed in b0c301c4c1fb532c53ae7d6ec5b618222673f024

glean/src/core/internal_metrics.ts
rosahbruno: Ahh, good callout. I will fix this.

rosahbruno:
> We're firm on 30mins? Might make testing difficult
Based on everything that Abhishek found when researching sessions, it seems like 30 minutes is the standard here. I believe we can make 30 minutes the default.

> (btw, where are the tests?)
Tests are on their way (after I fix the last part)

> unless we make that parametrized. (in addition to the existing testing difficulties with using `Date`)
I agree this should be parameterized. I think 30 will be the default value unless someone provides a different value when initializing Glean. I will make some changes and push that up.

glean/src/core/internal_metrics.ts
rosahbruno: Fixed in https://github.com/mozilla/glean.js/pull/1850/commits/4a183f86f6520e5c951226aca2e7845f2007dc9f

glean/src/core/sessions.ts
rosahbruno: Fixed in https://github.com/mozilla/glean.js/pull/1850/commits/b30774da1da358c9b7c5b76fcc8acc6678262417

glean/src/metrics.yaml
rosahbruno: I've added a more thorough description of what a session is in the metrics.yaml for both metrics.

automation/compat/index.js
```diff
@@ -38,8 +32,40 @@ console.info = function () {
...
   }
+
+  const sessionRegex = /"session_id": .+"/;
+  const sessionInfo = sessionRegex.exec(message);
+  if (!!sessionInfo) {
+    const currSessionId = sessionInfo?.[0].split(`"`)?.[3];
+    if (!!sessionId) {
+      if (currSessionId !== sessionId) {
+        var elem = document.getElementById("session_msg");
+        elem.innerHTML = "Session IDs updated successfully.";
```
chutten, line 47: I know this doesn't matter, but my reflexes are forcing me to ask that this become `innerText`

glean/src/core/config.ts
```diff
@@ -55,6 +55,8 @@ export interface ConfigurationInterface {
   readonly enableAutoElementClickEvents?: boolean,
   // Experimentation identifier to be set in all pings
   experimentationId?: string,
+  // Allow the client to explicitly define the length of a session in MINUTES.
```
chutten, line 58: "client" is one of those words we define in different ways, but I'm not sure is suitable here. "Embedder", perhaps?

glean/src/core/internal_metrics.ts
```diff
@@ -28,6 +30,8 @@ export class CoreMetrics {
   readonly osVersion: StringMetricType;
   readonly architecture: StringMetricType;
   readonly locale: StringMetricType;
+  readonly sessionId: UUIDMetricType;
+  readonly sessionCount: CounterMetricType;
```
chutten, line 34: ...in what way is sessionCount not a session id? Do we even need a session id if we have a session count and a client id?

Also, how does this interact with pings forbidding client_id? Do we forbid session_id as well? (If we don't, we could deanonymize)

glean/src/core/internal_metrics.ts
```diff
@@ -159,6 +181,52 @@ export class CoreMetrics {
     }
   }
 
+  /**
+   * Update local stored session information for Glean. This is called whenever
+   * the app is initialized and on every read/write to storage.
```
chutten, line 186: "on" here meaning "just before" or "just after"?

chutten: Also: we update `lastActive` on reads as well as writes?

glean/src/core/internal_metrics.ts
```diff
@@ -159,6 +181,52 @@ export class CoreMetrics {
...
+   *
+   * 2. If the session is not expired, then we only update the lastActive time.
+   *
+   * 3. If the session is expired (inactive threshold is more recent than lastActive)
+   * then we update the session ID, the session sequence number, and the lastActive time.
+   */
+  updateSessionInfo(): void {
+    if (isWindowObjectUnavailable()) {
+      return;
+    }
```
chutten, line 204: Remind me: this only happens for SSR?

glean/src/core/internal_metrics.ts
```diff
@@ -159,6 +181,52 @@ export class CoreMetrics {
...
+      } catch (e) {
+        // Error parsing the last active timestamp, create a new session.
+        this.generateNewSession();
+      }
+    } else {
+      // There is no previous session information, create a new session.
+      this.generateNewSession();
+    }
+
+    // Set the session new session last active time in LocalStorage.
```
chutten, line 226: "Set the session new session last..." huh?

glean/src/metrics.yaml
```diff
@@ -122,6 +122,48 @@ glean.internal.metrics:
...
+    data_sensitivity:
+      - technical
+    notification_emails:
+      - glean-team@mozilla.com
+    expires: never
+
+  session_count:
+    type: counter
+    description: |
+      A running counter of the number of sessions for this client. A session is
```
chutten, line 149: We might want to spell out that this is client's lifetime count, it is only reset on opt-out. (Actually, does glean.js wipe `glean_client_info` on opt-out? I think it's supposed to)

glean/src/core/internal_metrics.ts
rosahbruno: > ...in what way is sessionCount not a session id? Do we even need a session id if we have a session count and a client id?

The `session_count` is the total number of sessions that this client has had. This gets updated every time that we trigger a new session. GA has a similar `session_number` that they track.

> Also, how does this interact with pings forbidding client_id? Do we forbid session_id as well? (If we don't, we could deanonymize)

This is not something that we have discussed. As a data steward, what are your thoughts here? Should we allow for session information without a client ID? Or should pings with no client ID also have no session info?

glean/src/core/internal_metrics.ts
rosahbruno: > "on" here meaning "just before" or "just after"?

It is happening immediately after read/writes. Since everything is running synchronously, this will always execute before we return from the functions. So I don't think "on" is technically incorrect, but I see what you are saying and I can adjust to say "just after".

> Also: we update lastActive on reads as well as writes?

Yes. Basically anytime that something on the screen is being interacted with. We wouldn't trigger a read or a write unless the user was interacting with the page. We don't have the traditional "onpageload" that we can fire in all applications, so this is the most reusable way we can accomplish this.

glean/src/core/config.ts
rosahbruno: Maybe I will just remove it altogether. Say something like "Allow for overriding the default session length of 30 minutes".

glean/src/core/internal_metrics.ts
rosahbruno: Yes, this is our check to ensure that we aren't trying to do any sort of storage actions on the server side.

glean/src/metrics.yaml
rosahbruno: > We might want to spell out that this is client's lifetime count, it is only reset on opt-out.

Good call, I will update this.

> (Actually, does glean.js wipe glean_client_info on opt-out? I think it's supposed to)

Yes, on opt out all the metrics are cleared from all the local databases (events, pings, metrics).

automation/compat/index.js
rosahbruno: Easy enough, I will update

glean/src/core/internal_metrics.ts
rosahbruno: Fixed in https://github.com/mozilla/glean.js/pull/1850/commits/9f89f0acf6e11a76be0571df25a3eb92fecaccc1

quiiver (approved):
:shipit:

quiiver:
> This is not something that we have discussed. As a data steward, what are your thoughts here? Should we allow for session information without a client ID? Or should pings with no client ID also have no session info?

Probably best to disallow session tracking if client_id is disabled

glean/src/core/internal_metrics.ts
chutten: (( The reason I ask about session count is because `client_id` + `session_count` has identical uniqueness semantics to `session_id`, but using less data. It's not a point that needs close interrogation (especially since not all pings have a `client_id` to make a `session_count` useful), but it's a note towards how we might want to think about identifiers. ))

Pings without client_id should also be without session_id. Otherwise it can be trivial to supply a client_id for most pings by joining on session_id.

Or, differently put: Any `session_id` on a ping without a `client_id` must not be linkable to a `session_id` on a ping of any other doctype (with `client_id` or not). (They _can_ have a `session_id`, but it must not be shared with any other ping or else they'll be linkable)

chutten (changes requested):
I don't know how easy it'll be, but if we want to support pings without `client_id` we need to ensure we aren't making it trivially obtainable via `session_id`

rosahbruno:
@chutten 
So I just want to make sure I am understanding - a ping sent with no `client_id` should also not include the `session_id` or the `session_count`? I agree with what you've said above, I just want to make sure I understand completely what the ask is before I start making any changes.

chutten:
> @chutten So I just want to make sure I am understanding - a ping sent with no `client_id` should also not include the `session_id` or the `session_count`? I agree with what you've said above, I just want to make sure I understand completely what the ask is before I start making any changes.

I don't know of a strong reason to exclude `session_count`. Yeah, there's some timing fingerprinting possible, but given the sizes of our populations and the ability for custom ping definitions to limit the precision of the timestamps, I'm not too worried.

But, yes, if there's no `client_id` there should be no `session_id` (as that would trivially permit deriving the `client_id` by looking for a ping where a `client_id` is sent and it shares the same `session_id`).

rosahbruno:
@chutten 
I've made updates to exclude the `session_id` whenever the ping doesn't include the `client_id`. 

A ping for reference: https://debug-ping-preview.firebaseapp.com/pings/glean-from-website/fa376d6f-f83d-45c6-a697-9fecdcfededf#L18

glean/src/metrics.yaml
```diff
@@ -122,6 +122,49 @@ glean.internal.metrics:
...
     expires: never
 
+  session_id:
+    type: uuid
+    description: |
+      A UUID uniquely identifying the client's current session. A session is
+      the period of time in which a user interacts with the application. After
+      a period of inactivity (default being 30 minutes) a new session will be
+      created the next time the user interacts with the application. On each
+      new session, the session_id will be updated.
```
chutten, line 132: Please update the description to note that this isn't included for pings where `include_client_id` is `false`

chutten (approved):

PR #1851, opened 2023-12-21
Bug1850323 - Add experimentation_id to Glean.js

travis79:

### DIFF
--- a/glean/src/core/pings/maker.ts
+++ b/glean/src/core/pings/maker.ts
@@ -233,7 +233,7 @@ export function collectPing(ping: CommonPingData, reason?: string): PingPayload 
   // !IMPORTANT! Events data needs to be collected BEFORE other metrics,
   // because events collection may result in recording of error metrics.
   const eventsData = Context.eventsDatabase.getPingEvents(ping.name, true);
-  const metricsData = Context.metricsDatabase.getPingMetrics(
+  let metricsData = Context.metricsDatabase.getPingMetrics(
     ping.name,
     true
   );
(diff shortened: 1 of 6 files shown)


glean/src/core/pings/maker.ts
```diff
@@ -249,6 +249,19 @@ export function collectPing(ping: CommonPingData, reason?: string): PingPayload
     );
   }
 
+  // Insert the experimentation id if the metrics aren't empty
+  if (ping.includeClientId && Context.config.experimentationId) {
+    if (metricsData != undefined) {
+      metricsData["string"]["glean.client.annotation.experimentation_id"] = Context.config.experimentationId;
```
rosahbruno, line 255: Is there any reason that we can't just record the metric rather than putting it into the metrics manually?

I may have answered my own question as I was writing this. I think I remember you said before that we just need to pass it along in every ping right? So we don't really need to record it, just make sure it is passed along. I guess if that is true then that's fine.

This is just a pattern that I don't think we use very often, but I think it makes sense for what you are trying to do here.

glean/src/metrics.yaml
```diff
@@ -316,6 +316,25 @@ glean.internal.metrics:
       - glean-team@mozilla.com
     expires: never
 
+glean.client.annotation:
+  experimentation_id:
+    type: string
+    lifetime: application
```
rosahbruno, line 322: Based on the code, does it actually matter what value we set here? We don't ever actually set it right? It just gets appended on every ping?

I'm not saying anything needs to change, but maybe a clarifying comment here could be useful if that is the case so that in the future if someone is trying to debug this or figure out how it works, they aren't misdirected by the lifetime

rosahbruno (approved):
I didn't have a chance to actually test this, but everything looks good. I had questions about a few things, but besides that I think the changes look good!

Always happy to see different people contributing to Glean.js 💯

glean/src/core/pings/maker.ts
travis79: In glean-core I did just this, record the metric (in a special store named `all-pings`) and then do this same logic but reading the metric from storage. The way I did this here was going to be like that and then I realized that this could be simplified and avoid the round trip through an actual metric and its storage. 

I know this is a bit more fragile, but since we control the use of this and with the added test coverage, I figured the risk was small that this would cause a regression that we wouldn't catch.

glean/src/metrics.yaml
travis79: That's a good idea, documentation via a comment would help to clarify this in the future. This is `application` lifetime because this is what it is in glean-core. Also, since the application is required to supply this with the config passed into `initialize()`, the behavior of this in the code stored as a runtime variable is pretty much identical to how it would be handled if this were just recorded as an actual metric, aside from any special handling that we do around application-lifetime metrics for "dirty-startup" stuff, but I'm pretty sure we only do that in glean-core.

PR #1866, opened 2024-01-30
Bug 1867821 - Implement an uploader that defaults to sendBeacon and falls back to…

badboy:
First try at this to get it going, see inline comments for questions.

### DIFF
--- a/glean/src/platform/browser/sendbeacon_fallback_uploader.ts
+++ b/glean/src/platform/browser/sendbeacon_fallback_uploader.ts
@@ -1,0 +1,42 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import type PingRequest from "../../core/upload/ping_request.js";
+
+import log, { LoggingLevel } from "../../core/log.js";
+import Uploader from "../../core/upload/uploader.js";
+import BrowserFetchUploader from "./fetch_uploader.js";
+import BrowserSendBeaconUploader from "./sendbeacon_uploader.js";
+import { UploadResultStatus } from "../../core/upload/uploader.js";
+import type { UploadResult } from "../../core/upload/uploader.js";
+
+const LOG_TAG = "platform.browser.SendBeaconFallbackUploader";
+
+class BrowserSendBeaconFallbackUploader extends Uploader {
+  fetchUploader = BrowserFetchUploader;
+  sendBeaconUploader = BrowserSendBeaconUploader;
+
+  // eslint-disable-next-line @typescript-eslint/require-await
+  async post(
+    url: string,
+    pingRequest: PingRequest<string | Uint8Array>
+  ): Promise<UploadResult> {
+
+    // Try `sendBeacon` first,
+    // fall back to `fetch` if `sendBeacon` reports an error.
+    const beaconStatus = await this.sendBeaconUploader.post(url, pingRequest);
+    if (beaconStatus.result == UploadResultStatus.Success) {
+      return beaconStatus;
+    }
+    log(LOG_TAG, "The `sendBeacon` call was not serviced by the browser. Falling back to the fetch uploader.", LoggingLevel.Warn);
+
+    return this.fetchUploader.post(url, pingRequest);
+  }
+
+  supportsCustomHeaders(): boolean {
+    return false;
+  }
+}
+
+export default new BrowserSendBeaconFallbackUploader();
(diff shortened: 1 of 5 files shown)


glean/package.json
```diff
@@ -54,7 +54,7 @@
     "test:unit": "run-s test:unit:core test:unit:platform",
     "test:unit:core": "npm run test:base -- \"tests/unit/core/**/*.spec.ts\" --recursive",
     "test:unit:platform": "npm run test:base -- \"tests/unit/platform/**/*.spec.ts\" --recursive --timeout 0",
-    "test:base": "node --experimental-modules --experimental-specifier-resolution=node --loader=ts-node/esm node_modules/mocha/lib/cli/cli.js",
+    "test:base": "node --experimental-modules --experimental-specifier-resolution=node --import=tsx/esm node_modules/mocha/lib/cli/cli.js",
```
badboy, line 57: This seems to be required for newer node, at least for my node v21.6.1 here.
Otherwise `npm run test:unit:platform` just fails with `Error: null`. Found https://github.com/mochajs/mocha/issues/5048 about this.

glean/src/platform/browser/sendbeacon_fallback_uploader.ts
```diff
@@ -0,0 +1,42 @@
...
+    // Try `sendBeacon` first,
+    // fall back to `fetch` if `sendBeacon` reports an error.
+    const beaconStatus = await this.sendBeaconUploader.post(url, pingRequest);
+    if (beaconStatus.result == UploadResultStatus.Success) {
+      return beaconStatus;
+    }
+    log(LOG_TAG, "The `sendBeacon` call was not serviced by the browser. Falling back to the fetch uploader.", LoggingLevel.Warn);
+
+    return this.fetchUploader.post(url, pingRequest);
+  }
```
badboy, line 35: The heart of it all.  We really just invoke the fetch uploader if the beacon uploader failed.

Question: should we store this state so that we don't even try sendBeacon again?

glean/src/platform/browser/sendbeacon_fallback_uploader.ts
```diff
@@ -0,0 +1,42 @@
...
+    pingRequest: PingRequest<string | Uint8Array>
+  ): Promise<UploadResult> {
+
+    // Try `sendBeacon` first,
+    // fall back to `fetch` if `sendBeacon` reports an error.
+    const beaconStatus = await this.sendBeaconUploader.post(url, pingRequest);
+    if (beaconStatus.result == UploadResultStatus.Success) {
+      return beaconStatus;
+    }
+    log(LOG_TAG, "The `sendBeacon` call was not serviced by the browser. Falling back to the fetch uploader.", LoggingLevel.Warn);
```
badboy, line 32: `sendBeaconUploader` logs a `LoggingLevel.Error` on failure, basically the same message as here. But because we try something else it's not really an error. Should we pass a flag that avoids that log message maybe?

glean/src/platform/browser/sendbeacon_fallback_uploader.ts
rosahbruno: Hmm.. that is a good question. I wonder if we want to maybe have a number of retries or something? If the `sendBeacon` call fails 3 times within a session (can track in SESSION_STORAGE) then we should stop falling back and just use `fetch` instead?

That may be overkill? What do you think?

glean/src/platform/browser/sendbeacon_fallback_uploader.ts
rosahbruno: Yeah, I think we should. If coming from this uploader, then we just skip that log message since it'll get logged after. Plus, when using this uploader we aren't actually deleting the data, we are sending it via fetch.

glean/package.json
badboy: I would have never guessed that o_O

glean/src/platform/browser/sendbeacon_fallback_uploader.ts
badboy: I wouldn't even go that far to store that anywhere but in a variable.
If sendBeacon fails once, don't try again unitl a proper page reload.

But even that seems maybe too much and the current code might be good enough.

rosahbruno: We chatted offline about this and for now this is good. We are going to create a follow-up bug for checking if `sendBeacon` is defined first as part of the fallback uploader.

rosahbruno (approved):
LGTM, I will file the follow up bug discussed offline and in previous comments.

rosahbruno:
@badboy 
Would you mind making a follow up PR to this with a changelog entry? We don't need to be too descriptive since I am going to create a google doc with info about the uploaders.

PR #1877, opened 2024-02-29
Bug 1862955 - fix browserstack session tests

rosahbruno:
This is a followup to #1850. The changes in that PR broke the browserstack tests. This updates our browserstack tests to rely on data from LocalStorage rather than trying to pull session IDs from log messages.

-  [x] **Documentation**: This PR includes documentation changes, an explanation of why it does not need that or a follow-up bug has been filed to do that work

### DIFF
--- a/automation/compat/index.js
+++ b/automation/compat/index.js
@@ -8,10 +8,26 @@ import Glean from "@mozilla/glean/web";
 import { benchmark } from "./generated/pings.js";
 import * as metrics from "./generated/sample.js";
 
+Glean.setSourceTags(["automation"]);
+Glean.initialize("glean-compat-benchmark", true, {
+  enableAutoPageLoadEvents: true,
+  // Setting the override to 0 means every action will trigger
+  // a new session. We use this to check that the session ID
+  // changes whenever a session has expired.
+  sessionLengthInMinutesOverride: 0
+});
+
+metrics.pageLoaded.set();
+benchmark.submit();
+
 // !BIG HACK!
 //
 // Overwrite the console.info function in order to know when (and if) the benchmark ping was sent.
 // If a success ping message is logged we show that in the document.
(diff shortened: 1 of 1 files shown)


automation/compat/index.js
```diff
@@ -37,10 +49,13 @@ console.info = function () {
     }
   }
 
-  const sessionRegex = /"session_id": .+"/;
-  const sessionInfo = sessionRegex.exec(message);
-  if (!!sessionInfo) {
-    const currSessionId = sessionInfo?.[0].split(`"`)?.[3];
+  // Pull all user lifetime metrics from Glean.
+  const userLifetimeMetrics = window.localStorage.getItem("userLifetimeMetrics");
```
Dexterp37, line 53: Instead of the big hack in this file, why can't we use the ping testing APIs?

travis79: If the hack is necessary, let's try and explain why `testBeforeNextSubmit` won't work here.

rosahbruno: Hmm.. it looks like `testBeforeNextSubmit` doesn't work here because Glean isn't in "testing mode". Whenever you try and invoke `testBeforeNextSubmit` from the browserstack app it hits [this check](https://github.com/mozilla/glean.js/blob/main/glean/src/core/utils.ts#L284) which throws the error message since Glean isn't in testing mode.

Is this the same way that the Glean SDKs work? Or can you call this function any time without having to be in testing mode?

travis79: No you would need to be in test mode, I think. There are other tests in glean-core where we operate outside of testmode too, so it's fine if that's what it takes to test this particular thing. Just leave a comment indicating this so that we understand why we aren't using it in this case.

travis79 (approved):

PR #1919, opened 2024-04-18
Bug 1888430 - Add `page_id`  to Glean automatic events

travis79:

### DIFF
--- a/glean/src/core/glean_metrics.ts
+++ b/glean/src/core/glean_metrics.ts
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import { Context } from "./context.js";
+import UUIDMetric from "./metrics/types/uuid.js";
 import EventMetricType from "./metrics/types/event.js";
 import { EVENTS_PING_NAME } from "./constants.js";
 import { Lifetime } from "./metrics/lifetime.js";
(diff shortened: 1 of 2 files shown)


glean/src/core/glean_metrics.ts
```diff
@@ -28,6 +31,8 @@ type ElementClickEventContext = {
  * via the Glean config or they are called manually by the client.
  */
 namespace GleanMetrics {
+  let currentPageId: string;
```
travis79, line 34: It complained about using `var` here, but I worry that `let` scoping might not do what I want here and keep this around as long as Glean is around?

rosahbruno: If you want to keep something around the entire duration that Glean is around, you could add a variable in the Glean.js `Context`. This is a singleton that we use pretty much everywhere to managing values that we want to persist for the current duration of the Glean.js instance.

You can probably accomplish the same thing in some way using the `GleanMetrics` namespace, but if you want it to behave exactly like the rest of the Glean instance, then Context might be the way to go.

travis79: Might be moot, since I've decided to go another way with the implementation here. But at least I know where to put things like this in the future.

rosahbruno (approved):
LGTM 👍

I will just leave this here until we get confirmation from Wil that everything looks ok, then we can land it.

PR #1935, opened 2024-05-06
Bug 1894064: remove migration code

rosahbruno:
The only documentation that we had of this was in the CHANGELOG, so there are no accompanying doc changes.

### Why this existed
We are removing this migration code since it is no longer needed. The reason for this code was to help move `client_id`s and `first_run_date`s over from our previous storage solution from when Glean.js was asynchronous. Newer versions of Glean.js use `localStorage` which allows synchronous read/writes. Glean.js used to use `IndexedDB` which forced asynchronous read/writes.

Synchronous is necessary so that we can ensure read/writes execute before pages are unloaded. If we start to write to storage async and the page unloads, that process never finishes and we incur data loss.

### Why it can go away
The only product that used Glean.js in production that was migrating to our new storage solution was MDN. They have been the only ever consumer of this feature. We have had the migration code running for ~9 months in MDN, so the majority of users would have migrated over their data by now.

We are removing it because of an error we are seeing when submitting the first ping for any user before they have migrated. A ping is submitted missing both the `client_id` and the `first_run_date` since they get lost in a race condition. This is causing a lot of schema ingestion errors.

We've run the migration code for long enough and now it is safe to turn it off. This will also stop the large amount of `first_run_date` errors that we have been seeing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 [Full changelog](https://github.com/mozilla/glean.js/compare/v5.0.1...main)
 
+* [#1935](https://github.com/mozilla/glean.js/pull/1935): **BREAKING CHANGE**: Remove `migrateFromLegacyStorage` capability and configuration option. If your project currently sets the `migrateFromLegacyStorage` value, this will no longer work.
+
 # v5.0.1 (2024-04-30)
 
 [Full changelog](https://github.com/mozilla/glean.js/compare/v5.0.0...v5.0.1)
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 [Full changelog](https://github.com/mozilla/glean.js/compare/v5.0.1...main)
 
+* [#1935](https://github.com/mozilla/glean.js/pull/1935): Remove `migrateFromLegacyStorage` capability and configuration option.
```
chutten, line 5: Presumably this is a breaking change and needs to be annotated as such?

chutten (approved):

CHANGELOG.md
rosahbruno: Good callout, updated now.

PR #1977, opened 2025-03-31
Add try-catch around `window.sessionStorage` access

chenba:

  - I was not able to find any existing tests around sessionStorage or some kind of mock for the window object
  - I'm not sure if we are going to tag a release right away.  I can update the changelog if that's the case.
  - This PR doesn't change the API.  It prevents an uncaught error that could break some JS based webapps when Glean is in the dependency tree.

### DIFF
--- a/glean/src/core/glean.ts
+++ b/glean/src/core/glean.ts
@@ -533,10 +533,19 @@ declare global {
 }
 
 // Only set `Glean` values whenever running inside of a browser.
-if (
-  typeof window !== "undefined" &&
-  typeof window.sessionStorage !== "undefined"
-) {
+// When cookies/localStorage/sessionStorage are disabled (at least in Firefox
+// and Chrome), an access attempt on the window.sessionStorage property will
+// throw an error ("The operation is insecure." in Firefox).
+// So try-catch is needed here.
+let hasStorage = false;
+try {
+  hasStorage = typeof window !== "undefined" &&
+  typeof window.sessionStorage !== "undefined";
+} catch (e) {
+  console.error("No session storage available", e);
+}
+
+if (hasStorage) {
   window.Glean = {
     setLogPings: (flag: boolean) => {
       setDebugOptionInSessionStorage(DebugOption.LogPings, flag);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased changes
 
+* [#1977](https://github.com/mozilla/glean.js/pull/1977): Add try-catch around window.sessionStorage access to prevent uncaught error upon `import`.
+
 [Full changelog](https://github.com/mozilla/glean.js/compare/v5.0.3...main)
 
 # v5.0.3 (2024-08-02)


glean/src/core/glean.ts
```diff
@@ -533,48 +533,56 @@ declare global {
...
-          `https://debug-ping-preview.firebaseapp.com/stream/${Context.config.debugViewTag}#${sessionId}`,
-          "_blank"
+// When cookies/localStorage/sessionStorage are disabled (at least in Firefox
+// and Chrome), an access attempt on the window.sessionStorage property will
+// throw an error ("The operation is insecure." in Firefox).
+// So try-catch is needed here.
+try {
+  if (
+    typeof window !== "undefined" &&
+    typeof window.sessionStorage !== "undefined"
```
Dexterp37, line 543: Instead of wrapping the whole block in try-catch, what about something like:

```typescript
let hasStorage = false;
try {
    hasStorage = typeof window !== "undefined"
        && typeof window.sessionStorage !== "undefined";
} catch (e) {
    console.error("No session storage available", e)
}

if (hasStorage) {
// ...
```

This would not increase indentation and would reduce the scope of the try block.

chenba:
Updated the patch based on @Dexterp37's suggestion.  Thank you, @Dexterp37

Dexterp37 (approved):
@chenba do you need a release cut ? Is this blocking something or was this problem found by some other investigation?

chenba:
> @chenba do you need a release cut ? Is this blocking something or was this problem found by some other investigation?

I do need a release.  That error is blocking https://accounts.firefox.com/cookies_disabled from being rendered if the user disabled cookies/session storage/local storage in their browser preferences.

chenba:
Updated patch to include an entry in the changelog under "Unreleased changes".

Dexterp37 (approved):

Dexterp37:
> Updated patch to include an entry in the changelog under "Unreleased changes".

Thanks. I merged this, but now I'm also wondering if this would lead to other problems if the generated metrics are ever called (or if windows.Glean.* stuff) is ever called. @chenba I'm ok with cutting a release for this, but have you tested it? Does it solve your problem or do we need more changes?

chenba:
Thanks, @Dexterp37.

> have you tested it? Does it solve your problem or do we need more changes?

I did test it.  And it solves our issue, which is that the error is thrown at `import`.

> this would lead to other problems if the generated metrics are ever called (or if windows.Glean.* stuff) is ever called.

I'm not familiar with this code base but it looks like the `window.Glean` API is mainly for development and debugging?  _If_ there's an issue then it'd be developer facing issue and not an end user one perhaps.  (Not that accidentally leaving debugging code in production is unheard of.)

Dexterp37:
> Thanks, @Dexterp37.
> 
> > have you tested it? Does it solve your problem or do we need more changes?
> 
> I did test it. And it solves our issue, which is that the error is thrown at `import`.

Ok, cutting a release

PR #1788, opened 2023-10-11
Bug 1858515 - Fix window is undefined error

rosahbruno:
This PR also has some minor code changes

This error was reported by the MDN folks when upgrading to v2.0.2 & v2.0.4. I was able to test this by pushing up a branch with the dist folder checked in to my fork (https://github.com/rosahbruno/glean.js/tree/1858515-DIST-TEST) and install this using `gitpkg` in the MDN project and run the app locally.

`yarn run dev` with fixes on branch
Screenshot 2023-10-12 at 3 00 36 PM

`yarn run dev` with v2.0.4
Screenshot 2023-10-12 at 2 59 05 PM

### Pull Request checklist ###

- [ ] **Quality**: Make sure this PR builds and runs cleanly.
  - Inside the `glean/` folder, run:
    - `npm run test` Runs _all_ tests
    - `npm run lint` Runs _all_ linters
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog**: This PR includes a changelog entry to `CHANGELOG.md` or an explanation of why it does not need one
- [ ] **Documentation**: This PR includes documentation changes, an explanation of why it does not need that or a follow-up bug has been filed to do that work
  - Documentation should be added to [The Glean Book](https://mozilla.github.io/glean/book/index.html) when making changes to Glean's user facing API
    - When changes to the Glean Book are necessary, link to the corresponding PR on the [`mozilla/glean`](https://github.com/mozilla/glean) repository
  - Documentation should be added to [the Glean.js developer documentation](https://github.com/mozilla/glean.js/tree/main/docs) when making internal code changes

PR #2010, opened 2025-08-27
Bug 1967425 - remove faulty check for sessionStorage to prevent uncau…

bershanskiy:
If user instructs User Agent to block cookies, every access to sessionStorage outside of try-catch block results in an error. Prior to this commit, `setDebugOptionsFromSessionStorage()` contained a faulty check for this case leading to unpredictable breakages in glean embedders. In particular, Glean was breaking pages on MDN https://developer.mozilla.org/

For details see: https://github.com/mdn/yari/issues/13039

This is just one of many ways o fix the bug. Specifically, I removed faulty check in `setDebugOptionsFromSessionStorage()` in favor of a properly working check (the try-catch block) in `getDebugOptionFromSessionStorage()`.

Tests:
I did not add any tests here because I did not any extra functionality, just removed a faulty check in favor of a functional one.

Changelog:
Updated `CHANGELOG.md` with a reference to this PR.

Documentation:
This change does not require any documentation.

### DIFF
--- a/glean/src/core/glean.ts
+++ b/glean/src/core/glean.ts
@@ -194,14 +194,6 @@ namespace Glean {
    * Fetch debug options from SessionStorage and set the Glean preInit debug options.
    */
   function setDebugOptionsFromSessionStorage() {
-    // If we cannot access browser APIs, we do nothing.
-    if (
-      typeof window === "undefined" ||
-      typeof window.sessionStorage === "undefined"
-    ) {
-      return;
-    }
-
     const logPings = getDebugOptionFromSessionStorage(DebugOption.LogPings);
     if (logPings) {
       preInitLogPings = extractBooleanFromString(logPings);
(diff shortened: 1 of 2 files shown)


glean/src/core/glean.ts
```diff
@@ -194,14 +194,6 @@ namespace Glean {
    * Fetch debug options from SessionStorage and set the Glean preInit debug options.
    */
   function setDebugOptionsFromSessionStorage() {
-    // If we cannot access browser APIs, we do nothing.
```
Dexterp37, line 197: @bershanskiy @travis79 if we remove this, wouldn't we hit what #1788 was supposed to fix, again?

bershanskiy: `getDebugOptionFromSessionStorage()` catches this exception properly (lines 73-82):
```js
const getDebugOptionFromSessionStorage = (
  option: DebugOptionValue
): string | undefined => {
  try {
    return sessionStorage.getItem(`Glean.${option.toString()}`) || undefined;
  } catch (e) {
    console.warn(e);
    return undefined;
  }
};
```

This PR removes one faulty redundant check to let the code execution to actually reach the proper try-catch.

